Issue #5, opened 2026-03-13
Stale handshake


v0l:
When restarting `fips` all connections dont re-establish and have the following error:

```bash
Mar 13 14:45:07 kieran-x fips[80168]: 2026-03-13T14:45:07.185134Z  INFO fips::node::handlers::rx_loop: RX event loop started
Mar 13 14:45:07 kieran-x fips[80168]: 2026-03-13T14:45:07.185192Z  INFO fips::control: Control socket listening path=/run/fips/control.sock
Mar 13 14:45:37 kieran-x fips[80168]: 2026-03-13T14:45:37.186240Z  INFO fips::node::handlers::timeout: Stale handshake connection timed out link_id=link:1 direction=outbound idle_secs=30
Mar 13 14:46:18 kieran-x fips[80168]: 2026-03-13T14:46:18.186013Z  INFO fips::node::handlers::timeout: Stale handshake connection timed out link_id=link:2 direction=outbound idle_secs=31
Mar 13 14:46:39 kieran-x fips[80168]: 2026-03-13T14:46:39.214551Z  INFO fips::node::handlers::handshake: Outbound handshake completed peer=fips-v0l-io link_id=link:3 their_index=2dfad24d
Mar 13 14:46:39 kieran-x fips[80168]: 2026-03-13T14:46:39.214637Z  INFO fips::node::handlers::handshake: Connection promoted to active peer peer=fips-v0l-io link_id=link:3 our_index=60d1db9a their_index=2dfad24d
Mar 13 14:46:39 kieran-x fips[80168]: 2026-03-13T14:46:39.214653Z  INFO fips::node::handlers::handshake: Peer promoted to active peer=fips-v0l-io
Mar 13 14:46:39 kieran-x fips[80168]: 2026-03-13T14:46:39.215010Z  INFO fips::node::tree: Processed TreeAnnounce from=fips-v0l-io seq=1 depth=0 root=9d1192e800bad0eed34b0eb2738e0e5d
Mar 13 14:46:39 kieran-x fips[80168]: 2026-03-13T14:46:39.215207Z  INFO fips::node::tree: Parent switched, flushed coord cache, announcing to all peers new_parent=fips-v0l-io new_seq=2 new_root=9d1192e800bad0eed34b0eb2738e0e5d depth=1
[...]

PR #6, opened 2026-03-13
Fix stale session and backoff reset after peer removal (issue #5)

v0l:
## Summary

Fixes two bugs that caused connections to fail re-establishing after a restart, as reported in #5.

## Bug 1: `remove_active_peer` does not remove the end-to-end session

**Location:** `src/node/handlers/dispatch.rs` — `remove_active_peer`

When a peer is evicted (link-dead timeout or disconnect), the function removed the peer from `self.peers` and freed Noise session indices, but **never removed the entry from `self.sessions`**. The stale `Established` session entry caused:

1. `check_session_mmp_reports()` kept iterating the dead session and emitting `"MMP session metrics"` with frozen counters on every 30s tick — visible in the issue log at 14:49:01 and 14:49:31, after the peer was removed at 14:48:45.
2. `initiate_session()` found `is_established() == true` on the stale entry and **silently returned `Ok(())`** without creating a new session. The link layer would reconnect, but the end-to-end session was never re-established, blocking all application traffic.
3. `purge_idle_sessions()` eventually cleaned up the stale entry, but only after `idle_timeout_secs` elapsed — explaining the "MMP session teardown" log at 14:49:57 with the same frozen packet counts.

**Fix:** `remove_active_peer` now removes `self.sessions[node_addr]` and `self.pending_tun_packets[node_addr]`, emitting the session MMP teardown log immediately.

## Bug 2: `schedule_reconnect` resets exponential backoff on every link-dead cycle

**Location:** `src/node/retry.rs` — `schedule_reconnect`

Every call created a **fresh `RetryState` with `retry_count=0`**, unconditionally overwriting any existing entry. After each link-dead removal → reconnect → death cycle, the accumulated backoff was discarded and the node restarted from the base interval (5s), hammering reconnects indefinitely.

**Fix:** When a retry entry already exists, increment its count rather than overwriting it, so backoff continues to accumulate across link-dead cycles.

## Tests

- `test_disconnect_clears_session` — verifies `session_count()` drops to 0 after peer removal via disconnect
- `test_schedule_reconnect_preserves_backoff` — verifies reconnect increments existing retry count rather than resetting it
- `test_schedule_reconnect_fresh_state` — verifies a fresh reconnect (no prior entry) still starts at base delay

All 814 tests pass.

### DIFF
--- a/src/node/handlers/dispatch.rs
+++ b/src/node/handlers/dispatch.rs
@@ -109,12 +109,28 @@ impl Node {
         }
 
         // MMP teardown log (before we drop the peer)
+        let peer_name = self.peer_aliases.get(node_addr)
+            .cloned()
+            .unwrap_or_else(|| peer.identity().short_npub());
         if let Some(mmp) = peer.mmp() {
-            let name = self.peer_aliases.get(node_addr)
-                .cloned()
-                .unwrap_or_else(|| peer.identity().short_npub());
-            Self::log_mmp_teardown(&name, mmp);
+            Self::log_mmp_teardown(&peer_name, mmp);
         }
+
+        // Remove any end-to-end session associated with this peer.
+        //
+        // Sessions are tracked separately from peers (self.sessions vs self.peers).
+        // Leaving a stale session alive after removing the peer causes:
+        //   1. check_session_mmp_reports() keeps logging stale "MMP session metrics"
+        //      with frozen counters until purge_idle_sessions() eventually fires.
+        //   2. initiate_session() finds is_established() == true on the stale entry
+        //      and silently returns Ok(()), preventing a new session from being
+        //      established even after the link layer reconnects successfully.
+        if let Some(session_entry) = self.sessions.remove(node_addr) {
+            if let Some(mmp) = session_entry.mmp() {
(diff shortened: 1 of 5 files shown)


jmcorgan:
I've recently push a set of protocol fixes that touch these same areas. It's pending CI passage, then it will get merged to master. Could you reevaluate the need for the above based on latest master when that's done?

jmcorgan:
Nice work on fips-dash, btw.

jmcorgan:
I've cherry-picked commits 1 and 2 onto current master as branch `trial/pr6-fixes` (`e8ef15a`). Commit 3 (auto_reconnect ignoring max_retries) was already fixed in `324535e` so it was skipped.

Minor fixups needed to apply cleanly on current master:
- Collapsed nested `if let` in `dispatch.rs` to satisfy `clippy::collapsible_if`
- Removed third `bool` argument from `schedule_retry()` calls in the backoff test (signature changed since this PR's base)

All 869 tests pass (866 existing + 3 new from this PR), clippy clean. I'm waiting on CI to pass and I'll merge.

jmcorgan:
The trial branch has been merged into master, so I'm closing this out now.

PR #12, opened 2026-03-16
Fix post-rekey jitter spikes from drain-window frames

### LINKED ISSUE #10: Post-rekey jitter spikes from drain-window frames
## Symptom

After each rekey cutover (~every 2 minutes with default `after_secs=120`), the jitter metric spikes to 2,000–7,000ms before decaying back to the steady-state 0.2–1.0ms over 60–90 seconds. This means the jitter metric is corrupted roughly half the time.

Observed on a UDP link (~60ms SRTT):

| Time | Jitter (ms) | Context |
|------|-------------|---------|
| 11:22:53 | — | Rekey cutover (responder) |
| 11:22:55 | **7,089.9** | First report after rekey |
| 11:23:25 | 114.7 | Decaying |
| 11:23:55 | 2.1 | Normal |
| 11:24:54 | — | Rekey cutover (responder) |
| 11:24:57 | **6,231.3** | First report after rekey |

## Problem

`reset_for_rekey()` in `src/mmp/receiver.rs` resets `last_recv_time` to `None` and `last_sender_timestamp` to `0`, which correctly skips jitter update on the first post-rekey frame. However, during the 10-second drain window (`DRAIN_WINDOW_SECS`), frames encrypted with the old session key can still arrive and be decrypted via `previous_session`. These old-session frames carry sender timestamps that are minutes into the old session (large values), while the next new-session frame has a timestamp near 0. The resulting `send_delta_us` is hugely negative, producing an enormous `transit_delta` that spikes the EWMA jitter estimator.

## Proposed Fix

Add a post-rekey grace period that skips jitter updates for the first N frames after rekey, not just the first one. This absorbs any old-session frames arriving via the drain window.

In `ReceiverState`:

```rust
rekey_grace_remaining: u8,  // frames to skip after rekey
```

In `reset_for_rekey()`:

```rust
self.rekey_grace_remaining = 3;  // skip first 3 frames
```

In `record_recv()`, replace the existing jitter update block:

```rust
if self.rekey_grace_remaining > 0 {
    self.rekey_grace_remaining -= 1;
} else if let Some(prev_recv) = self.last_recv_time {
    let recv_delta_us = now.duration_since(prev_recv).as_micros() as i64;
    let send_delta_us = sender_us - (self.last_sender_timestamp as i64 * 1000);
    let transit_delta = (recv_delta_us - send_delta_us) as i32;
    self.jitter.update(transit_delta);
}

self.last_sender_timestamp = sender_timestamp_ms;
self.last_recv_time = Some(now);
```

The baselines still update every frame, but jitter calculation is suppressed until drain-window frames have flushed. An alternative is time-based grace (e.g., 15s after rekey) instead of frame count, which would be more robust regardless of MMP heartbeat interval.

jmcorgan:
## Summary

- Add time-based grace period (15s = drain window + 5s margin) to suppress jitter updates after rekey cutover
- Prevents old-session frames arriving via the drain window from producing 2,000–7,000ms jitter spikes
- Baselines still update every frame so jitter calculation resumes cleanly after grace expires

Fixes #10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   backoff capped at 300 seconds.
 - Control socket permissions: non-root users couldn't connect. Daemon now
   chowns socket and directory to `root:fips` group at bind time.
+- Post-rekey jitter spikes: old-session frames arriving via the drain window
+  produced 2,000–7,000ms jitter spikes that corrupted the EWMA estimator.
+  Added a 15-second grace period after rekey cutover that suppresses jitter
+  updates until drain-window frames have flushed. (#10)
 
 ## [0.1.0] - 2026-03-12
 
--- a/src/mmp/mod.rs
+++ b/src/mmp/mod.rs
@@ -198,8 +198,8 @@ impl MmpPeerState {
     }
 
     /// Reset counter-dependent state for rekey cutover.
-    pub fn reset_for_rekey(&mut self) {
-        self.receiver.reset_for_rekey();
+    pub fn reset_for_rekey(&mut self, now: Instant) {
+        self.receiver.reset_for_rekey(now);
         self.metrics.reset_for_rekey();
     }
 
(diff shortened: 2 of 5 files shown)


jmcorgan (commented):
Ah, this needs a changelog entry.

PR #15, opened 2026-03-16
fix: use tun_mtu instead of transport_mtu for TCP MSS clamping

v0l:
## Problem

TCP connections routed through the FIPS mesh (`fips0`) would stall after the handshake. Data segments were too large for the actual path MTU — FIPS dropped them and sent ICMPv6 Packet Too Big responses, but the kernel ignored them (PMTUD blackhole).

The root cause was that `run_tun_reader` and the TUN lifecycle setup were computing the maximum TCP MSS from `transport_mtu` (the local wire MTU, e.g. 1472) rather than `tun_mtu` (e.g. 1280). This produced an MSS that was too large because it did not account for `FIPS_IPV6_OVERHEAD` (77 bytes of encapsulation added on top of every inner IPv6 packet).

## Fix

Derive the effective IPv6 MTU from the TUN MTU instead:

```
effective_ipv6_mtu = tun_mtu - FIPS_IPV6_OVERHEAD  (e.g. 1280 - 77 = 1203)
max_mss            = effective_ipv6_mtu - 40 - 20   (e.g. 1203 - 60 = 1143)
```

`FIPS_IPV6_OVERHEAD` is already defined in `src/upper/icmp.rs` and `effective_ipv6_mtu()` is already exported — this change just passes `mtu` (TUN MTU) to it instead of `transport_mtu`.

## Changes

- `src/upper/tun.rs`: remove `transport_mtu` parameter from `run_tun_reader`; compute `effective_mtu` from `mtu` (TUN MTU)
- `src/node/lifecycle.rs`: call `effective_ipv6_mtu(mtu)` with TUN MTU instead of `self.effective_ipv6_mtu()` (which used transport MTU)

### DIFF
--- a/src/node/lifecycle.rs
+++ b/src/node/lifecycle.rs
@@ -551,15 +551,15 @@ impl Node {
                     let name = device.name().to_string();
                     let our_addr = *device.address();
 
+                    // Store TUN MTU so effective_ipv6_mtu() uses it from here on.
+                    self.tun_mtu = Some(mtu);
+
                     info!("TUN device active:");
                     info!("     name: {}", name);
                     info!("  address: {}", device.address());
                     info!("      mtu: {}", mtu);
 
-                    // Calculate max MSS for TCP clamping from the TUN MTU.
-                    // effective_ipv6_mtu(tun_mtu) subtracts FIPS_IPV6_OVERHEAD (77 bytes)
-                    // to account for FIPS encapsulation on top of the inner IPv6 packet.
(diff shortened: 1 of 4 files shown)


v0l:
Closing — the daemon's MSS clamping was correct all along. The TUN MTU (1280) is the inner IPv6 packet size visible to the OS; subtracting FIPS overhead from it before computing MSS was wrong. The correct formula is tun_mtu - 60 = 1220, which is exactly what the kernel already does. The actual fix for the sidecar TCP stall was in entrypoint.sh: the ip6tables TCPMSS rule was using the wrong formula.

PR #16, opened 2026-03-16
Fix ICMPv6 Packet Too Big ignored by kernel (PTB from self)

v0l:
## Summary

- Fix PMTUD blackhole where Linux kernel ignored ICMPv6 Packet Too Big messages. Add test coverage for the fix.

## Root Cause

When FIPS detects an oversized outbound packet on the TUN interface, it generates an ICMPv6 Packet Too Big and writes it back to TUN so the kernel updates its PMTU cache.

FIPS acts as an on-path router, but shares the same address as the sending application. The PTB source and destination were both the local FIPS address — which is semantically correct (we are the router that rejected the packet), but the kernel ignores PTBs where src == dst.

```
src: fd41:...:local  dst: fd41:...:local  ICMP6, packet too big, mtu 1203  (ignored — src == dst)
```

The PMTU cache was never updated, so TCP kept sending oversized segments, FIPS kept generating ignored PTBs, and connections stalled after the handshake.

## Fix

Use the remote peer's address as the PTB source. This is not strictly accurate (the remote peer didn't reject the packet — we did), but it's the only available address that differs from the local address and makes the kernel honor the PTB:

```
src: fddf:...:remote  dst: fd41:...:local  ICMP6, packet too big, mtu 1203  (honored — src != dst)
```

## Tests

- New unit test `test_build_packet_too_big_remote_source_for_pmtud` in `src/upper/icmp.rs`
- Added PTB source/destination address assertions to existing integration tests `test_tun_outbound_path_mtu_generates_ptb` and `test_multihop_pmtud_heterogeneous_mtu`

### DIFF
--- a/src/node/handlers/session.rs
+++ b/src/node/handlers/session.rs
@@ -1665,7 +1665,6 @@ impl Node {
     /// misconfigured applications sending repeated oversized packets.
     pub(in crate::node) fn send_icmpv6_packet_too_big(&mut self, original_packet: &[u8], mtu: u32) {
         use crate::upper::icmp::build_packet_too_big;
-        use crate::FipsAddress;
         use std::net::Ipv6Addr;
 
         // Extract source address for rate limiting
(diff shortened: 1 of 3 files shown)


jmcorgan:
Merged to master at a5130b3 with minor fixups (whitespace cleanup, debug log field naming).

PR #20, opened 2026-03-19
Add sidecar example with Nostr relay

Origami74:
## Nostr relay sidecar example

Adds a complete example that runs a [strfry](https://github.com/hoytech/strfry) Nostr relay exclusively over the FIPS mesh using the sidecar pattern.

### What's included

- **Docker Compose stack** — two containers sharing a network namespace: the FIPS sidecar (daemon + dnsmasq) and the app (strfry + nginx)
- **Network isolation** — iptables rules lock the app container to the FIPS mesh only; no direct IPv4/IPv6 access to the host or internet
- **DNS resolution** — `.fips` names resolve via the FIPS daemon's built-in DNS, all other queries forward to Docker DNS
- **Build script** — compiles FIPS for Linux (cross-compilation from macOS via cargo-zigbuild) and builds Docker images
- **README** — architecture overview, step-by-step setup, environment variables, troubleshooting, and production guidance

### Other changes

- Renamed `packaging/openwrt/` → `packaging/openwrt-ipk/` and fixed the `package-openwrt.yml` CI workflow to reference the new path
- Added `testing/static/Dockerfile`


jmcorgan (commented):
The test harness refactoring will be removing this particular Dockerfile with the changes to install nak. Can you move that into the example?

Origami74:
Added nak to the sidecar itself for now. Ideally would be in the app container but since it's a different image would have to test more.

PR #21, opened 2026-03-19
Add Arch Linux AUR packaging (fips and fips-git)

dskvr:
## Summary
- Add complete AUR packaging for two packages: `fips` (release, tagged tarball) and `fips-git` (development, git master)
- Both PKGBUILDs are locally validated (makepkg builds, namcap clean)
- Includes sysusers.d/tmpfiles.d system integration, systemd service files, config backup, and post-install messaging
- Full AUR publication instructions in [`packaging/aur/README.md`](packaging/aur/README.md)

### New Files
- `packaging/aur/PKGBUILD` -- release package (builds from tagged tarball)
- `packaging/aur/PKGBUILD-git` -- git package (builds from master)
- `packaging/aur/fips.install` -- pacman post-install/upgrade messaging
- `packaging/aur/fips.sysusers` -- creates `fips` system group via sysusers.d
- `packaging/aur/fips.tmpfiles` -- creates `/run/fips/` via tmpfiles.d
- `packaging/aur/fips.service` -- symlink to `../debian/fips.service`
- `packaging/aur/fips-dns.service` -- symlink to `../debian/fips-dns.service`
- `packaging/aur/README.md` -- step-by-step AUR publication guide

### Modified Files
- `packaging/Makefile` -- added `make aur` target (builds fips-git, runs namcap)
- `packaging/README.md` -- added Arch Linux section

## After Merging

Follow the instructions in `packaging/aur/README.md` to:
1. Set up AUR account and SSH key
2. Push both packages to AUR
3. Verify installation via `yay -S fips-git`
4. Configure GitHub secret (`AUR_SSH_PRIVATE_KEY`) for future CI automation

## Test Plan
- [x] `makepkg -sf -p PKGBUILD-git` builds successfully
- [x] `namcap` reports no errors on PKGBUILD and built package
- [x] Package installs all 3 binaries, 2 services, 2 configs, sysusers, tmpfiles
- [x] `fips.yaml` installed with mode 0600 (sensitive config)
- [ ] Maintainer follows publication instructions to push to AUR (post-merge)
- [ ] `yay -S fips-git` installs successfully on clean Arch system (post-merge)

Requires #24 if automated publishing to AUR is desired.

PR #22, opened 2026-03-19
ci: rename package-openwrt.yml to release.yml

dskvr:
depends on #21 

## Summary
- Rename `package-openwrt.yml` to `release.yml` via `git mv` (clean rename)
- Update workflow display name from "OpenWrt Package" to "Release"
- No functional changes -- all jobs, triggers, and steps are identical

This prepares the workflow for adding AUR publish jobs in a follow-up PR, creating a unified release workflow that handles both OpenWrt .ipk packages and AUR package publishing. Can be easily extended later for other packages.

## Notes
- GitHub Actions UI may show both "OpenWrt Package" (old runs) and "Release" (new runs) entries temporarily. The old entry stops accumulating runs and can be deleted via Actions tab.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,4 @@
-name: OpenWrt Package
+name: Release
 
 on:
   push:


Origami74:
Other PR looks alright, can you provide more info on why this rename is needed though?

dskvr:
@Origami74 not needed

Edit: agent screwed up and didn't include the build step in the workflow, whoops.

PR #27, opened 2026-03-20
Update fips-intro.md

deavmi:
Thought I should correct this. Yggdrasil is zeroconf on any standard Linux (and Mac) machine.

Probably Windows too.

If you have normal setup, you have IPv6 link-local and if you have that you can do multicast discovery. So I want to fix that.


jmcorgan (commented):
I'm not sure this is an actual issue with the present documentation, can you elaborate why the info above is appropriate to add?

deavmi:
Hi,


Agreed, actually. I misread when it said "FIPS *extends* this to other 
mediums"

PR #8, opened 2026-03-13
Add AGENTS.md for repository-specific agent guidance

v0l:
## Summary
- add a repo-level `AGENTS.md` for coding agents working in `fips`
- document the canonical build, lint, unit-test, and integration-test commands, including single-test examples
- capture repo-specific Rust style, error handling, logging, ownership, and documentation expectations derived from the codebase and docs

PR #32, opened 2026-03-26
doc: improve contributor setup guidance

jodobear:
## Summary

- expand `CONTRIBUTING.md` with real contributor prerequisites and a practical local verification ladder
- add explicit guidance for when protocol/config/wire changes should update docs
- add a small contributor routing hint in `README.md` pointing to the design and testing docs

## Why

The repo already has strong protocol design docs and a substantial testing surface, but the public contributor route was still very thin. This patch makes it easier for contributors to understand:

- what they need installed locally
- which checks to run for different change types
- when docs need to be updated alongside code

## Notes

- this is meant to complement PR #8
- I intentionally did not add a tracked `AGENTS.md` here because that surface is already covered by PR #8

## Verification

- `git diff --check`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,17 +2,64 @@
 
 ## Getting Started
 
-Clone the repo and verify your setup:
+Clone the repo:
 
 ```
 git clone https://github.com/jmcorgan/fips.git
 cd fips
-cargo build
-cargo test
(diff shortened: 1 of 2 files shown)


jmcorgan:
Squashed both commits and rebased onto current master. Merged as 4bc30d2.

PR #38, opened 2026-03-28
ci: publish Linux release artifacts

jodobear:
# Summary

  - add Linux x86_64 and aarch64 artifact builds to the release workflow
  - rename `package-openwrt.yml` to `packages.yml`
  - make the tarball and `.deb` packaging scripts target-aware so packaging can reuse built outputs
  - publish a unified `checksums.txt` in the tag-driven release job

  ## Notes

  - Linux artifacts keep the default feature set, including BLE
  - the GitHub release step remains tag-only and still owns release publication in one place
  - OpenWrt artifact naming and release flow are preserved

  ## Validation

  - local Fedora smoke test passed for:
    - `cargo build --release`
    - tarball packaging
    - `.deb` packaging
  - manual `workflow_dispatch` run of `Packages` on `33-artifacts` succeeded
  - `git diff --check`

### DIFF
--- a/.github/workflows/package-linux.yml
+++ b/.github/workflows/package-linux.yml
@@ -1,0 +1,196 @@
+name: Linux Package
+on:
+  push:
+    tags:
+      - "*"
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  determine-versioning:
+    runs-on: ubuntu-latest
+    outputs:
+      linux_package_version: ${{ steps.linux_version.outputs.linux_package_version }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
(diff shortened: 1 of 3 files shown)


jmcorgan:
@Origami74

jmcorgan:
The tag trigger uses `"*"` which will fire on any tag push, not just version tags. Can you tighten this to `"v*"` to match the release convention? The build jobs don't need to run for non-version tags. The release job already gates on `startsWith(github.ref, 'refs/tags/')` but we'd still waste runner time on the builds.

jmcorgan (commented):
I'll go ahead and make the change myself and commit this.

jmcorgan:
Squash-merged to master as `5d27efb` with tag trigger tightened to `v*`. Thanks for the contribution.

Issue #39, opened 2026-03-29
Stabilize flaky rekey integration test baseline convergence


jodobear:
`Integration (rekey)` appears to be flaky due to a strict one-shot baseline connectivity check before rekeying starts, rather than a deterministic rekey bug.

  ## What happened

  On branch CI for `33-artifacts` , the failing run was in `testing/static/scripts/rekey-test.sh` during Phase 1:

  - fixed wait: `CONVERGE_WAIT=15`
  - baseline check: all 20 directed ping pairs
  - observed failure: a single early miss (`A -> C`)

  The same job later passed on rerun without code changes. In the passing rerun:

  - pre-rekey baseline: 20/20 passed
  - first rekey cycle: passed
  - post-first-rekey connectivity: 20/20 passed
  - second rekey cycle: passed
  - post-second-rekey connectivity: 20/20 passed
  - log analysis: passed

  That points to test brittleness around initial convergence, not an obvious rekey logic failure.

  ## Repo references

  - `testing/static/scripts/rekey-test.sh`
  - Phase 1 uses a fixed sleep before a strict full-matrix assertion

  ## Suggested direction

  Instead of taking one snapshot after a fixed `15s` wait, have the test wait until full baseline connectivity is reached up to a timeout, and only fail if convergence
  never completes within that window.

  That should reduce flakiness without weakening the rekey assertions that follow.

jmcorgan:
Fixed in 8e38d88 (master). Rekey test baseline now uses bounded convergence polling instead of fixed sleep.

PR #40, opened 2026-03-29
testing: stabilize rekey integration timing

jodobear:
## Summary

Per https://github.com/jmcorgan/fips/issues/39#issue-4163882734

  - replace the rekey test's one-shot baseline check with bounded convergence polling
  - make the first FMP rekey observation informational instead of a hard timing gate
  - wait boundedly for late FSP log evidence before final log assertions

  ## Why

  The rekey integration test was flaky for two timing reasons:

  - baseline connectivity could be checked too early
  - final FSP log assertions could run before the expected log markers appeared

  This keeps the test strict where it should be strict:
  - full baseline connectivity is still required
  - mid-test and post-second-rekey connectivity still require 20/20 success
  - final log analysis still requires actual FMP/FSP rekey evidence and zero bad events

  ## Validation

  - `bash -n testing/static/scripts/rekey-test.sh`
  - `git diff --check master...HEAD`
  - local Docker rekey run passed end to end: `69 passed, 0 failed`

### DIFF
--- a/testing/static/scripts/rekey-test.sh
+++ b/testing/static/scripts/rekey-test.sh
@@ -19,6 +19,7 @@
 set -e
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "$SCRIPT_DIR/../../lib/wait-converge.sh"
 TOPOLOGY="rekey"
 NODES="a b c d e"
 
@@ -58,12 +59,17 @@ fi
 trap 'echo ""; echo "Test interrupted"; exit 130' INT
 
 # Wait times derived from rekey timer
-CONVERGE_WAIT=15
-FIRST_REKEY_WAIT=40    # > REKEY_AFTER_SECS, allow margin
+BASELINE_CONVERGENCE_TIMEOUT=36
 REKEY_SETTLE=5         # settle time after rekey for cutover to complete
+# First FMP rekey should follow shortly after the 35s interval once the mesh is
+# fully converged. Keep this bounded to preserve a meaningful scheduling check
+# while still allowing for log visibility at the timeout edge.
+FIRST_REKEY_TIMEOUT=$((REKEY_AFTER_SECS + 15))
 SECOND_REKEY_WAIT=40   # wait for second cycle
+LOG_EVENT_POLL_INTERVAL=1
 
 TIMEOUT=5
+CONVERGENCE_PING_TIMEOUT=1
 PASSED=0
 FAILED=0
 TOTAL_PASSED=0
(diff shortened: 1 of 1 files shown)


jmcorgan:
Thanks for this. The baseline convergence fix is the right approach, polling with a bounded timeout instead of a fixed sleep directly addresses the flakiness we saw.

A couple of concerns with the scope though.

The Phase 2 FMP rekey observation was downgraded from a hard assertion to informational. The original test required at least one FMP rekey cutover after the first rekey window, and I think that should stay as a hard gate. If FMP rekey doesn't fire within REKEY_AFTER_SECS + 15s of a fully converged mesh, that's not a timing fluke, it means something is wrong with rekey scheduling. The issue was specifically about baseline convergence timing, not about relaxing the rekey assertions themselves.

The FSP log waits with `|| true` before the final assertions are fine since the subsequent `assert_min_count` calls are still enforcing the actual counts. But combined with the Phase 2 change, it starts to feel like the test is being made to pass rather than made to converge faster.

I'd be happy to merge this with Phase 2 restored to a hard failure if no FMP cutover is observed within the timeout. The rest of the changes (wait_for_full_baseline, wait_for_peers pre-check, dump_peer_connectivity on failure, shorter ping timeout for polling) all look good.

jodobear:
The baseline convergence fix is the real issue here, and relaxing Phase 2 widened the change beyond that scope.

I'll restore the first FMP rekey observation to a hard assertion at `REKEY_AFTER_SECS + 15`, and keep the baseline convergence polling plus the bounded FSP log waits before the final assertions.

jodobear:
Quick status update on the current branch state:

  - the branch is now a single commit
  - Phase 2 is hard again
  - the test change stays scoped to bounded baseline convergence plus the small timeout-edge fix in `wait_for_log_pattern_count()`

  I compared the latest runs on the fork:

  - `23796940591` on `3826cfd`: full `CI` passed, including `Integration (rekey)`
  - `23803766380` on `c488403`: `CI` failed, but not on the original Phase 1 issue

  In the latest failed run:
  - `Integration (rekey)` passed Phase 1 baseline `20/20`
  - Phase 2 also passed
  - the failure was later in Phase 5, with only `B -> E` and `E -> B` failing
  - final Phase 6 log analysis still passed

  The same overall run also had an `Integration (static-mesh)` failure (`A -> C`), so this does not look like a clean regression isolated to the baseline convergence
  change.

  So from the current evidence:
  - the original baseline-convergence failure mode looks improved
  - the latest red run appears to be a different integration flake
  - I have not widened this PR beyond the baseline-convergence fix

jmcorgan (approved):
Local CI passes: rekey test 70/70, all phases green. Baseline convergence polling and Phase 2 hard assertion look good. Approving.

jmcorgan:
Merged via rebase in 8e38d88. GitHub didn't auto-close due to SHA mismatch from rebase.

PR #42, opened 2026-04-05
Add diagnostic queries for security validation and mesh debugging

osh:
Extend the fipsctl control query interface with visibility into internal state critical for protocol security auditing, mesh troubleshooting, and operational monitoring.

New command:

  fipsctl show identity-cache

    Lists every node identity cached by the daemon (learned from DNS
    resolution, peer handshakes, sessions, and static config).  Shows
    npub, IPv6 address, display name, and LRU age alongside the
    configured cache capacity.

Extended queries:

  show peers — Noise session counters (send_counter, highest received
    counter) for rekey urgency assessment.  Per-peer replay suppression
    and consecutive decrypt failure counts for active attack detection.
    Session index visibility for hijack analysis.  Rekey lifecycle
    state (in_progress, draining, K-bit epoch).

  show sessions — Handshake resend count during establishment for
    connectivity debugging.  Rekey and session health fields
    (session_start, K-bit, coords warmup, drain state) when
    established.

  show cache — Individual coordinate cache entries with tree
    coordinates, depth, path MTU, and age.  Enables route-level
    debugging by showing exactly which destinations have cached
    routes and via what tree path.  Renames the top-level count
    field from "entries" to "count" for clarity.

  show routing — Pending discovery lookups expanded from count to
    per-target detail (attempt number, age, last sent).  Pending
    TUN packet queue depth for backpressure visibility.  Connection
    retry state per peer (retry count, next attempt, auto-reconnect
    flag).

Updates fipstop to match the revised show_cache and show_routing response schemas.  Updates README monitoring section with the complete fipsctl command list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,13 +194,18 @@ ssh   npub1bbb....fips
 Use `fipsctl` to query a running node:
 
 ```bash
-fipsctl show status       # Node status overview
-fipsctl show peers        # Authenticated peers
-fipsctl show links        # Active links
-fipsctl show tree         # Spanning tree state
-fipsctl show sessions     # End-to-end sessions
-fipsctl show transports   # Transport instances
-fipsctl show routing      # Routing table summary
+fipsctl show status         # Node status overview
+fipsctl show peers          # Authenticated peers and security state
+fipsctl show links          # Active links
+fipsctl show tree           # Spanning tree state
+fipsctl show sessions       # End-to-end sessions and rekey health
+fipsctl show bloom          # Bloom filter state
+fipsctl show mmp            # MMP metrics summary
+fipsctl show cache          # Coordinate cache entries and routes
+fipsctl show connections    # Pending handshake connections
+fipsctl show transports     # Transport instances
+fipsctl show routing        # Routing, discovery, and retry state
+fipsctl show identity-cache # Known node identities (npubs)
 ```
 
 `fipstop` provides an interactive TUI dashboard with live-updating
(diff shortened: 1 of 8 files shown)


jmcorgan:
Code review looks good overall — no security issues, follows existing patterns, and the diagnostic additions are useful.

Two things to address before merging:

**Breaking schema changes:** `show_cache` renames the `"entries"` count field to `"count"` and reuses `"entries"` for the detail array. Similarly, `show_routing` changes `pending_lookups` from an integer to an array. Both break existing consumers. Suggest keeping the original fields as-is and adding the new data under new names (e.g. `"cache_details"` and `"pending_lookup_details"`). That way existing scripts don't break.

**Visibility:** `PendingLookup` is changed from `pub(crate)` to `pub` — `pub(crate)` is sufficient since the only consumer is within the crate.

jmcorgan:
Thanks for this — the diagnostic surface you added is exactly the right set of fields for security auditing and mesh debugging. I pulled the work in-house to finish it out rather than let the PR age further.

Merged on master as `2d342a4` (you are the commit author, with me as co-author for the rebase/integration). The content is identical to what you proposed here; the extra churn was just because master had moved 25 commits forward (project-wide `cargo fmt` and a restructure of `show_peers` into an iterator chain) and a straight cherry-pick hit conflicts on the reformatted regions.

CHANGELOG entry is under Unreleased → Operator Tooling, and the two response-schema breaks (`show_cache.entries` and `show_routing.pending_lookups` changing from scalars to arrays) are called out under Changed.

Closing this PR in favor of the merged commit. Appreciate the work.

Issue #44, opened 2026-04-07
Windows platform support


0ceanSlim:
I've been running FIPS on Windows and have a working implementation of cross-platform support. Before 
  submitting a PR, I wanted to discuss the approach and trade-offs.

  ## What's implemented

  - **cfg-gating**: Platform-specific dependencies (`tun`, `libc`, `rtnetlink`, `bluer`) moved under    
  `[target.'cfg(unix)'.dependencies]` so the codebase compiles on Windows without BLE/libdbus
  - **TUN device**: Windows TUN support via the `wintun` crate (userspace packet I/O with ring buffers, 
  configured via `netsh`)
  - **Control socket**: TCP listener on `localhost:21210` (see trade-offs below)
  - **UDP socket**: Platform abstraction — Linux preserves `recvmsg()`/`SO_RXQ_OVFL` kernel drop        
  counting; Windows uses `tokio::net::UdpSocket` (drops always 0)
  - **Windows service**: Install/uninstall/run via `windows-service` crate (`--install-service`,        
  `--uninstall-service`, `--service` flags)
  - **Tests**: Platform-gated tests for key files, control socket, UDP socket, default paths

  ## Trade-offs and known limitations

  ### Control socket (TCP vs Unix domain socket)

  The current implementation uses TCP on localhost since Windows doesn't support Unix domain sockets    
  reliably. This means any local user can connect to the control port — no filesystem-level ACLs like   
  the Unix socket provides.

  Alternatives worth considering:
  1. **Named pipes** — The idiomatic Windows IPC mechanism. Supports Windows DACLs for access control.  
  Most Windows services use this (Docker, PostgreSQL, etc.)
  2. **AF_UNIX sockets** — Windows 10 1803+ has support, though with some limitations. Would keep the   
  code more unified across platforms.
  3. **TCP with auth token** — Keep TCP but add a shared-secret handshake

  TCP is the simplest approach to get Windows support in the door. A future improvement could migrate to
   named pipes, which would relate to #35 (access control lists).

  ### Other limitations

  - BLE transport requires libdbus (Linux-only) — Windows builds use `--no-default-features --features  
  tui`
  - `wintun.dll` must be in the executable directory or PATH
  - TUN creation requires Administrator privileges
  - Kernel drop counting not available on Windows (always reports 0)
  - The 56 node simulation tests have pre-existing timing issues on Windows (not caused by these        
  changes)

  ## Build verification

  - `cargo build --no-default-features --features tui` 
[...]

PR #45, opened 2026-04-07
Windows platform support

0ceanSlim:
## Summary

  - Gate platform-specific dependencies and code for cross-platform compilation
  - Add Windows TUN device support using the `wintun` crate
  - Add TCP localhost control socket for Windows (Unix domain sockets not available)
  - Add Windows service support (install/uninstall/run-as-service)
  - Add Windows platform tests and update documentation

  Addresses #44 

  ## Commits

  1. **Gate platform-specific dependencies and code** — Moves `tun`/`libc`/`rtnetlink`/`bluer` under    
  `cfg(unix)`, adds platform-conditional compilation throughout
  2. **Add Windows TUN device support** — `wintun` crate with ring buffer session API, `netsh` for      
  interface configuration
  3. **Add cross-platform control socket** — TCP listener on `localhost:21210` for Windows, shared      
  `handle_connection_generic` protocol handler, port validation in `fipsctl`/`fipstop`
  4. **Add Windows service support** — Refactors `main()` into shared `run_daemon()`, adds
  `--install-service`/`--uninstall-service`/`--service` CLI flags
  5. **Add Windows platform tests and documentation** — 7 new platform-gated tests, README Windows      
  section, configuration reference updates

  ## Known limitation: TCP control socket

  The Windows control socket uses TCP on localhost rather than Unix domain sockets. This means any local
   user can connect — no filesystem-level ACLs. See issue #XX for discussion of alternatives (named     
  pipes, AF_UNIX, auth tokens). TCP is used here as the simplest working approach.

  ## Test plan

  - [x] `cargo build --no-default-features --features tui` (Windows)
  - [x] `cargo test --no-default-features --features tui` — 915 passed, 0 failed
  - [x] `cargo clippy --no-default-features --features tui -- -D warnings` — clean
  - [ ] Verify Linux build is not broken (CI or WSL)
  - [x] Manual test: run daemon with TUN (requires Administrator + wintun.dll)
  - [x] Manual test: `fipsctl show status` connects via TCP

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,12 +47,21 @@ jobs:
           - os: ubuntu-latest
           - os: ubuntu-24.04-arm
           - os: macos-latest
+          - os: windows-latest
 
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set SOURCE_DATE_EPOCH from git
+      - name: Set SOURCE_DATE_EPOCH from git (Unix)
+        if: runner.os != 'Windows'
         run: echo "SOURCE_DATE_EPOCH=$(git log -1 --format=%ct)" >> "$GITHUB_ENV"
+
+      - name: Set SOURCE_DATE_EPOCH from git (Windows)
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          $epoch = git log -1 --format=%ct
+          echo "SOURCE_DATE_EPOCH=$epoch" >> $env:GITHUB_ENV
 
       - name: Install system dependencies (Linux only)
         if: runner.os == 'Linux'
@@ -73,7 +82,7 @@ jobs:
             ${{ runner.os }}-cargo-
 
       - name: Build
-        run: cargo build --release ${{ runner.os == 'macOS' && '--no-default-features --features tui' || '--features gateway' }}
+        run: cargo build --release ${{ (runner.os == 'macOS' || runner.os == 'Windows') && '--no-default-features --features tui' || '--features gateway' }}
 
       - name: SHA-256 hashes (Linux)
         if: runner.os == 'Linux'
@@ -82,6 +91,11 @@ jobs:
       - name: SHA-256 hashes (macOS)
         if: runner.os == 'macOS'
         run: shasum -a 256 target/release/fips target/release/fipsctl target/release/fipstop
+
+      - name: SHA-256 hashes (Windows)
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: Get-FileHash target\release\fips.exe, target\release\fipsctl.exe, target\release\fipstop.exe -Algorithm SHA256
 
       # Upload the Linux binary so integration jobs can use it without rebuilding
       - name: Upload Linux binary
@@ -164,6 +178,35 @@ jobs:
 
       - name: Set SOURCE_DATE_EPOCH from git
         run: echo "SOURCE_DATE_EPOCH=$(git log -1 --format=%ct)" >> "$GITHUB_ENV"
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Cache Cargo registry + build
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-
+
+      - name: Install cargo-nextest
+        uses: taiki-e/install-action@nextest
+
+      - name: Run unit tests
+        run: cargo nextest run --all --profile ci --no-default-features --features tui
+
+# ─────────────────────────────────────────────────────────────────────────────
+# Job 2c – Unit tests (Windows)
+# ─────────────────────────────────────────────────────────────────────────────
+  test-windows:
+    name: Unit tests (Windows)
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
 
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/package-windows.yml
+++ b/.github/workflows/package-windows.yml
@@ -1,0 +1,158 @@
+name: Windows Package
+on:
+  push:
+    branches:
+      - master
+      - maint
+      - next
+    tags:
+      - "v*"
+  pull_request:
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  determine-versioning:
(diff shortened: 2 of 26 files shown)


Origami74:
We just merged the #46 PR for MacOS support. Can you rebase on that branch?

0ceanSlim:
> We just merged the #46 PR for MacOS support. Can you rebase on that branch?

on it! I was patiently watching your commits today in preparation

0ceanSlim:
@jmcorgan I think it's ready for review.

jmcorgan (changes requested):
Thanks for this — solid work bringing Windows support to FIPS, and the platform separation follows the pattern from #46 well.

**Items to address before merge:**

**1. `send_request_over_stream` removes `shutdown(Write)` and timeouts.** The refactored function drops the half-close and read/write timeouts from the Unix path. On a hung daemon, `fipsctl` will now block indefinitely — this is a regression on all platforms. Please restore the shutdown and timeouts in both the Unix and TCP paths.

**2. `DEFAULT_HOSTS_PATH` hardcoded to `/etc/fips/hosts`.** No platform conditional, so the host map feature is silently non-functional on Windows and error messages reference a Unix path. Please make this `cfg`-conditional — `C:\ProgramData\fips\hosts` on Windows to match the packaging scripts.

**3. Duplicated startup logic in `run_service()`.** ~70 lines of daemon startup copy-pasted from `run_daemon()`. This will drift. Refactor `run_daemon()` to accept an optional shutdown signal so the service path can reuse it instead of duplicating the sequence.

**4. TCP control socket security must be documented.** The Unix socket uses file permissions for access control; the TCP socket allows any local user to issue control commands including `disconnect` and `inject-config`. Please document this as a known limitation in the configuration reference, and note named pipe or auth token support as a future improvement.

**Recommendations (non-blocking):**

- **Deduplicate `ControlClient::new()`** — both cfg-gated methods are identical, should be one unconditional method.
- **Extract shared `default_socket_path()`** — copy-pasted in three files, extract to `fips::config::default_control_path()`.
- **Preserve error source in `TunError::Create`** — the conversion to `String` loses the error chain. Consider `Box<dyn Error>` or `#[source]`.
- **Defensive `u16` cast** — `packet.len() as u16` in `allocate_send_packet` should use `u16::try_from()`.
- **Extract shared TUN reader logic** — the Windows `run_tun_reader` reimplements the packet processing loop from Unix. Extract common processing into a shared function.
- **CI `test-windows` rebuilds from source** despite declaring a dependency on `build`. Either download the artifact or remove the dependency.

Overall the architecture is sound. Looking forward to the update.

0ceanSlim:
Thanks for the review! All four required items have been addressed:                                   

  **1. `send_request_over_stream` — timeouts and shutdown restored**
  Both the Unix and TCP paths now set 5-second read/write timeouts and call `shutdown(Write)` after     
  sending the request (`fipsctl.rs` lines 145-151, 185-191).

  **2. `DEFAULT_HOSTS_PATH` — platform-conditional**
  `cfg(unix)` → `/etc/fips/hosts`, `cfg(windows)` → `C:\ProgramData\fips\hosts` (in `upper/hosts.rs`).  

  **3. `run_service()` — deduplication via `run_daemon()`**
  `run_daemon()` now accepts a `shutdown_signal: impl Future<Output = ()>` parameter. Both the
  foreground path and the Windows service path call into it with their respective shutdown signals — no 
  duplicated startup logic.

  **4. TCP control socket security — documented**
  Added a security note in `docs/design/fips-configuration.md` (lines 65-77) covering: the limitation,  
  specific risky commands, single-user vs shared machine guidance, future improvements (named pipes,    
  auth tokens), and firewall mitigation.

For the non-blocking recommendations:                                                                                                        
 **ControlClient::new() dedup** — Done. `new()` is now a single unconditional method. `connect()`      
  remains cfg-gated because the return types differ (`UnixStream` vs `TcpStream`).

  **default_socket_path() extraction** — `fipsctl` and `fipstop` both delegate to
  `fips::config::default_control_path()`. `ControlConfig::default_socket_path()` in `config/node.rs`    
  still has its own implementation since it returns `String` for serde defaults — happy to have it call 
  through to `default_control_path()` as well if you'd prefer full consolidation.

  **TunError::Create error source** — Done. Now uses `#[source] Box<dyn std::error::Error + Send +      
  Sync>`.

  **Defensive u16 cast** — Done. `allocate_send_packet` uses `u16::try_from(packet.len())` with a       
  warning and `continue` on overflow.

  **Shared TUN reader logic** — Extracted `tun_reader_setup()` and `handle_tun_packet()` as shared      
  helpers called by all three platform-specific readers. The per-platform loop scaffolding remains      
  separate since the I/O primitives differ (`tun::AsyncDevice` vs wintun session).

  **CI test-windows rebuild** — Removed the `needs: [build]` dependency so it runs independently. The   
  build job only uploads Linux artifacts, so there's nothing to download for Windows.

jmcorgan (approved):
Thorough review complete. The architecture is clean — the platform abstractions are well-placed and the common code changes are minimal and correct. Approving.

**Before merge:** Please squash your 9 commits into a single commit with a coherent message covering the full scope (your tooling will produce a better result than GitHub's auto-squash concatenation), and run `cargo fmt` — the format check is currently failing.

**Non-blocking notes (can be addressed in a follow-up):**

1. **Hardcoded Unix paths in user-facing text.** Several CLI help strings and doc comments reference `/etc/fips/hosts` unconditionally — these show Unix paths to Windows users:
   - `src/bin/fipsctl.rs` lines 56, 65 (`--connect` / `--disconnect` help text)
   - `src/node/mod.rs` lines 444, 919 (doc comments on host map and resolve_peer_name)

2. **README config path.** The Windows section says to place `fips.yaml` in `%APPDATA%\fips\`, but the service installer uses `%ProgramData%\fips\` and the service reads config via `FIPS_CONFIG`. A sentence clarifying the distinction (per-user foreground vs system service) would help.

3. **`gateway/control.rs`** has a hardcoded `GATEWAY_SOCKET_PATH = "/run/fips/gateway.sock"` with no platform gating. Not a real issue today since the gateway feature depends on `rustables` (Linux nftables), but worth noting if gateway ever becomes cross-platform.

0ceanSlim:
Squashed and the code is cargo fmt clean locally. The CI fmt job fails because rust-toolchain.toml pins 1.94.1 but the fmt step installs dtolnay/rust-toolchain@stable with rustfmt — at runtime cargo fmt uses the pinned 1.94.1 toolchain which doesn't have rustfmt installed. This is a pre-existing CI issue, not introduced by this PR. Happy to fix it here or in a follow-up, whichever you prefer.

jmcorgan (commented):
Regarding the rustfmt issue, yeah, I sort of borked things up with the way I did the pinning (for build reproducibility) and rustfmt addition. Not entirely sure how to fix but it will need to go onto maint branch, then get merged up to master, then merged up to next.

PR #46, opened 2026-04-07
Add macOS support: Ethernet/BPF, utun TUN, packaging, graceful shutdown

Origami74:
## Add macOS support

Brings macOS to feature parity with Linux: Ethernet transport via BPF,
TUN interface via utun, `.pkg` installer, and graceful shutdown under launchd.

## Summary

- Add Ethernet transport for macOS using BPF (`/dev/bpf*`)
- Add TUN support via utun devices with `ifconfig`/`route` configuration
- Fix graceful shutdown under launchd (SIGTERM handling, immediate Ethernet teardown)
- Add `.pkg` installer and launchd service plist
- Fold macOS CI into main workflow; add macOS package release workflow

## Key design decisions

**Ethernet socket split into `socket_linux.rs` and `socket_macos.rs`**

The two implementations are fundamentally different mechanisms — AF_PACKET sockets
on Linux vs BPF character devices on macOS — with no shared I/O logic. A single
file with `#[cfg]` blocks throughout would obscure both. Each file implements the
same `PacketSocket` API; `socket.rs` re-exports `PacketSocket` and provides the
platform-specific `AsyncPacketSocket` wrapper on top.

BPF also has different read semantics: a single `read()` may return multiple
frames, each prefixed with a `struct bpf_hdr`. `parse_next_frame()` in
`socket_macos.rs` walks these with `BPF_WORDALIGN` and is unit-tested directly.

**BPF reader thread instead of AsyncFd**

BPF character devices don't support kqueue, so `AsyncFd` doesn't work on macOS.
A single dedicated reader thread loops on `select()` and feeds frames through a
`tokio::sync::mpsc` channel. This avoids the ~84 Mbps throughput cap that came
from wrapping each packet in `spawn_blocking`.

**TUN differs significantly from Linux**

On Linux, TUN devices have user-assigned names (`fips0`), are configured via
netlink (rtnetlink), and write raw IPv6 packets directly. On macOS:

- utun devices are kernel-assigned (`utun0`, `utun1`, ...); the name is read back
  after creation and the user-configured name is ignored.
- Interface setup uses `ifconfig`/`route` instead of netlink; rtnetlink is now a
  Linux-only dependency.
- Every write to a utun fd must be prefixed with a 4-byte packet-information
  header (`[0, 0, 0, 30]` for AF_INET6). The TUN writer uses `writev` for
  scatter-gather to prepend this without an extra allocation per packet.
- The `tun` crate's `Read` impl handles stripping the PI header on reads, so
  `read_packet` is uniform across platforms.
- utun devices are destroyed automatically when their file descriptor is closed
  (no explicit deletion). `platform::delete_interface` on macOS brings the
  interface down with `ifconfig down` rather than deleting it; the actual teardown
  happens when `TunDevice` drops and the fd is closed.

The platform differences are isolated in a `platform` submodule inside `tun.rs`
(Linux: netlink; macOS: ifconfig/route), keeping the shared TUN logic untouched.

**Self-pipe for shutdown signaling**

On macOS, `close(fd)` from thread A does not interrupt a blocked `read(fd)` in
thread B (unlike Linux). Both the BPF reader and TUN reader use a `libc::pipe()`
self-pipe: the reader `select()`s on both the data fd and the pipe read end;
`request_shutdown()` writes a byte to the pipe write end to wake it immediately.

**No `task.await` after `task.abort()` on macOS**

On a `current_thread` tokio runtime, awaiting an aborted task that is blocked on
a channel fed by an OS thread deadlocks — the task needs to be polled, but we hold
the only thread. The Ethernet shutdown path skips `task.await` on macOS.

**SIGTERM handling**

`launchctl unload` sends SIGTERM, not SIGINT. The daemon now handles both signals,
so `launchctl unload` shuts down cleanly instead of hanging.

## Test plan

- [ ] `cargo build` on macOS arm64 and x86_64 — no warnings
- [ ] `cargo test` on macOS — all tests pass
- [ ] Install `.pkg` and start: `sudo launchctl load -w /Library/LaunchDaemons/com.fips.daemon.plist`
- [ ] Unload cleanly: `sudo launchctl unload /Library/LaunchDaemons/com.fips.daemon.plist` (no hang)
- [ ] Ctrl+C shutdown is immediate when Ethernet transport is active
- [ ] Ethernet transport connects to peers and passes traffic on macOS
- [ ] Linux CI (build + unit + integration) unaffected

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,8 +23,7 @@ env:
 # ─────────────────────────────────────────────────────────────────────────────
 # Job 1 – Build matrix
 #
-# Builds on Linux x86_64 and Linux aarch64.  macOS and Windows are in a
-# separate ci-compat.yml workflow so their failures don't mark this run red.
+# Builds on Linux x86_64, Linux aarch64, and macOS.
 # ─────────────────────────────────────────────────────────────────────────────
 jobs:
   build:
@@ -37,6 +36,7 @@ jobs:
         include:
           - os: ubuntu-latest
           - os: ubuntu-24.04-arm
+          - os: macos-latest
 
     steps:
       - uses: actions/checkout@v4
@@ -44,7 +44,8 @@ jobs:
       - name: Set SOURCE_DATE_EPOCH from git
         run: echo "SOURCE_DATE_EPOCH=$(git log -1 --format=%ct)" >> "$GITHUB_ENV"
 
-      - name: Install system dependencies
+      - name: Install system dependencies (Linux only)
+        if: runner.os == 'Linux'
         run: sudo apt-get update && sudo apt-get install -y libdbus-1-dev
 
       - name: Install Rust toolchain
@@ -62,10 +63,15 @@ jobs:
             ${{ runner.os }}-cargo-
 
       - name: Build
-        run: cargo build --release
+        run: cargo build --release ${{ runner.os == 'macOS' && '--no-default-features --features tui' || '--features gateway' }}
 
-      - name: SHA-256 hashes
+      - name: SHA-256 hashes (Linux)
+        if: runner.os == 'Linux'
         run: sha256sum target/release/fips target/release/fipsctl target/release/fipstop target/release/fips-gateway
+
+      - name: SHA-256 hashes (macOS)
+        if: runner.os == 'macOS'
+        run: shasum -a 256 target/release/fips target/release/fipsctl target/release/fipstop
 
       # Upload the Linux binary so integration jobs can use it without rebuilding
       - name: Upload Linux binary
@@ -117,7 +123,7 @@ jobs:
         uses: taiki-e/install-action@nextest
 
       - name: Run unit tests
-        run: cargo nextest run --all --profile ci
+        run: cargo nextest run --all --profile ci --features gateway
 
       - name: Publish test report (Checks tab)
         uses: dorny/test-reporter@v2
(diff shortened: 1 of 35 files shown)


jmcorgan:
Starting review of this PR soon. My primary focus will be verifying that macOS-specific changes are properly isolated behind cfg gates and that any cross-platform changes are justified as bug fixes or enhancements rather than behavioral changes on Linux.

Also heads up — @amperstrand has found some potential BLE-related bugs during his ESP32 porting work that may touch overlapping code, so we may need to coordinate there before merging.

@0ceanSlim FYI — once this lands I'll move on to reviewing your Windows PR next.

jmcorgan (changes requested):
The platform abstraction approach is clean and the BPF implementation is well done. Two items I'd like addressed before merge, plus some recommendations.

## Needs fixing

**EAGAIN detection via string matching (macOS TUN reader)**

The `TunError` wraps errors as strings, so the reader loop checks for EAGAIN like this:

```rust
if err_str.contains("would block") || err_str.contains("Resource temporarily unavailable") {
    break;
}
```

String matching is fragile across locales and OS versions. Better to either expose `io::Error::kind()` from `read_packet()` so the caller can check `ErrorKind::WouldBlock`, or check `io::Error::last_os_error().raw_os_error() == Some(libc::EAGAIN)` directly in the reader loop.

**`bluer` dependency not Linux-gated**

`bluer` is under general `[dependencies]` (optional, behind `ble` feature). It depends on D-Bus which doesn't exist on macOS. If someone runs `cargo build --features ble` on macOS they get a confusing bluer build failure. Move it to `[target.'cfg(target_os = "linux")'.dependencies]` so the error is clear.

## Recommendations (not blocking)

- **Launchd auto-start on install**: The postinstall script calls `launchctl bootstrap` which starts the daemon immediately with default config before the user has configured it. Consider removing `bootstrap` from postinstall or adding `Disabled: true` to the plist.

- **Shutdown fd cleanup**: `shutdown_fd` is manually closed in three places. A small RAII wrapper would prevent fd leak bugs if new error paths are added later.

- **Ethernet module doc comment**: Still says "AF_PACKET sockets with SOCK_DGRAM" — should mention BPF on macOS.

- **BPF reader thread join on drop**: `AsyncPacketSocket` stores the reader thread handle but never joins it on drop. Low risk in practice but could matter if the socket is dropped and immediately recreated.

Origami74:
🔥

0ceanSlim:
now it is on me!

PR #48, opened 2026-04-08
Added instructions on how to use fips without sudo.

sh1ftred:
Just a README.md udpate.


jmcorgan:
This should probably be branched off `macos-support` (PR #46) rather than `macos-ble`, since that's where the macOS platform work lives. Worth rebasing onto that branch and retargeting the PR.

sh1ftred:
@Origami74 ?

Issue #35, opened 2026-03-27
Pubkey access control lists (allow/deny)


jmcorgan:
## Background

Any node that knows a peer's transport address can initiate a Noise handshake and become a link-layer peer. There is no mechanism to restrict which public keys a node will accept connections from.

This matters for:
- **Private mesh segments**: only pre-approved nodes should join (home, office, infrastructure)
- **Abuse mitigation**: block specific misbehaving nodes without changing transport addresses
- **Selective peering**: an operator running a well-connected node may want to limit who can consume its bandwidth and routing resources

## Proposal

Add optional access control lists that filter peer connections by npub, modeled on the Unix `hosts.allow` / `hosts.deny` pattern. Two files, evaluated in order:

1. **`/etc/fips/peers.allow`** — if non-empty, only listed npubs are permitted
2. **`/etc/fips/peers.deny`** — listed npubs are rejected

Evaluation logic (same as TCP Wrappers):
- If the npub matches an entry in `peers.allow` → **permit**
- Else if the npub matches an entry in `peers.deny` → **reject**
- Else → **permit** (default-open)

For a closed mesh (default-deny), the operator adds a wildcard `ALL` entry to `peers.deny` and explicitly lists permitted npubs in `peers.allow`.

### File format

One npub per line, comments with `#`:

```
# peers.allow — only these nodes may connect
npub1u33pdnwuaf2nmv4j64ghaycg7ucud9vyylexd20gwqlmg3njnuusspd8px
npub1hs4ju064yed7dkp7jz2mjprfsyvvznfmucj9ew0gc25zhvpvtslsqa4m5l
npub1nqjupzg8x35qgjhc54wq2ec7jl6ysmfpad7p0jxdyq3ck432utmscqk5de
```

```
# peers.deny — block these nodes, allow all others
npub1bad0actor...
```

```
# peers.deny — default-deny (use with peers.allow)
ALL
```

### Enforcement point

Access control is checked after the Noise handshake reveals the initiator's identity:
- **IK pattern** (current FMP): responder learns initiator pubkey from msg1 — reject before sending msg2
- **XX pattern** (post-forklift): responder learns initiator pubkey from msg3 — reject after handshake completes but before link activation

For outbound connections, the check is also applied after the responder's identity is learned, preventing the local node from completing a handshake with a denied peer.

### Runtime reload

[...]

PR #50, opened 2026-04-08
Add peer ACL enforcement and allowlist coverage

alexxie16:
## Summary
- add peer ACL enforcement using `/etc/fips/peers.allow` and `/etc/fips/peers.deny`, with allow-first precedence, wildcard support, alias-aware entries, reload-on-change behavior, and denial checks on outbound connect plus handshake paths
- add `fipsctl acl show` support so operators can inspect the loaded ACL, including the original file entries (`allow_file_entries` / `deny_file_entries`) and the resolved effective peer identities (`allow_entries` / `deny_entries`)
- add the six-node `testing/acl-allowlist/` Docker harness, keep its naming/docs aligned with the runtime behavior, and run it in CI so the end-to-end ACL path is exercised before merge
- serialize the `100_nodes` UDP integration tests behind a shared lock and keep the shared Linux test build on the repository's default feature set so the ACL harness uses the expected binaries

## Harness Expectations
- `node-a` sees  all the nodes. 
- `node-b` sees  all the nodes.
- `node-c` sees no peers
- `node-d` sees no peers
- `node-e` sees all the nodes.
- `node-f` sees all the nodes.

## Testing
- `cargo test --lib 100_nodes`
- `cargo test --lib`
- `cargo test test_acl_`
- `cargo test --bin fipsctl`
- `./testing/scripts/build.sh --no-docker`
- `./testing/acl-allowlist/test.sh --skip-build`


alexxie16:
This PR is for issue https://github.com/jmcorgan/fips/issues/35

src/bin/fipsctl.rs
```diff
@@ -7,7 +7,7 @@ use clap::{Parser, Subcommand};
 use fips::config::{write_key_file, write_pub_file};
 use fips::upper::hosts::HostMap;
 use fips::version;
-use fips::{encode_nsec, Identity};
+use fips::{Identity, encode_nsec};
```
SatsAndSports, line 13: This change is a no-op.

Not really specific to this PR, but we need a formatting policy 😀. I've made an issue here: https://github.com/jmcorgan/fips/issues/55

SatsAndSports: looking down below, I see a few more examples of this problem below

jmcorgan (changes requested):
Thanks for this — solid work for a first contribution. The architecture is clean and the test harness is well thought out.

Three items to address before this can merge:

**1. Allow/deny evaluation semantics**

The current implementation makes the allowlist authoritative — when `peers.allow` is non-empty, `peers.deny` is effectively ignored. Issue #35 specified TCP Wrappers ordering (check allow, check deny, default permit). An operator who creates both files expecting TCP Wrappers behavior will get surprising results. Please implement the TCP Wrappers evaluation order and add a comment block at the top of `acl.rs` documenting the evaluation rules.

**2. `build.sh` changes**

The change from `cargo build --release` to `--no-default-features --features tui` is unrelated to ACL and reduces the feature set for all test harnesses. Please revert this to keep the default build flags. If you need a reduced build for your local testing, that can go in the ACL test harness's own script.

**3. Operator precedence**

In `default_decision()`:
```rust
if self.allow_all || self.deny.is_empty() && !self.deny_all && self.allow.is_empty() {
```
This relies on `&&` binding tighter than `||`, which is correct but fragile. Please add parentheses around the `&&` chain.

**4. Formatting**

Please separate the rustfmt reformatting from behavioral changes — either in its own commit or omitted entirely. Mixing formatting with logic changes inflates the diff significantly and makes it much harder to review the actual ACL enforcement code.

jmcorgan (changes requested):
Thanks for the rework. The four items from the previous round all land correctly, and the tip commit's wiring repair is clean.

Three items before merge:

**1. Add `fipsctl acl show` subcommand (required)**

The harness at `testing/acl-allowlist/test.sh` calls `fipsctl acl show`, but there's no `Acl` subcommand in `src/bin/fipsctl.rs`. Only the backend RPC exists. Please add an `Acl { #[command(subcommand)] ... }` variant with a `Show` mapping to `"show_acl"`, mirroring the existing `ShowCommands` pattern.

**2. Clean up `api_reload_acl`**

`src/node/acl.rs:415-418` defines `api_reload_acl` but nothing dispatches it. Please either wire it to a `"reload_acl"` command or drop the method.

**3. Add the ACL harness to CI**

The Docker harness in `testing/acl-allowlist/` isn't in the integration matrix. Point 1 went unnoticed because nothing runs it. Please add an `Integration (acl-allowlist)` job matching the `rekey` / `sidecar` pattern.

Issue #34, opened 2026-03-27
Nostr relay peer discovery for overlay transports


jmcorgan:
## Background

FIPS nodes currently discover peers through two mechanisms:
- **Static configuration**: operator-defined `peers:` entries in `fips.yaml` with explicit transport addresses
- **Local-media discovery**: Ethernet beacons and BLE advertisements on shared physical media

There is no dynamic discovery mechanism for overlay transports (UDP, TCP, Tor). If a node's IP address changes or a new node joins the mesh, operators must manually update peer configs on both sides. This limits the mesh to a manually curated topology.

## Proposal

Since FIPS identities are Nostr keypairs, Nostr relays are a natural out-of-band signaling channel. A node with listening overlay transports can publish a signed replaceable event advertising its reachable endpoints. Other nodes query relays by npub (or by event kind) to discover transport addresses, then connect using the existing auto-connect path.

This issue covers the full lifecycle: design, NIP specification, and implementation.

### Event format

A FIPS peer advertisement should be a **NIP-78 replaceable event** (or a new dedicated kind) signed by the node's npub. The event carries one or more transport endpoints as structured tags:

```
kind: <TBD — replaceable>
content: "" (or optional metadata blob)
tags:
  ["t", "fips"]
  ["transport", "udp", "203.0.113.5:2121"]
  ["transport", "tcp", "node.example.com:8443"]
  ["transport", "tor", "abc...xyz.onion:8443"]
  ["version", "0.2.0"]           # optional: protocol version
```

Key properties:
- **Replaceable**: only the latest event per-npub matters — stale addresses are replaced, not accumulated
- **Multi-transport**: a single event advertises all listening transports, not one event per transport. A discovering node receives the full set and selects by transport availability and priority.
- **Self-authenticating**: the event is signed by the node's Nostr keypair, which is the same keypair used for FMP/FSP handshakes — no separate trust chain needed

### Discovery modes

1. **Known-peer lookup**: query relays for the latest event from a specific npub. Used when the operator has a peer's identity but its address has changed. This replaces manually updating `fips.yaml` after IP changes.

2. **Broadcast discovery**: query relays for all events of the FIPS advertisement kind. Used for bootstrapping — a new node finds peers without knowing any npub in advance. Requires the event kind to be filterable.

3. **Private advertisement** (future): NIP-59 gift-wrapped events visible only to specific recipients. For nodes that should not be publicly discoverable.

### Integration with existing discovery

[...]

PR #53, opened 2026-04-08
Add Nostr-based udp:nat bootstrap and example handoff flow

### LINKED ISSUE #37: UDP NAT traversal via Nostr relay signaling
## Background

UDP endpoints behind NAT gateways cannot accept inbound connections. Two nodes both behind NAT cannot establish a direct UDP link without a coordination mechanism to perform simultaneous hole punching. Currently, nodes behind NAT must use TCP or Tor transports, or have one side with a public IP.

## Existing protocol draft

A generic Nostr-signaled UDP hole punching protocol has been drafted at [`docs/proposals/nostr-udp-hole-punch-protocol.md`](https://github.com/jmcorgan/fips/blob/master/docs/proposals/nostr-udp-hole-punch-protocol.md). It covers the full lifecycle:

1. **Service advertisement** (kind 30078, NIP-78 parameterized replaceable) — responder publishes availability with relay list and STUN server preferences
2. **Discovery** — initiator queries relays by pubkey or protocol tag
3. **STUN binding** — initiator discovers its reflexive address via RFC 8489 STUN Binding Request
4. **Offer** (kind 21059, NIP-59 ephemeral gift-wrap) — initiator sends its reflexive address to responder, encrypted via NIP-44
5. **Answer** — responder performs its own STUN binding and replies with its reflexive address
6. **Hole punching** — both peers send punch packets (magic `0x4E505443` + session hash) every 200ms to each other's reflexive address; ack (`0x4E505441`) confirms the channel is open
7. **Application handoff** — the punched UDP socket is handed to the application protocol; Nostr signaling is no longer needed

The protocol uses ephemeral gift-wrapped events for signaling privacy, NIP-40 expiration for automatic cleanup, and includes LAN optimization (parallel local+reflexive address attempts when both peers are on the same subnet).

### Handoff to Noise handshake

Phase 6 of the generic protocol is where FIPS takes over. Once the punch/ack exchange confirms bidirectional UDP connectivity, the punched socket becomes the transport for a standard FMP Noise handshake. The existing UDP transport's `send`/`recv` paths need to work with a socket that was opened for hole punching rather than bound to a configured listen address.

### Relationship to peer discovery (#34)

Issue #34 proposes Nostr relay peer discovery using replaceable events to advertise transport endpoints. NAT traversal's Phase 0 (service advertisement) serves a similar purpose but adds NAT-specific metadata (STUN servers, relay list for signaling). These could be:

- **Combined**: a single replaceable event advertises both direct endpoints (for publicly reachable transports) and NAT traversal parameters (for UDP behind NAT)
- **Separate**: peer discovery events for direct endpoints, a second event type for NAT traversal availability

The combined approach is simpler — a discovering node sees all connectivity options in one event and chooses the appropriate method. A node with a public UDP endpoint just includes the `transport` tag; a node behind NAT includes `stun` and `relays` tags indicating it supports hole punching.

### Socket lifecycle

The punch socket must remain open from STUN binding through the entire connection lifetime. This conflicts with the current UDP transport model where a single bound socket handles all peers. Options:

- **Per-peer punch sockets**: each NAT-traversed connection uses its own socket. The UDP transport manages multiple sockets, demuxing by remote address.
- **Shared socket with port prediction**: use the listen socket for STUN binding and hole punching. Simpler but may not work with all NAT types (endpoint-dependent mapping).

Per-peer sockets are more reliable across NAT types but add file descriptor overhead and complicate the transport's recv loop.

### Integration with auto-connect and retry

- A configured peer with NAT traversal enabled follows the existing auto-connect path but invokes the hole punch sequence instead of direct UDP connect
- On failure (punch timeout after 10s), the retry system applies exponential backoff as with any failed connection
- Symmetric NAT on both sides is unresolvable — the node should fall back to TCP or Tor if configured as alternative transports for that peer

### Keepalive

NAT mappings expire if idle (typically 30-120s depending on the NAT). The existing MMP heartbeat interval (configurable, default varies) must be shorter than the NAT mapping timeout. The generic protocol recommends 15s keepalive; MMP heartbeats already satisfy this if the interval is configured appropriately.

## Configuration

```yaml
peers:
  - npub: npub1...
    addresses:
      - transport: udp
        addr: "nat"                      # signals NAT traversal needed
        priority: 10
      - transport: tor
        addr: "abc...xyz.onion:8443"     # fallback
        priority: 20

discovery:
  nostr:
    nat_traversal:
      enabled: true
      stun_servers:
        - "stun.l.google.com:19302"
        - "stun1.l.google.com:19302"
      relays:                            # can share with discovery relays
        - "wss://relay.damus.io"
      punch_timeout_secs: 10
      punch_interval_ms: 200
```

### Phase 1: Protocol finalization

- Review the existing protocol draft for FIPS-specific requirements
- Decide on combined vs. separate events for discovery and NAT traversal
- Define the Noise handshake handoff mechanics (which Noise pattern, how the punched socket is promoted to a transport link)
- Determine socket lifecycle strategy (per-peer vs. shared)
- Submit as a NIP or NIP-adjacent specification

### Phase 2: STUN client

- Minimal RFC 8489 STUN Binding Request/Response implementation (or evaluate existing Rust crates)
- Reflexive address discovery with fallback across multiple STUN servers
- Unit tests for STUN message encoding/decoding

### Phase 3: Nostr signaling

- Offer/answer construction with NIP-44 encryption and NIP-59 gift-wrap
- Relay subscription management for incoming signaling events
- Session ID tracking for replay protection
- Ephemeral event cleanup (NIP-09 deletion, NIP-40 expiration)
- May share infrastructure with #34 (peer discovery relay client)

### Phase 4: Hole punch and handoff

- Punch/ack packet exchange on the UDP socket
- Timeout and retry logic
- Handoff to FMP Noise handshake on the punched socket
- Integration with the UDP transport's connection and link management
- Auto-connect and retry system integration

### Phase 5: Testing

- Unit tests for STUN client, signaling message construction, punch/ack exchange
- Integration test: two nodes behind simulated NAT (Docker with iptables MASQUERADE) discover each other and establish a link
- Symmetric NAT failure test: verify clean timeout and fallback to alternative transport
- LAN optimization test: peers on the same subnet connect via local address

## Failure modes

| Failure | Mitigation |
|---|---|
| Symmetric NAT (one side) | Retry with port prediction; fall back to TCP/Tor |
| Symmetric NAT (both sides) | Cannot punch; must use TCP/Tor |
| Relay latency > 60s | Reflexive address stale; use low-latency relays |
| STUN server unreachable | Fall back to alternate STUN server |
| Firewall blocks outbound UDP | Protocol cannot proceed; TCP/Tor fallback |
| NAT mapping timeout during session | MMP heartbeat keeps mapping alive; if lost, link-dead detection triggers reconnect and fresh punch attempt |

tidley:
Closes #37.

Summary:
- refactor the in-tree Nostr bootstrap into focused internal modules while preserving the public API
- implement `udp:nat` bootstrap via Nostr adverts, encrypted offer/answer signaling, STUN-based address discovery, UDP punching, and handoff into the normal FIPS transport
- add UDP socket adoption so the punched socket can be handed into the existing handshake/session stack
- expand/document default STUN servers while keeping `node.discovery.nostr.stun_servers` fully configurable
- remove the shell-oriented example path and narrow the example crate to the remaining console and video flows
- update config, example, and proposal docs to match the current code

Maintainer notes:
- Built-in defaults include contributor-operated infrastructure:
  - advert relay: `wss://strfry.bitsbytom.com`
  - STUN server: `stun:fips.tomdwyer.uk:3478`
- The traversal design is intentionally multi-port:
  - each attempt binds a fresh UDP socket for STUN, punching, and FIPS handoff
  - the long-lived application listener is not reused as the punch socket
- Signal payload identities are now bound to the actual Nostr sender, and outbound STUN now uses only the local configured allowlist; peer-advertised STUN remains informational only
- Inbox-relay lookup now queries across configured DM + advert relays and falls back to local DM relays if remote metadata fetch fails
- Failed adopted-traversal handoff now rolls back the inserted UDP transport instead of leaving an orphaned live transport behind
- `fips-console-server` handles one active session at a time, then recreates its listener for the next peer
- Symmetric NAT port prediction is still not implemented; unsupported topologies still fail and fall back

Validation:
- cargo fmt --all
- cargo check --lib --no-default-features
- cargo test --lib bootstrap::nostr --no-default-features
- cargo test --lib test_adopted_udp_traversal_completes_handshake --no-default-features
- cargo test --lib test_failed_adopted_traversal_cleans_up_transport --no-default-features
- cargo check --manifest-path examples/nostr-bootstrap/Cargo.toml --bins
- cargo test --manifest-path examples/nostr-bootstrap/Cargo.toml
- cargo clippy --all -- -D warnings
- manually verified the console client/server example still connects across machines

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -301,6 +301,16 @@ jobs:
           # ── Sidecar deployment ──────────────────────────────────────────
           - suite: sidecar
             type: sidecar
+          # ── NAT traversal lab (Nostr/STUN UDP hole punch) ───────────────
+          - suite: nat-cone
+            type: nat
+            scenario: cone
+          - suite: nat-symmetric
+            type: nat
+            scenario: symmetric
+          - suite: nat-lan
+            type: nat
+            scenario: lan
 
     steps:
       - uses: actions/checkout@v4
@@ -426,3 +436,21 @@ jobs:
             docker logs "sidecar-${node}-fips-1" 2>&1 || true
             echo ""
           done
+
+      # ── NAT traversal lab ───────────────────────────────────────────────
+      - name: Run NAT lab scenario
+        if: matrix.type == 'nat'
+        run: bash testing/nat/scripts/nat-test.sh ${{ matrix.scenario }}
+
+      - name: Collect logs on failure (nat)
+        if: matrix.type == 'nat' && failure()
+        run: |
+          docker compose -f testing/nat/docker-compose.yml \
+            --profile ${{ matrix.scenario }} logs --no-color
+
+      - name: Stop containers (nat)
+        if: matrix.type == 'nat' && always()
+        run: |
+          docker compose -f testing/nat/docker-compose.yml \
+            --profile cone --profile symmetric --profile lan \
+            down --volumes --remove-orphans
--- a/.github/workflows/package-openwrt.yml
+++ b/.github/workflows/package-openwrt.yml
@@ -247,7 +247,7 @@ jobs:
         id: publish
         shell: bash
         env:
-          RELAYS: "wss://relay.damus.io wss://nos.lol wss://nostr.mom wss://relay.primal.net"
+          RELAYS: "wss://relay.damus.io wss://nos.lol wss://nostr.mom wss://offchain.pub"
           NSEC: ${{ steps.keys.outputs.nsec }}
         run: |
           : ${GITHUB_OUTPUT:=/tmp/github_output}
@@ -295,7 +295,7 @@ jobs:
         if: ${{ steps.publish.outputs.eventId != '' }}
         env:
           EVENT_ID: ${{ steps.publish.outputs.eventId }}
-          RELAYS: "wss://relay.damus.io wss://nos.lol wss://nostr.mom wss://relay.primal.net"
+          RELAYS: "wss://relay.damus.io wss://nos.lol wss://nostr.mom wss://offchain.pub"
         run: |
           echo "Verifying event $EVENT_ID on relays..."
           FOUND=0
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,9 @@ __pycache__/
 *.egg-info/
 *.egg
 
+# Runtime artifacts from running fips in-tree during local testing.
+# Root-anchored so legitimately-tracked fips.yaml under packaging/ and
+# examples/ stays included.
+/fips.key
+/fips.pub
+/fips.yaml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   forwarding, proxy NDP, RA route advertisements, and IPv6 forwarding
   sysctls. Gateway enabled by default on OpenWrt
 
+#### Nostr-Mediated Discovery and NAT Traversal
+
+- Optional overlay-discovery and NAT-hole-punching path behind the
+  `nostr-discovery` cargo feature. Nodes publish signed overlay adverts
+  as Nostr kind `37195` parameterized replaceable events listing
+  reachable transport endpoints to a configurable set of public relays,
+  and consume peer adverts to populate fallback addresses for
+  `via_nostr` peers or, under `policy: open`, for non-configured peers
+  within a budget cap. The kind value is FIPS-specific: `37195` sits in
+  the application-defined replaceable range `30000–39999`, and the
+  digits visually spell `FIPS` (7=F, 1=I, 9=P, 5=S)
+- STUN-assisted UDP hole punching for `addr: "nat"` UDP endpoints. STUN
+  reflexive observation, gift-wrap (NIP-59) offer/answer signaling, and
+  candidate-pair punch planner (LAN-private + reflexive paths attempted in
+  parallel). Successful punches hand the live socket into the standard
+  FIPS UDP transport via a bootstrap-handoff API
+- New `node.discovery.nostr.*` configuration tree with operator-tunable
+  resource caps, replay tracking, and punch timing; new `peers[].via_nostr`
+  and per-transport `advertise_on_nostr` / `public` flags. Cross-field
+  validation at startup catches mis-configured combinations
+- Docker NAT lab covering cone, symmetric (TCP-fallback), and LAN
+  scenarios, wired into the integration CI matrix
+
 #### Examples
 
 - macOS WireGuard sidecar: run FIPS in a local Docker container and
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -123,6 +123,37 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8b75356056920673b02621b35afd0f7dda9306d03c79a30f5c56c44cf256e3de"
 
 [[package]]
+name = "async-utility"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a34a3b57207a7a1007832416c3e4862378c8451b4e8e093e436f48c2d3d2c151"
+dependencies = [
+ "futures-util",
+ "gloo-timers",
+ "tokio",
+ "wasm-bindgen-futures",
+]
+
+[[package]]
+name = "async-wsocket"
+version = "0.13.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1c92385c7c8b3eb2de1b78aeca225212e4c9a69a78b802832759b108681a5069"
+dependencies = [
+ "async-utility",
+ "futures",
+ "futures-util",
+ "js-sys",
+ "tokio",
+ "tokio-rustls",
+ "tokio-socks",
+ "tokio-tungstenite",
+ "url",
+ "wasm-bindgen",
+ "web-sys",
+]
+
+[[package]]
 name = "atomic"
 version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -130,6 +161,12 @@ checksum = "a89cbf775b137e9b968e67227ef7f775587cde3fd31b0d8599dbd0f598a48340"
 dependencies = [
  "bytemuck",
 ]
+
+[[package]]
+name = "atomic-destructor"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ef49f5882e4b6afaac09ad239a4f8c70a24b8f2b0897edb1f706008efd109cf4"
 
 [[package]]
 name = "atomic-waker"
@@ -148,6 +185,12 @@ name = "base64"
 version = "0.22.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
+
+[[package]]
+name = "base64ct"
+version = "1.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2af50177e190e07a26ab74f8b1efbfe2ef87da2116221318cb1c2e82baf7de06"
 
 [[package]]
 name = "bech32"
@@ -173,6 +216,17 @@ dependencies = [
  "rustc-hash",
  "shlex",
  "syn 2.0.114",
+]
+
+[[package]]
+name = "bip39"
+version = "2.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "90dbd31c98227229239363921e60fcf5e558e43ec69094d46fc4996f08d1d5bc"
+dependencies = [
+ "bitcoin_hashes",
+ "serde",
+ "unicode-normalization",
 ]
 
 [[package]]
@@ -204,6 +258,7 @@ checksum = "26ec84b80c482df901772e931a9a681e26a1b9ee2302edeff23cb30328745c8b"
 dependencies = [
  "bitcoin-io",
  "hex-conservative",
+ "serde",
 ]
 
 [[package]]
@@ -223,6 +278,15 @@ name = "block-buffer"
 version = "0.10.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
+dependencies = [
+ "generic-array",
+]
+
+[[package]]
+name = "block-padding"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a8894febbff9f758034a5b8e12d87918f56dfc64a8e1fe757d65e29041538d93"
 dependencies = [
  "generic-array",
 ]
@@ -340,6 +404,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dec551ab6e7578819132c713a93c022a05d60159dc86e7a7050223577484c55a"
 dependencies = [
  "rustversion",
+]
+
+[[package]]
+name = "cbc"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "26b52a9543ae338f279b96b0b9fed9c8093744685043739079ce85cd58f289a6"
+dependencies = [
+ "cipher",
 ]
 
 [[package]]
@@ -759,6 +832,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "data-encoding"
+version = "2.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d7a1e2f27636f116493b8b860f5546edb47c8d8f8ea73e1d2a20be88e28d1fea"
+
+[[package]]
 name = "dbus"
 version = "0.9.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -985,12 +1064,14 @@ dependencies = [
  "hex",
  "hkdf",
  "libc",
+ "nostr",
+ "nostr-sdk",
  "portable-atomic",
  "rand 0.10.0",
  "ratatui",
  "rtnetlink",
  "rustables",
- "secp256k1",
+ "secp256k1 0.30.0",
  "serde",
  "serde_json",
  "serde_yaml",
@@ -1031,6 +1112,15 @@ name = "foldhash"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77ce24cb58228fbb8aa041425bb1050850ac19177686ea6e0f41a70416f56fdb"
+
+[[package]]
+name = "form_urlencoded"
+version = "1.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cb4cb245038516f5f85277875cdaa4f7d2c9a0fa0468de06ed190163b1581fcf"
+dependencies = [
+ "percent-encoding",
+]
 
 [[package]]
 name = "futures"
@@ -1148,8 +1238,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ff2abc00be7fca6ebc474524697ae276ad847ad0a6b3faa4bcb027e9a4614ad0"
 dependencies = [
  "cfg-if",
+ "js-sys",
  "libc",
  "wasi",
+ "wasm-bindgen",
 ]
 
 [[package]]
@@ -1183,6 +1275,18 @@ name = "glob"
 version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0cc23270f6e1808e30a928bdc84dea0b9b4136a8bc82338574f23baf47bbd280"
+
+[[package]]
+name = "gloo-timers"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bbb143cf96099802033e0d4f4963b19fd2e0b728bcf076cd9cf7f6634f092994"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "js-sys",
+ "wasm-bindgen",
+]
 
 [[package]]
 name = "half"
@@ -1255,6 +1359,104 @@ dependencies = [
 ]
 
 [[package]]
+name = "http"
+version = "1.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e3ba2a386d7f85a81f119ad7498ebe444d2e22c2af0b86b069416ace48b3311a"
+dependencies = [
+ "bytes",
+ "itoa",
+]
+
+[[package]]
+name = "httparse"
+version = "1.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6dbf3de79e51f3d586ab4cb9d5c3e2c14aa28ed23d180cf89b4df0454a69cc87"
+
+[[package]]
+name = "icu_collections"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2984d1cd16c883d7935b9e07e44071dca8d917fd52ecc02c04d5fa0b5a3f191c"
+dependencies = [
+ "displaydoc",
+ "potential_utf",
+ "utf8_iter",
+ "yoke",
+ "zerofrom",
+ "zerovec",
+]
+
+[[package]]
+name = "icu_locale_core"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "92219b62b3e2b4d88ac5119f8904c10f8f61bf7e95b640d25ba3075e6cac2c29"
+dependencies = [
+ "displaydoc",
+ "litemap",
+ "tinystr",
+ "writeable",
+ "zerovec",
+]
+
+[[package]]
+name = "icu_normalizer"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c56e5ee99d6e3d33bd91c5d85458b6005a22140021cc324cea84dd0e72cff3b4"
+dependencies = [
+ "icu_collections",
+ "icu_normalizer_data",
+ "icu_properties",
+ "icu_provider",
+ "smallvec",
+ "zerovec",
+]
+
+[[package]]
+name = "icu_normalizer_data"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "da3be0ae77ea334f4da67c12f149704f19f81d1adf7c51cf482943e84a2bad38"
+
+[[package]]
+name = "icu_properties"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bee3b67d0ea5c2cca5003417989af8996f8604e34fb9ddf96208a033901e70de"
+dependencies = [
+ "icu_collections",
+ "icu_locale_core",
+ "icu_properties_data",
+ "icu_provider",
+ "zerotrie",
+ "zerovec",
+]
+
+[[package]]
+name = "icu_properties_data"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8e2bbb201e0c04f7b4b3e14382af113e17ba4f63e2c9d2ee626b720cbce54a14"
+
+[[package]]
+name = "icu_provider"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "139c4cf31c8b5f33d7e199446eff9c1e02decfc2f0eec2c8d71f65befa45b421"
+dependencies = [
+ "displaydoc",
+ "icu_locale_core",
+ "writeable",
+ "yoke",
+ "zerofrom",
+ "zerotrie",
+ "zerovec",
+]
+
+[[package]]
 name = "id-arena"
 version = "2.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1265,6 +1467,27 @@ name = "ident_case"
 version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b9e0384b61958566e926dc50660321d12159025e767c18e043daf26b70104c39"
+
+[[package]]
+name = "idna"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3b0875f23caa03898994f6ddc501886a45c7d3d62d04d2d90788d47be1b1e4de"
+dependencies = [
+ "idna_adapter",
+ "smallvec",
+ "utf8_iter",
+]
+
+[[package]]
+name = "idna_adapter"
+version = "1.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3acae9609540aa318d1bc588455225fb2085b9ed0c4f6bd0d9d5bcd86f1a0344"
+dependencies = [
+ "icu_normalizer",
+ "icu_properties",
+]
 
 [[package]]
 name = "indexmap"
@@ -1293,6 +1516,7 @@ version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "879f10e63c20629ecabbb64a8010319738c66a5cd0c29b02d63d272b03751d01"
 dependencies = [
+ "block-padding",
  "generic-array",
 ]
 
@@ -1307,6 +1531,18 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.114",
+]
+
+[[package]]
+name = "instant"
+version = "0.1.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e0242819d153cba4b4b05a5a8f2a7e9bbf97b6055b2a002b395c96b5ff3c0222"
+dependencies = [
+ "cfg-if",
+ "js-sys",
+ "wasm-bindgen",
+ "web-sys",
 ]
 
 [[package]]
@@ -1451,6 +1687,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "df1d3c3b53da64cf5760482273a98e575c651a67eec7f77df96b5b642de8f039"
 
 [[package]]
+name = "litemap"
+version = "0.8.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "92daf443525c4cce67b150400bc2316076100ce0b3686209eb8cf3c31612e6f0"
+
+[[package]]
 name = "litrs"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1545,6 +1787,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "negentropy"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f0efe882e02d206d8d279c20eb40e03baf7cb5136a1476dc084a324fbc3ec42d"
+
+[[package]]
 name = "netlink-packet-core"
 version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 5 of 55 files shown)


jmcorgan (commented):
This is a solid design — Nostr relay signaling for NAT traversal with socket handoff into the existing transport stack is the right architecture, and the Docker NAT test harness is useful infrastructure. A few items to address before consideration for merge.

## Structural

**Feature gating required.** `nostr-sdk` adds ~50 transitive crates (rustls, tungstenite, ring, bip39, scrypt, a second secp256k1). This needs to be behind a cargo feature flag so builds that don't use NAT traversal don't pay the cost.

**Split the app port system into a separate PR.** `src/app.rs`, `bind_app_port()`, `AppDatagram`, and the related session handler changes are a distinct feature from the NAT bootstrap. Reviewing and merging these independently will be easier.

**Rustfmt divergence.** Large sections of the diff are reformatting of existing code (e.g., every `fn default_*()` in `config/node.rs`). Please configure your formatter to match the project settings, or separate formatting changes so the substantive diff is reviewable.

## Technical

**Transport cleanup.** Each NAT-traversed peer gets a dedicated `UdpTransport` via `adopt_established_traversal()`. There's no cleanup when the peer disconnects — transport count grows unbounded with churn. Need a teardown path.

**STUN parser is IPv4-only.** `parse_mapped_address` and `parse_xor_mapped_address` only handle `family == 0x01`. IPv6 STUN responses are silently ignored. Consider using an existing STUN crate, or at minimum document the limitation.

**Mixed clock sources.** The bootstrap module uses `SystemTime` for TTL and freshness checks, while the rest of FIPS uses `Instant`. Wall-clock adjustments (NTP jumps, suspend/resume) could cause stale signal acceptance or premature expiry.

**Unbounded task spawning.** Inbound offer handling in the notify loop spawns tasks with no concurrency limit. A flood of offers would create unbounded work.

**Hardcoded external infrastructure.** Default STUN servers (Google, Twilio, contributor-operated) and Nostr relays are baked into config defaults. These should be documented as defaults that operators are expected to override, and the contributor-operated endpoints should be noted as best-effort.

**`std::sync::mpsc` in async context.** The app port delivery uses `std::sync::mpsc::Sender` while the rest of the codebase uses `tokio::sync::mpsc`. Should be consistent.

## Minor

- `current_time_ms()` free function duplicates `Node::now_ms()`
- Session handler behavior change: unknown FSP service ports now route to app ports instead of being dropped. Functionally equivalent when no ports are bound, but worth calling out in the commit message.

tidley:
Addressed the review points in the current branch.

Structural
- Feature gating: `nostr` / `nostr-sdk` are now behind the `nostr-bootstrap` cargo feature, so builds that do not use NAT traversal do not pull in that dependency set.
- App port split: the app-port work (`src/app.rs`, `bind_app_port()`, `AppDatagram`, and the related session-handler routing) has been removed from this PR. This branch now only carries the NAT/bootstrap work.
- Rustfmt divergence: this branch has been rebased onto the repository-wide `cargo fmt` baseline now on `master`, and `cargo fmt --check` passes.

Technical
- Transport cleanup: adopted traversal transports are now cleaned up on failed handoff, stale handshake timeout, and peer/session teardown so they do not accumulate with churn.
- STUN IPv6: the in-tree STUN parser now handles both IPv4 and IPv6 `MAPPED-ADDRESS` / `XOR-MAPPED-ADDRESS` responses.
- Clock source: bootstrap freshness/TTL handling now uses a monotonic `Instant`-anchored `now_ms()` path rather than raw wall-clock checks.
- Inbound offer handling: notify-loop offer processing is now bounded with a semaphore instead of spawning unbounded work.
- Hardcoded infrastructure: the docs now explicitly describe the built-in STUN/relay values as defaults operators are expected to override, and note the contributor-operated endpoints as best-effort.
- `std::sync::mpsc` in async context: the app-port path that introduced that inconsistency was removed from this PR, so this issue no longer applies to the remaining diff.

Minor
- `current_time_ms()` duplication: removed in favor of the existing `Node::now_ms()` path.
- Unknown FSP service ports: they are dropped again; the app-port routing behavior is no longer present in this PR.

Additional follow-up included in this branch
- Signal payload identities are bound to the actual Nostr gift-wrap sender.
- Outbound STUN now uses only the local configured allowlist; peer-advertised STUN values are informational only.
- Inbox-relay lookup now queries across configured DM + advert relays and falls back to local DM relays if remote metadata fetch fails.
- The missing `gateway` module export was restored so gateway-enabled builds pass again.

Validation run on the current rebased branch
- `cargo fmt --check`
- `cargo clippy --all -- -D warnings`
- `cargo test --lib --no-default-features`
- `cargo test --lib bootstrap::nostr --no-default-features --features nostr-bootstrap`
- `cargo test --manifest-path examples/nostr-bootstrap/Cargo.toml`
- `cargo build --release --features gateway`
- `cargo nextest run --all --profile ci --features gateway`

tidley:
Fixed two CI issues on the PR branch.

1. macOS build failure
Cause:
- `run_tun_reader()` on macOS now takes a shutdown pipe fd
- `src/node/lifecycle.rs` was still calling it with the older 6-argument form

Fix:
- added a macOS-only TUN reader shutdown fd field to `Node`
- create/store the shutdown pipe during TUN startup
- pass the read end into `run_tun_reader()`
- signal and close the write end during shutdown before joining the reader thread

2. rustfmt toolchain/component failure
Cause:
- the pinned toolchain in `rust-toolchain.toml` only specified `channel = "1.94.0"`
- on runners that honored that file but did not already have `rustfmt` installed, `cargo fmt --check` failed before checking formatting

Fix:
- updated `rust-toolchain.toml` to include:
  - `components = ["rustfmt"]`

This is now pushed on the PR branch at `0676087`.

Revalidated locally after these fixes:
- `cargo fmt --check`
- `cargo clippy --all -- -D warnings`
- `cargo test --lib --no-default-features`
- `cargo build --release --features gateway`

tidley:
I found a CI wiring issue that likely explains the NAT integration failures, and I also cleaned up the NAT diagnostics.

The NAT lab containers are built from the Linux release artifact produced in `ci.yml`. That artifact was being built with `--features gateway`, but the NAT scenarios require `nostr-bootstrap` as well. In other words, the integration image used by `nat-cone`, `nat-symmetric`, and `nat-lan` did not actually contain the Nostr/STUN bootstrap code path those scenarios exercise.

I’ve updated CI so the Linux build/test path uses:
- `cargo build --release --features "gateway nostr-bootstrap"`
- `cargo nextest run --all --profile ci --features "gateway nostr-bootstrap"`

I also updated `testing/nat/scripts/nat-test.sh` because parts of the NAT diagnostics were misleading:
- the “active probe” is now a real STUN binding request instead of `nc -u` sending the literal payload `probe`
- LAN diagnostics now use the LAN relay/STUN addresses (`172.31.10.30` / `172.31.10.40`) instead of the WAN-only `172.31.254.*` addresses
- the previous `nc -u` probe output should not be treated as evidence that the STUN service was broken, because the test STUN server intentionally ignores arbitrary non-STUN UDP payloads

This is now pushed on the PR branch at `578f456`.

jmcorgan:
@tidley, one piece of context up front: the prior review rounds were largely mechanical passes focused on specific items (feature gating, clock sources, task spawning, identity verification). Useful and necessary, but narrow in scope. This is the first round where I've had the chance to do a thorough manual analysis of the full implementation end-to-end.

Credit where it's due: you've been responsive across every round, and what's here works. Feature gating, app-port split, rustfmt alignment, transport cleanup on disconnect, STUN IPv6, monotonic clock, bounded task spawning: all addressed, with a working multi-scenario NAT test harness to back it up. The UDP hole-punching mechanism itself is solid: offer/answer via NIP-59, STUN-driven reflexive discovery, mutual UDP probes, and socket handoff into the FIPS transport stack all hang together, and the integration tests confirm the Noise handshake completes over the punched socket. That's the hard part of #37, and it's proven out.

I know this round lands differently than the previous ones, so some background on why. #34 (Nostr relay peer discovery for overlay transports) was always intended to be the first step of this work, with hole punching layered on top once the advert and discovery plumbing was in place. The current PR went straight to the harder problem, which worked (the protocol runs end-to-end), but the advert format, subscription scheme, and discovery flow that came out of it are shaped specifically for NAT punch. Merging as-is means a wire-format migration later when we come back to finish #34 properly. I'd rather backfill #34 into this PR so both issues close together on a single schema.

The asks break into four concrete structural items, which I'll lay out in separate comments below so we can discuss each independently. There's also a handful of smaller items I noticed, but I'll hold those for a follow-up round: some may reshape or dissolve during the rework anyway. Push back on anything that feels off. The payoff is that after the restructuring the feature lands as "Nostr-mediated overlay peer discovery with UDP hole-punch fallback," which closes both #37 and #34 in one PR and fits cleanly into the shape the daemon was built to host.

jmcorgan:
**Structural item 1 of 4: generalize the advert and build out the peer discovery mechanism — close #34 in this PR**

The feature this wants to land is **relay-mediated overlay peer discovery**: a node with a listening UDP/TCP/Tor endpoint publishes a signed advert on one or more Nostr relays, and other nodes subscribe and learn where to reach peers they care about. Conceptually it plays the same role that Ethernet and BLE beacons already play in the codebase — a way for nodes to find each other without manually configured addresses — but operating over public relay infrastructure rather than a single L2 segment, so it reaches past local scope. NAT hole-punching becomes one endpoint type in that advert (`"I only accept connections via rendezvous"`), not the whole point of it.

Mechanically it breaks into three pieces, each of which the current PR has partial primitives for:

1. **Production.** Each overlay transport (UDP, TCP, Tor) gets an `advertise_on_nostr: bool` flag in its config. When set, the transport's listening endpoint is bundled into the node's advert at publish time. The nostr subsystem queries operational transports at advert-build time — so if a transport goes down, the next refresh excludes it. Transports don't depend on the nostr crate; they just expose a method that returns their advertisable endpoint. (Config surface and validator rules are covered in item 3.)

2. **Publication.** The advert carries a list of `{transport, addr}` endpoints with the same shape `PeerConfig.addresses` already uses in fips.yaml, with `addr:"nat"` (UDP-only) signaling "rendezvous required." `signalRelays` and `stunServers` are only present when the advert includes a `udp:"nat"` entry — pure direct-dial advertisers don't publish them at all.

3. **Consumption.** A periodic tick walks `advert_cache`, filters by a configurable policy, and synthesizes in-memory `PeerConfig` entries from adverts for eligible npubs. Those go through `try_peer_addresses` — which is already the merged dial entry point. Direct endpoints dial via the existing UDP/TCP/Tor transports unchanged; `udp:"nat"` endpoints fall into the rendezvous path you already built.

The PR today has partial (1) (a single hardcoded advert built from one config block), partial (2) (schema is NAT-specific and can't express the other endpoint types), and none of (3). Backfilling these is what closes #34 alongside #37.

**Proposed advert shape:**

```json
{
  "publisherNpub": "npub1...",
  "publishedAt":   "...",
  "expiresAt":     "...",
  "sequence":      "...",
  "endpoints": [
    { "transport": "udp", "addr": "203.0.113.5:2121" },
    { "transport": "udp", "addr": "nat" },
    { "transport": "tcp", "addr": "198.51.100.7:8443" },
    { "transport": "tor", "addr": "abc...def.onion:8443" }
  ],
  "signalRelays": ["wss://..."],
  "stunServers":  ["stun:..."]
}
```

The `{transport, addr}` entries are structurally identical to `PeerConfig.addresses`, so synthesizing a peer config from an advert is a direct copy. No `natRequired` flag is needed — `"nat"` as addr is UDP-only by construction (the only transport with a punch pathway in this PR) and reuses the existing sentinel recognition in `try_peer_addresses`.

The subscription filter should include `.identifier("fips-overlay-v1")` so a node only ingests relevant adverts. Currently it's wildcard over Kind 30078, which picks up every NIP-33 event on that kind from every unrelated app — `advert_cache` grows unbounded on a busy relay.

**Connection policy**, configurable, default `config`:

| policy | behavior |
|--------|----------|
| `config` | Dial only npubs in the static `peers:` list. For entries with `via_nostr: true`, consult the peer's advert to resolve the endpoint address. |
| `open` | In addition to `config` behavior, also dial npubs that appear in the advert cache but aren't in static config. Public-mesh / bootstrap-node mode. |

Per-npub cooldown prevents an advert refresh from triggering a burst of reconnects.

**Closes** #37 and #34 together — the point of the restructuring is that after this item lands, the PR delivers the feature in the shape the daemon was built to host it.

jmcorgan:
**Structural item 2 of 4: start transports before any Nostr activity**

`NostrBootstrap::start()` currently runs at [lifecycle.rs:525](https://github.com/jmcorgan/fips/blob/issue37/src/node/lifecycle.rs#L525), before `create_transports()` and before any `handle.start()` on the transports. That sequence needs to flip.

Even with the current NAT-specific advert, there's a window between "advert published" and "UDP transport listening" during which a peer who sees the advert and dials gets a connection-refused: the advert says "I accept rendezvous" but the daemon hasn't bound its UDP socket yet. Small but real.

Once item 1 lands, the ordering becomes hard-required: the advert's `endpoints` list is built by querying each operational transport for its advertisable endpoint. If transports haven't started yet there's nothing to enumerate — you'd either publish an empty advert or have to defer publication anyway.

**The fix:** move the `NostrBootstrap::start()` block to after the transport-start loop. Proposed new order in `Node::start()`:

1. State transition + packet channel (unchanged)
2. `create_transports(...).await`
3. Loop: `handle.start().await` on each, insert into `self.transports`
4. TUN setup (unchanged)
5. **← `NostrBootstrap::start()` here.** `publish_advert` can now enumerate real transport endpoints.
6. Spawn rx_loop + remaining periodic tasks

`NostrBootstrap::start()` failure stays non-fatal (warn + continue), same as today.

**Shutdown order is already correct** — `bootstrap.shutdown()` runs before transport stop in `Node::stop()`, which withdraws the advert before transports go dark. Symmetric with the new startup order. Nothing to change there.

**Tor onion propagation caveat:** a Tor transport `handle.start()` returning `Ok` means local Tor sockets are up, not that the onion descriptor has propagated to the Tor DHT (typically 30–60 seconds on a warm Tor). If the advert lists a Tor onion immediately, peers dialing within that window will fail and retry. My preference is to accept the window — retries are already in the dialer path, descriptor propagation is fast enough in practice, and waiting on it would add plumbing for a transient condition. Worth one sentence in the proposal doc so operators aren't surprised.

jmcorgan:
**Structural item 3 of 4: per-transport advertise flag, peer `via_nostr` flag, config validator rule**

This is the configuration-surface and layering detail for item 1's "production" piece. It also supersedes the current `node.discovery.nostr.advertise` flag — that belongs on the transports, not in the nostr block.

### Transport config surface

Each overlay transport gets an `advertise_on_nostr: bool`. UDP additionally gets a `public: bool` that determines whether the advert lists a direct address or the rendezvous sentinel:

```yaml
transports:
  udp:
    bind: "0.0.0.0:2121"
    advertise_on_nostr: true
    public: true                  # or false, for a NAT'd node
  tcp:
    bind: "0.0.0.0:8443"
    advertise_on_nostr: true
  tor:
    onion: "abc.onion:8443"
    advertise_on_nostr: true
```

UDP semantics:
- `public: true` → advert gets `{udp, "<bind-addr>"}`. Peers dial directly; NAT mapping on their side opens on first packet.
- `public: false` → advert gets `{udp, "nat"}`. Peers initiate the rendezvous + punch flow to reach this node.

This is binary — either the UDP endpoint is publicly reachable and rendezvous is not necessary, or it isn't and rendezvous is the only path. No meaningful third state.

TCP and Tor don't have this duality in this PR, so they only have the advertise flag.

### Layering

Transports don't depend on the nostr crate. Each transport exposes a method roughly like:

```rust
fn advertisable_endpoint(&self) -> Option<AdvertEndpoint>;
```

The nostr subsystem iterates `self.transports` at advert-build time, calls the method on each, builds the endpoints list. No nostr types leak into the transport modules.

### Config validator rule

`advertise_on_nostr: true` on any transport requires `node.discovery.nostr.enabled = true` — otherwise there's nothing to advertise with. Reject the config at load time.

That's the only structural validator rule needed; the rest falls out of the binary public/non-public semantics above.

### Peer config: `via_nostr`

Today `peers:` entries bind an npub to a hard-coded address list. With adverts as a runtime source of endpoints, peer entries gain a `via_nostr` field:

```yaml
peers:
  - npub: npub1X
    via_nostr: true                 # resolve endpoints from advert
  - npub: npub1Y                    # legacy: explicit addresses
    addresses:
      - {transport: udp, addr: "1.2.3.4:2121"}
  - npub: npub1Z                    # hybrid: static + advert fallback
    addresses:
      - {transport: udp, addr: "5.6.7.8:2121"}
    via_nostr: true
```

At dial time, `try_peer_addresses` merges:

1. Explicit `addresses:` from config (tried first — operator intent).
2. If `via_nostr: true` and an advert is cached for this npub, append the advert's endpoints.

`{transport:"udp", addr:"nat"}` from either source triggers the rendezvous path — unchanged from the current PR.

### Three sources of a dial attempt

After items 1 and 3 land, a connection attempt can originate from:

| source | controlled by |
|--------|---------------|
| Static addresses | Operator — `addresses: [...]` on a peer entry |
| Advert for a statically-named npub | `via_nostr: true` on a peer entry |
| Advert for a non-configured npub | Discovery loop + `discover_policy: open` |

All three feed the same merged list consumed by `try_peer_addresses`.

jmcorgan:
**Structural item 4 of 4: delete `examples/nostr-bootstrap/`**

The sub-crate at `examples/nostr-bootstrap/` (package `fips-nostr-rendezvous`) looks like the original prototype that validated the design before the integrated version was carved into `src/bootstrap/nostr/`. Now that the integration is in place, the sub-crate can go.

**It's safe to delete** — nothing in the repo depends on it. `grep -r "examples/nostr-bootstrap\|fips-nostr-rendezvous"` across `testing/nat/` returns zero matches. The NAT harness's `docker-compose.yml` runs `image: fips-test:latest`, and the node entrypoint execs the real fips daemon — same pattern as `testing/static/` and `testing/rekey/`. The integration tests already exercise the real runtime end-to-end against a real strfry relay and a Python STUN server. No harness work needed.

**Reasons to remove it:**

- **Duplicate protocol code.** Wire types, STUN parsing, punch packet logic, and signaling helpers all exist twice — here and in `src/bootstrap/nostr/`. Future protocol changes have to be made in two places and can drift silently.
- **Diverged shape.** The sub-crate is split into `client_runtime/` + `server_runtime/`; the integrated version uses a single unified `NostrBootstrap` for both roles. The older shape is frozen and nothing else in the codebase follows it.
- **No runnable binaries.** The README notes that the demo binaries were removed in a prior review round (the AppDatagram split). What's left is library code with no entry points.
- **Diff-size cost.** ~3,200 of the 8,488 additions in this PR are here. That's review load for zero shipped functionality.

**What to do:** delete the directory, and port any tests from `examples/nostr-bootstrap/src/tests.rs` that aren't already covered by `src/bootstrap/nostr/tests.rs` into the main crate. A cursory comparison suggests most protocol tests are already in both places.

**Small side note on naming:** the sub-crate is named `fips-nostr-rendezvous` while the main-crate feature is `nostr-bootstrap`. You already reached for a different word to name the standalone version, which is a hint that "bootstrap" isn't the right name for the whole thing. We'll settle on a consistent name before merge — just flagging that the naming is on the list.

SatsAndSports:
A small suggestion on privacy, deviating from the protocol:

The giftwrapped messages are addressed to the pubkeys of the two nodes. I suggest changing the 'offer', which is sent by the initiator with the STUN results, such that it includes an ephemeral pubkey to be used for the reply. The final 'answer', which includes the responder's STUN results, can then be addressed to that pubkey

With this setup, third parties can see the original advertizement and can see that somebody initiatiated a connection to that npub (i.e. the responder's npub), but they can't see the npub of the initiator

jmcorgan:
Good suggestion, and the cost is essentially a field on the offer. I'd fold this into the advert/offer schema rework rather than a separate round, since that schema is already being generalized. One thing worth flagging: the responder still sees the initiator's pubkey inside the offer payload (needed for auth), so this protects against relay observers, not against the other endpoint, which I assume matches your intent.

SatsAndSports:
So to clarify the expected behaviour on the _initiator's_ side:

They will initiate a nostr-mediated connection to a given npub only if all of the following are true:
 - The peer npub is listed in their list of `peers`, with `via_nostr: true`, and
 - (if present), the addresses in the list of static `addresses` have already been (recently) tried and failed, and
 - (Question) is it also necessary that `auto_connect` be true?

I ask partly because I was using the assumption that we would auto-connect to any recent advert, regardless of the npub. Although that would get crazy when thousands of people are advertizing 😀

```
  - npub: npub1Z                    # hybrid: static + advert fallback
    addresses:
      - {transport: udp, addr: "5.6.7.8:2121"}
    via_nostr: true
```

tidley:
Thanks both for the clarifications and suggestions - no push back.

In terms of initiators and peers, I imagined a new node could opt to connect with no configured peers to ease onboarding. There may be a lot of `fips-overlay-v1` adverts eventually, but that seems like a good thing.

jmcorgan (commented):
So the concept here breaks down into a few capabilties for the node operator to choose from.

Outbound connections:

1. The user wants to only outbound connect to static  UDP, TCP, or .onion peers with known, public IP addresses. This is what we have now.

2. The user wants #1, but the static peer listing end point address is not known ahead of time, but is instead advertised by the other end over Nostr. This is like "dynamic DNS" for FIPs nodes. This would be signaled by the via_nostr option to have the node query the configured relay to obtain the current endpoint address of the npub.

3. This is the case of #2 but the UDP endpoint advertised is behind NAT. There would be an additional step then of doing the Nostr-based rendezvous to connect to the peer.

Inbound connections:

1. The node listens on UDP, TCP, or .onion address publicly reachable, and may or may not want to advertise on Nostr for inbound connections.

2. This is #1, but the UDP endpoint is behind NAT, so in the advertisment over Nostr the information for the rendezvous is provided.

Discovery:

The user wants to make outbound connections up to a certain maximum number of sessions, and periodically scans for new endpoints on Nostr.


All this breaks down to a few configuration options and logic behind them.

SatsAndSports:
Probably not very useful to others, but - in my own fork of this repo - I decided to write up my understanding in my own words of the requirements around #34 and #37: https://github.com/SatsAndSports/fips/issues/3 

Sharing just for transparency, in the unlikely event that either of you find it interesting or useful

jmcorgan:
Hey Tom, just checking in. The structural rework I asked for is a heavier lift than the last round, so no rush, wanted to ask whether you're planning to take it on. For what it's worth, folding in #34 does need to happen as part of this PR rather than as a follow-up, since the advert/offer schema is shared between the two and we'd end up reshaping it twice otherwise. Just want to make sure we're aligned before you invest more time.

tidley:
Hi Johnathan, I'll have a new commit ready today with all the changes in.

tidley:
This PR now delivers Nostr-mediated overlay endpoint discovery + UDP NAT traversal as one integrated change set (closing #37 and rolling in #34 scope).

  What changed:
  - Generalized overlay adverts (`fips-overlay-v1`) for UDP/TCP/Tor endpoints, with `udp:nat` support.
  - Added config surface and validation:
    - per-transport `advertise_on_nostr`
    - UDP `public`
    - peer `via_nostr`
    - discovery policy (`disabled` / `configured_only` / `open`) with bounded open-discovery queueing.
  - Enforced advert freshness/expiry semantics and app/protocol scoping.
  - Set advert defaults to TTL `3600s` and refresh `1800s`.
  - Added private-subnet (RFC1918/ULA) local candidate support and parallel LAN/reflexive punch planning.
  - Started transports before Nostr advert activity.
  - Removed `examples/nostr-bootstrap` (prototype path) in favor of the in-tree runtime.
  - Updated docs/config docs/protocol docs to match shipped behavior.

jmcorgan (commented):
Excellent. I likely won't be able to give this a review until tomorrow or Friday.

jmcorgan:
Thanks Tom, this was a substantial rework and I appreciate you taking
it on. I've done a full pass on the rework commits (`75e5c3f` through
`cecd761`) and everything structural checks out. The generalized
`fips-overlay-v1` advert schema with per-transport `{transport, addr}`
endpoints, per-transport `advertise_on_nostr` flags, startup order
(transports before Nostr), config validation, dynamic advert building,
layering (zero nostr imports in transport modules), and the
`via_nostr` dial path with static-first priority all verified
end-to-end. The `examples/nostr-bootstrap/` sub-crate is gone. PR
closes #37 and rolls in #34 as intended.

A handful of smaller items remain before I can merge. Master has also
advanced since your last rebase and picked up a couple of changes that
touch the same code paths, so a rebase is needed anyway. Details
below.

## Change Requests

### 1. STUN defaults: remove your personal host (HIGH)

`default_stun_servers()` currently includes `fips.tomdwyer.uk:3478`.
Could you drop it from the upstream defaults and keep it in your own
config? I don't want to ship a personal host as a default that every
downstream user will hit.

### 2. Naming: reflect the broader scope this feature has grown into

The PR started life as a UDP NAT bootstrap and has since grown into a
general Nostr-mediated peer discovery mechanism, with UDP hole
punching as one capability among several (public UDP, TCP, Tor
adverts, `via_nostr` fallback dialing). The config namespace already
reflects that with `node.discovery.nostr`, but the code-level naming
still reads like the original UDP-bootstrap scope: `nostr-bootstrap`
feature, `src/bootstrap/` module, `NostrBootstrap` /
`NostrBootstrapConfig`. I'd like to rename to match the feature's
current shape:

- Feature: `nostr-bootstrap` → `nostr-discovery`
- Module: `src/bootstrap/` → `src/discovery/`
- Structs: `NostrBootstrap` → `NostrDiscovery`,
  `NostrBootstrapConfig` → `NostrDiscoveryConfig`

### 3. Offer/answer schema cleanup

`TraversalOffer` / `TraversalAnswer` still carry `app` and `eventKind`
fields from the pre-rework schema. They're inconsistent with the
cleaned-up `OverlayAdvert` (which derives envelope metadata from the
Nostr event). Could you remove them?

### 4. `seen_sessions` pruning

`advert_cache` is properly bounded (TTL expiry + 2048-entry cap).
`seen_sessions` is pruned only on access, so it grows unbounded for a
node that receives many offers but accepts few. Could you add either
periodic cleanup (tick-driven) or a max-size cap matching
`advert_cache`? Either works for me.

### 5. `AccessDenied` propagation during the rebase (NEW)

This one wasn't visible at review time. It comes from the peer-ACL
merge (`745b523`, PR #50) that landed after your last rebase.
Master's `initiate_peer_connection` bails out immediately when
`initiate_connection` returns `NodeError::AccessDenied`, because the
ACL is keyed per-npub and continuing to try other addresses is wasted
effort (and generates extra rejected handshakes that the responder
has to reject again at msg3). Your refactored
`attempt_peer_address_list` logs-and-continues on every `Err`, which
would swallow `AccessDenied` after the rebase and try every
configured address in turn.

While resolving the conflict in `src/node/lifecycle.rs`, could you add
an early-return arm to the inner match on `initiate_connection()`'s
result?

```rust
Err(e @ NodeError::AccessDenied(_)) => return Err(e),
```

Small and mechanical, just want to flag it so it doesn't get lost
during conflict resolution.

## Rebase Summary

Master is at `be0708a` (10 commits ahead of your base `83b20b3`). I
did a test merge locally on my end; it surfaces four conflicts:

- **`src/node/lifecycle.rs`**: imports (additive, trivial) and one
  real semantic conflict in `initiate_peer_connection`. Your
  refactored helper is the right direction structurally. Keep your
  structure and fold in the `AccessDenied` propagation above.
- **`src/node/handlers/rx_loop.rs`**: master added a
  `self.reload_peer_acl()` call and inlined the epoch-millis
  computation; you extracted `Self::now_ms()`. Keep your helper and
  add master's `reload_peer_acl()` call alongside it.
- **`src/node/tests/mod.rs`**: trivial. You added `mod bootstrap;`,
  master added `mod bloom_poison;`. Keep both.
- **`Cargo.lock`**: regenerate.

Non-conflicting but worth knowing about:

- `src/node/handlers/handshake.rs` now runs the peer-ACL check at
  inbound handshake entry (PR #50). Your outbound path feeds this.
- `src/node/bloom.rs` now validates FilterAnnounce fill ratio on
  ingress. Discovery-synthesized peers that cross a saturated-filter
  peer will have those announces rejected.
- `src/protocol/tree.rs` tightened TreeAnnounce ancestry validation.
- `src/upper/dns.rs` added an arrival-interface filter on the
  in-process DNS responder. Doesn't interact with the bootstrap
  path, just a heads-up.

If you'd prefer to rebase onto master rather than merge, that keeps
the history linear. Your one existing merge commit (`4ee07fb`) was
the last one, so a rebase now would be a one-time cleanup. Totally up
to you though.

## Nice-to-have Follow-ups (not blocking)

- `PeerConfig.addresses` `#[serde(default)]` so `via_nostr`-only peers
  can omit the field entirely in YAML rather than `addresses: []`.
- Cross-field validation: `punch_start_delay_ms` <
  `attempt_timeout_secs * 1000`. Defaults are safe; user-provided
  values are unchecked.
- SatsAndSports's ephemeral-pubkey suggestion for the offer/answer
  relay flow. Nice privacy improvement, fine as a follow-up PR.

## On My End

I'm going to spend some session time walking through the feature
hands-on, both the bundled `testing/nat/` docker harness (cone /
symmetric / LAN scenarios) and a two-node test against a live relay,
to shake out any ergonomic issues with the real config surface. I'll
report back on here if anything turns up that wasn't obvious from the
code review. Nothing blocking from that yet; the five items above are
what I'd need to see land before I merge.

Thanks again for the rework. Heavy lift and the result is a much
cleaner foundation to build the rest of the discovery story on.

tidley:
Many thanks for the review, latest rebase:

- remove personal STUN default from upstream config/docs
- rename nostr-bootstrap to nostr-discovery across features/modules/types/docs
- remove stale traversal offer/answer fields
- bound seen_sessions
- preserve AccessDenied propagation after rebase
- update CI/build/test harness references to the renamed discovery feature/module paths

tidley:
Sorry I just remembered another relay I run was still in the code - now removed.

jmcorgan:
Hey @tidley — rather than queue up another round of review comments, I've pushed a single maintainer follow-up commit (`b279964`) directly on top of your `f4507fc`. It's a fast-forward update on `tidley:issue37`, so CI will re-run and no rebase is needed on your end.

The commit groups the follow-up work into the following sections:

### Repository hygiene

- Remove inadvertently committed FIPS credentials and the sample `fips.yaml` from the tree root. Root-anchor `.gitignore` entries so `packaging/` and `examples/` configs stay tracked.
- Restore AI-agent `.gitignore` entries (AGENTS.md, CLAUDE.md, `agents/`) that had been collapsed to just `.claude/`. Drop the dead "Vendored experimental bootstrap subcrate" block since `examples/nostr-bootstrap/` was deleted.
- Move `testing/nat/generated-configs/` ignore out of the root into `testing/nat/.gitignore`, matching per-harness convention.
- Drop the `.planning/` gitignore entry — not a recognized agentic-tooling convention and no such path has ever been tracked.
- Remove the stale `docs/proposals/issue-37-status-and-closure-plan.md` proposal; it captured a pre-implementation plan that no longer maps to what shipped.

### Defaults and dependencies

- Replace contributor-operated advert/DM relays and STUN servers with widely-operated public infrastructure: Damus, nos.lol, and Primal for both advert and DM relays; Google, Cloudflare, and Twilio for STUN. Ensures a fresh-install node can reach its defaults without maintainer-specific DNS.
- Bump `nostr` and `nostr-sdk` from 0.39 to 0.44. Migrations applied: `Options` → `ClientOptions`, `.gossip(bool)` removed (now a builder opt-in), `send_event_to(..., &Event)`, `EventBuilder::delete(EventDeletionRequest)`, `Timestamp::as_u64` → `as_secs`, `PublicKey::to_bech32` is now infallible. No yanked crates and no open CVEs across the 0.39→0.44 range. The 0.44 gift-wrap tightening (rumor-author-must-match-seal-author) is sidestepped by design — the hand-rolled unwrap in `src/discovery/nostr/signal.rs` trusts `seal.pubkey` exclusively.

### Cross-platform correctness (macOS)

- Restore macOS Ethernet transport construction. The PR had tightened `#[cfg(unix)]` to `#[cfg(target_os = "linux")]` on the Ethernet block in `src/node/mod.rs`, silently disabling Ethernet on macOS despite the BPF implementation in `socket_macos.rs`.
- Revert three unrelated macOS TUN shutdown-pipe changes (pipe fail-closed behavior, shutdown ordering reversal, rename `tun_shutdown_fd` → `tun_reader_shutdown_fd`). These are out of scope for this PR and I'll land them separately.
- Restore the two-line doc comment on `tun_reader_shutdown_fd`.

### Resource and configuration hygiene

- Hoist three Nostr discovery resource caps from module-level `const` into `NostrDiscoveryConfig` so operators can tune without a rebuild: `max_concurrent_incoming_offers` (default 16) bounds the semaphore used as inbound-offer rate limit; `advert_cache_max_entries` (default 2048) bounds the overlay advert cache; and `seen_sessions_max_entries` (default 2048) bounds replay-detection session-id retention. `ADVERT_CACHE_STALE_GRACE_MULTIPLIER` stays as a const since it governs protocol grace semantics tied to advert TTL, not deployment tuning.
- Feature-gate `poll_nostr_discovery` at both the method definition and its tick-loop call site so a non-feature build contains neither the method nor a no-op await in the hot path.
- Log `UdpTransport::drop` when the Drop path finds a lingering recv task or socket so cleanups on error paths and test teardown are visible in tracing rather than silent.

### Observability and logging

- Demote `debug!` → `trace!` on gift-wrap unwrap failure in `src/discovery/nostr/runtime.rs`; this fires for every SIGNAL_KIND event on the relay, including events addressed to other deployments sharing the same relay.
- Demote `warn!` → `debug!` on traversal-offer handler failure; `BootstrapError` covers expected conditions (punch timeout on symmetric NAT), and real problems are already reported upstream via `BootstrapEvent::Failed`.
- Sharpen the `warn!` on inbound-offer rate limiting to name `max_concurrent_incoming_offers` and state clearly that the offer was rate-limited rather than failing. Kept at warn because the 16-slot default is unlikely to be hit by accident.
- Add entry-point `debug!` to `connect_peer`, tagging `peer_npub` so downstream STUN, punch, and handshake logs for the same peer can be correlated.
- Add `debug!` on responder-side punch success, mirroring the existing initiator-side log.
- Add `debug!` when the advert cache or seen-sessions cache overflows its configured cap; silent eviction previously masked mis-sized limits.
- Add `debug!` at entry to `adopt_established_traversal` and `info!` on successful return. The bootstrap handoff is a major per-peer state transition and previously had no logs on the happy path.
- Add `debug!` in `cleanup_bootstrap_transport_if_unused` when the reference-count check drops a transport; pairs with the `UdpTransport::drop` log so the bootstrap-transport lifecycle is traceable end-to-end.
- Add `debug!` on STUN binding success tagging the server that answered and the reflexive address observed.

### CI

- Run the 18 Nostr discovery unit tests on Linux. The build step enabled `"gateway nostr-discovery"` but the unit-test step only enabled `"gateway"`, so the feature-gated test module compiled out and never ran.
- Add three NAT lab matrix entries (`nat-cone`, `nat-symmetric`, `nat-lan`) to the integration workflow. The harness existed but no CI entry fired it.
- Tear down profile-gated services in NAT harness cleanup; `docker compose down` was skipping them, leaving orphan containers after runs.

### Documentation

- README: move Nostr-mediated overlay endpoint discovery and UDP hole punching out of "Near-term priorities" into "What works today".
- `docs/design/fips-intro.md`: replace the paragraphs asserting that Nostr-based endpoint discovery is "planned" and that NAT traversal "is not currently addressed by the protocol" — both claims now stale.
- `docs/design/fips-transport-layer.md`: drop the "(future direction)" qualifier from the Nostr Relay Discovery section header, expand the mechanism description to cover the `udp:nat` advertisement path and the bootstrap handoff, and update the Current State callout.
- `docs/design/fips-configuration.md`: document the full `node.discovery.nostr.*` configuration surface, including the three hoisted resource caps.
- `docs/design/fips-mesh-layer.md`: append a sentence to the roaming discussion clarifying that mid-session NAT rebinding (handled by roaming) is distinct from initial NAT traversal (handled by this feature).
- `docs/proposals/README.md`: add a Status column, mark the hole-punching proposal Implemented, note that implemented proposals are retained as protocol references until superseded.
- `docs/proposals/nostr-udp-hole-punch-protocol.md`: add a Status callout at the top.
- `CHANGELOG.md`: add a "Nostr-Mediated Discovery and NAT Traversal" entry under Unreleased > Added.

### Formatting

- `cargo fmt` fixup on one touched closure body.

A dedicated `fips-nostr-discovery.md` design doc is planned post-merge to consolidate this documentation; the edits above are intended to keep the existing doc set internally consistent until that lands.

Thanks again for the rework.

PR #56, opened 2026-04-09
feat: add tun-support feature flag for mobile platforms

k0sti:
## Summary
- Add `tun-support` feature flag to make TUN/DNS/ICMP/rtnetlink dependencies optional
- Gate TUN/DNS initialization, cleanup, and rx_loop arms behind the feature flag
- Enable building fips for mobile platforms (Android/iOS) that don't support TUN interfaces

## Test plan
- [ ] Verify build with `tun-support` enabled (default, existing behavior)
- [ ] Verify build without `tun-support` for mobile targets
- [ ] Review gated code paths for correctness

> Draft PR for visual diff review — not ready for merge.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,6 +76,7 @@ tokio = { version = "1", features = ["test-util"] }
 [[bin]]
 name = "fipsctl"
 path = "src/bin/fipsctl.rs"
+required-features = ["tun-support"]
 
 [[bin]]
 name = "fipstop"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,9 +9,10 @@ repository = "https://github.com/jmcorgan/fips"
 readme = "README.md"
 
 [features]
-default = ["tui", "ble"]
+default = ["tui", "ble", "tun-support"]
 tui = ["dep:ratatui"]
 ble = ["dep:bluer"]
+tun-support = ["dep:tun", "dep:rtnetlink"]
 
 [dependencies]
 ratatui = { version = "0.30", optional = true }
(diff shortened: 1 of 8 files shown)


jmcorgan:
Thanks for this — making TUN optional for mobile is the right direction, and I appreciate you taking it on.

A few things to address before this can move forward:

**Architecture**: The 20+ `#[cfg(feature = "tun-support")]` annotations scattered across `node/mod.rs` struct fields will be fragile to maintain. A cleaner approach would be to extract TUN-related state (tun_state, tun_name, tun_tx, tun_outbound_rx, tun_reader_handle, tun_writer_handle, dns_identity_rx, dns_task, icmp_rate_limiter, host_map) into a dedicated sub-struct, then hold it as `Option<TunContext>` — one cfg gate on the type, not twenty on fields.

**Config struct duplication**: The `#[cfg(not(feature = "tun-support"))]` block in `lib.rs` copy-pastes `DnsConfig` and `TunConfig`. These will drift from the real definitions. Instead, move the config structs (which are just serde types, no platform deps) to a shared location that both paths can import.

**Return type mismatch**: `tun_state()` returns `TunState` with the feature and `&str` without — callers can't be generic across both. Keep the return type consistent.

**fipsctl**: It's gated behind `tun-support` in Cargo.toml but has nothing to do with TUN — it's the control socket CLI. This would break `--no-default-features` builds unnecessarily.

**FIPS_OVERHEAD**: The move to `protocol::mod.rs` is fine, but the original in `upper::icmp` needs to be removed (or re-exported) to avoid two definitions.

**Scope**: The `set_control_channel()` / `control_rx` embedding API is a useful addition but is orthogonal to the TUN feature flag. Would be cleaner as a separate small PR that can land independently.

Overall the goal is solid. The main rework is consolidating the cfg gates into a single TUN context type rather than sprinkling them across the node struct. Happy to discuss the approach further.

PR #62, opened 2026-04-13
Switch std::atomic to portable_atomic for mips support

andrewheadricke:
A simple change switch atomic to portable_atomic to support mips targets. A quick compile and test run on x86_64 and mips seems to run fine.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -985,6 +985,7 @@ dependencies = [
  "hex",
  "hkdf",
  "libc",
+ "portable-atomic",
  "rand 0.10.0",
  "ratatui",
  "rtnetlink",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ futures = "0.3"
 simple-dns = "0.11.2"
 socket2 = { version = "0.6.2", features = ["all"] }
 tokio-socks = "0.5"
+portable-atomic = { version = "1", features = ["std"] }
 
 rustables = { version = "0.8.7", optional = true }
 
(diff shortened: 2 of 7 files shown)


Origami74:
@jmcorgan This looks good to me.


@andrewheadricke would you be willing to uncomment the mips as a target in the package-openwrt.yml? Then it'll be included in the release process automatically.

andrewheadricke:
I've done that, but i'm not sure if the ble/bluer/libdbus issue may cause the build to fail.

Also, the mips-linux-musl target is no longer supported by rust, so i'm not sure if it will be pulled in automatically.

Edit: I force-pushed a rebase to my local fork to update the PR, but i'm reading the above that it updated jmcorgan/fips instead. Apologies if i've broken it.

Origami74:
I wasn't aware of that. And indeed the packaging workflow fails

```
error: toolchain '1.94.1-x86_64-unknown-linux-gnu' has no prebuilt artifacts available for target 'mipsel-unknown-linux-musl'
note: this may happen to a low-tier target as per https://doc.rust-lang.org/nightly/rustc/platform-support.html
note: you can find instructions on that page to build the target support from source
```

Looking into the docs it seems like it can be resolved by building inside an openwrt container.

2 options we have now:
- remove MIPS from the workflow again so we can merge this PR and people can at least build it for MIPS locally (if all workflows succeed)
- create an new workflow for MIPS so that it can be included in the releases.

I'll leave it up to you

andrewheadricke:
No problem, ill remove the mips workflow tomorrow. We'll just get the portable-atomic in and deal with the builds separately.

andrewheadricke:
Ok, reverted to just the portable_atomic changes.

Origami74 (approved):
Ready for merge

CC @jmcorgan

PR #63, opened 2026-04-13
Arch packaging fix (aur): fix namcap issues in PKGBUILD

sh1ftred:
Changes made to fix namcap issues:

1. Added `dbus` to dependencies - the fips binary links against libdbus-1.so.3, so dbus must be listed as a runtime dependency.

2. Disabled split debug package - added `!debug` to options to prevent creating a broken -debug package. The debug package was generating symlinks with incorrect paths, causing namcap errors.

3. Made config file world-readable - changed fips.yaml permissions from 0600 to 0644.

Remaining namcap warnings (can be ignored):
- "Unused shared library '/usr/lib64/ld-linux-x86-64.so.2'" - False positive. The dynamic linker is required for the binary to run.
- "Dependency libgcc detected and implicitly satisfied" / "gcc-libs may not be needed" - These are contradictory. The binaries do link to libgcc_s.so.1, and gcc-libs provides it. Keeping an explicit dependency is correct.

### DIFF
--- a/packaging/aur/PKGBUILD-git
+++ b/packaging/aur/PKGBUILD-git
@@ -1,6 +1,6 @@
 # Maintainer: Johnathan Corgan <jcorgan@corganlabs.com>
 pkgname=fips-git
-pkgver=0.2.0.rel.r41.g1e4f375
+pkgver=0
 pkgrel=1
 pkgdesc="Distributed, decentralized network routing protocol for mesh nodes"
 url="https://github.com/jmcorgan/fips"
@@ -58,7 +58,7 @@ package() {
     install -Dm0644 packaging/debian/fips-dns.service "$pkgdir/usr/lib/systemd/system/fips-dns.service"
 
     # Config files (from packaging/common/)
-    install -Dm0644 packaging/common/fips.yaml "$pkgdir/etc/fips/fips.yaml"
+    install -Dm0600 packaging/common/fips.yaml "$pkgdir/etc/fips/fips.yaml"
     install -Dm0644 packaging/common/hosts "$pkgdir/etc/fips/hosts"
 
     # System integration (from local source files in $srcdir)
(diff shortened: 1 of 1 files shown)


jmcorgan:
Thanks — `dbus` dep and `!debug` option both look correct.

One request: keep `fips.yaml` at 0600. Every other FIPS packaging target (deb, systemd, ipk, and the release `PKGBUILD`) uses 0600, and once ACLs land (PR #50) the file carries per-peer allow/deny rosters.

@dskvr, you originally authored the AUR packaging — could you take a look before we merge?

packaging/aur/PKGBUILD-git
```diff
@@ -58,7 +58,7 @@ package() {
     install -Dm0644 packaging/debian/fips-dns.service "$pkgdir/usr/lib/systemd/system/fips-dns.service"
 
     # Config files (from packaging/common/)
-    install -Dm0600 packaging/common/fips.yaml "$pkgdir/etc/fips/fips.yaml"
```
dskvr, line 61: Permissions should be 600 on fips.yaml

packaging/aur/PKGBUILD-git
```diff
@@ -1,12 +1,12 @@
 # Maintainer: Johnathan Corgan <jcorgan@corganlabs.com>
 pkgname=fips-git
-pkgver=0.1.0.r19.g8c349f5
```
dskvr, line 3: Since we're tracking this the primary repo, it should probably be`pkgver=0` since it's a computed value;  `fips-git` always overrides the header value with computed value from `pkgver()` anyways. 

`pkgver=0` would better align with Arch WIKI VCS Guidelines, and it should have been like this from the start (my bad)

dskvr (changes requested):

packaging/aur/PKGBUILD-git
sh1ftred: Thank you. My bad.

packaging/aur/PKGBUILD-git
sh1ftred: Yes, I wasn't super sure about this change. Thank you.

dskvr (approved):
LGTM

sh1ftred:
Thanks for the comments. @jmcorgan @dskvr.

PR #70, opened 2026-04-15
Validate various properties of an announcement - e.g. that the final ('root') must be the smallest node_addr in the coordinate

SatsAndSports:
I noticed that I could trick an honest node into accepting an ancestry where the smallest node_addr was _not_ at the end of the list. More details at the end of this description

I'm not certain what the intended behaviour is. This PR adds the following to these files:

 - `src/protocol/tree.rs` adds a new method (with tests) `TreeAnnounce.validate_semantics` which checks the following:
   - first node in the ancestry must equal the node_addr of the signer
   - if `is_root`, then the ancestry must be a list of just one item
   - the final node_addr in the ancestry must be the smallest node_addr
 - `src/tree/mod.rs` Error types for that new method
 - `src/node/tree.rs` apply the check, discarding invalid announcements
 - `src/node/tests/spanning_tree.rs` test that the validation is correctly applied, i.e. a node rejects an invalid announcement

=====

I was able to setup a two-node setup where the 'honest' node believed this was its coordinate, because the dishonest node announcement and ancestry where the final node wasn't the node with the smallest node_addr:

```
$ docker exec fips-node-b fipsctl show tree
{
  "declaration_sequence": 3,
  "declaration_signed": true,
  "depth": 3,
  "is_root": false,
  "my_coords": [
    "8e302c287eb48d6268122fda760b7727",  # node_addr of this, the honest node
    "69e08d65cc3a6b9c2c2ac4bd405e4b0e",  # node_addr of the dishonest peer
    "00000000000000000000000000000000", # 'fake node': smallest node_addr in this coordinate
    "00000000000000000000000000000001"  # 'fake node': it's the 'root', even though its node_addr is _not_ the smallest in this coordinate
  ],
  ...
 "root": "00000000000000000000000000000001",
  ...
```
```

### DIFF
--- a/src/node/tests/spanning_tree.rs
+++ b/src/node/tests/spanning_tree.rs
@@ -5,6 +5,8 @@
 //! reused by bloom filter tests.
 
 use super::*;
+use crate::protocol::TreeAnnounce;
+use crate::tree::{CoordEntry, ParentDeclaration, TreeCoordinate};
 
 /// A test node bundling a Node with its transport and packet channel.
 pub(super) struct TestNode {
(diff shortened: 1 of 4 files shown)


jmcorgan:
Hi @SatsAndSports, thanks for this. The ancestry validation is a useful tightening to match the spec. I've landed it on `maint` as [`b36966b`](https://github.com/jmcorgan/fips/commit/b36966b) for the upcoming v0.2.1 release, and merged forward through `master` and `next`, so it ships in the next release and continues through the development branch.

You're preserved as the commit author, with me added as co-author. Closing this PR as integrated. Thanks again for the contribution.

PR #75, opened 2026-04-16
My attempt to fix DNS queries from systemd-resolved on Ubuntu 22

hzrd149:
I do not know what this code does since I worked with codex to debug and fix the DNS resolving on my Ubuntu 22 system

I'm opening this PR in hopes that some code changes in this can be useful in finding the actual bug

### DIFF
--- a/packaging/common/fips-dns-setup
+++ b/packaging/common/fips-dns-setup
@@ -35,10 +35,8 @@ save_backend() {
 }
 
 get_fips_interface_dns_addr() {
-    ip -6 -o addr show dev "$FIPS_INTERFACE" scope global 2>/dev/null | while read -r _ _ _ addr _; do
-        echo "${addr%%/*}"
-        return 0
-    done
+    ip -6 -o addr show dev "$FIPS_INTERFACE" scope global 2>/dev/null \
+        | awk 'NR==1 {split($4, a, "/"); print a[1]; exit}'
 }
 
 # Wait for fips0 interface (up to 30s).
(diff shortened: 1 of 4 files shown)


jmcorgan:
Thanks for the report, this was a real bug. I got it reproducing on Ubuntu 22 (systemd 249) in a Docker container and traced what was happening.

`fips-dns-setup` runs `resolvectl dns fips0 127.0.0.1:5354`. On systemd 249, resolved routes per-link DNS servers through the link itself, so it tries to reach 127.0.0.1 via fips0. fips0 only has `fd00::/8` routes, so the connection gets refused and the query disappears. Newer systemd handles the loopback case specially, which is why it only shows up on 22.

The actual fix ended up much smaller than the Codex patch. TCP, SOA/NS, EDNS, the bigger buffer, the multi-bind stuff, none of it relates to this bug. resolved is just doing UDP forwarding against a per-domain match.

I put a minimal version on branch `fix-dns-ubuntu22`. One commit, about 50 lines:

1. DNS default bind goes from `127.0.0.1` to `::`, with `IPV6_V6ONLY=0` set explicitly so v4 clients still work.
2. `fips-dns-setup` grabs fips0's IPv6 address and points resolved at that instead of 127.0.0.1.
3. Dropped a hardcoded `bind_addr: "127.0.0.1"` from the packaged `fips.yaml` that was shadowing the new default.

After that, `dig @127.0.0.53 AAAA <npub>.fips` works on Ubuntu 22.

Can you pull that branch and confirm it works on your end too? If yes I'll cut a real PR and close this one.

jmcorgan:
Also, i still need to verify this doesn't break any other deb-based installs.

hzrd149:
Your changes worked for me, thanks for investigating this :+1: 
Also I had a feeling the whole TCP DNS server was just random things the agent was trying. closing this in favor of #77

PR #80, opened 2026-05-09
noise: switch ChaCha20-Poly1305 backend to ring (BoringSSL asm) — 2.5× throughput on aarch64

mmalmi:
## Summary

The `chacha20` crate (RustCrypto) ships SSE2 + soft backends only — on aarch64 (Apple Silicon, ARM Linux servers, Docker on M-series Macs) it falls through to a portable software impl at ~600–800 MB/s/core. `ring 0.17` wraps BoringSSL's hand-tuned ChaCha20-Poly1305, which dispatches to NEON on aarch64 and AVX2/AVX-512 on x86_64 — typically 3-5 GB/s/core on the same hardware.

Same wire format — ChaCha20-Poly1305 is byte-deterministic for a given `(key, nonce, plaintext, aad)`, and the full noise test suite (IK + XK roundtrip, replay window, multi-message nonce sequencing, 100-message stress) passes at **1129 / 1129**.

## Bench numbers

Docker e2e from a downstream consumer of this crate, DURATION=10, identical hardware before/after, aarch64 Linux on Apple Silicon:

| | before (RustCrypto soft) | **after (ring NEON)** | factor |
|---|---|---|---|
| 2-node TCP 1-stream | 437 Mbps | **1097 Mbps** | 2.51× |
| 2-node TCP 4-stream | 439 Mbps | 1109 Mbps | 2.53× |
| 2-node TCP 8-stream | 445 Mbps | 1069 Mbps | 2.40× |
| 2-node UDP @1000 Mbit | 599 / **40% loss** | **1000 / 0% loss** | lossless at line rate |
| 3-node TCP 1-stream (A→C→B) | 438 Mbps | **1019 Mbps** | 2.33× |
| 3-node UDP @1000 Mbit | 475 / **52% loss** | **1000 / 0% loss** | lossless at line rate |
| 3-node ping under load | 7.68 ms avg / 215 ms max | 0.72 ms / 3.6 ms max | ~10× |

The relay-path lift is the cleanest tell on what was bottlenecking: the transit node was crypto-bound (single-threaded soft chacha couldn't keep up with offered rate), so the queue accumulated under load. With NEON the relay isn't crypto-bound and the queue stops accumulating — the 215ms ping-tail collapses to 3.6ms.

## Implementation notes

- `LessSafeKey` and `UnboundKey` deliberately do not implement `Clone` for safety. `CipherState`'s manual `Clone` impl rebuilds the keyed AEAD from the retained 32-byte key — cheap for ChaCha20-Poly1305 since construction is essentially a key copy + a constant-time check.
- The keyed AEAD is now **cached** in `CipherState.cipher` instead of being re-derived per packet. This was already a perf win for the chacha20poly1305 backend (`new_from_slice` per packet was hot in profiles); for ring it's a bigger win because `LessSafeKey` construction also derives the Poly1305 key.
- Public `Vec<u8>`-returning API preserved. New module-private `seal` / `open` helpers wrap ring's `seal_in_place_append_tag` / `open_in_place` so the per-packet allocation pattern lives in one place.
- `EndToEndState::Established` triggers `clippy::large_enum_variant` after the swap — `NoiseSession` grew from ~600 to ~1.5 KB because ring precomputes the Poly1305 key state at construction. That precomputation is exactly the win — boxing the variant would re-add an indirection per packet and work against it. `#[allow]`'d at the enum decl with a justifying comment.

## Dependency change

ring is widely deployed (rustls, hyper-rustls, AWS SDK) and is a pure-Rust API on top of BoringSSL's vendored asm. It adds no new C toolchain requirements beyond what any rustls user already has. `chacha20poly1305` is dropped from `Cargo.toml`.

## Test plan
- [x] `cargo build --release` clean
- [x] `cargo test --release --lib` — 1129 passed, 0 failed (the full noise / replay / handshake / 100-node session suite, which is the byte-level wire-format proof since round-trip pass is only possible if ring produces what the receiver expects)
- [x] `cargo clippy` clean on the touched files (`src/noise/mod.rs`, `src/node/session.rs`); pre-existing master clippy warnings in unrelated areas (firewall_state.rs, tun.rs, listening.rs, ethernet/socket_macos.rs) are not introduced by this PR
- [x] Docker e2e bench at downstream consumer (numbers above)

### DIFF
--- a/.github/workflows/package-openwrt.yml
+++ b/.github/workflows/package-openwrt.yml
@@ -78,7 +78,9 @@ jobs:
             rust_target: aarch64-unknown-linux-musl
             rust_channel: stable
             # MT3000, MT6000, Flint 2, RPi 3/4/5
-          # MIPS disabled: 32-bit MIPS lacks AtomicU64; needs portable-atomic crate
+          # MIPS disabled: nostr-relay-pool 0.44 uses std::sync::atomic::AtomicU64
+          # directly (fips's own atomics already use portable_atomic). Re-enable
+          # once an upstream portable-atomic patch lands (or via [patch.crates-io]).
           # - build_arch: mipsel
           #   openwrt_arch: mipsel_24kc
           #   rust_target: mipsel-unknown-linux-musl
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1056,7 +1056,6 @@ version = "0.3.0-dev"
 dependencies = [
  "bech32",
  "bluer",
- "chacha20poly1305",
  "clap",
  "criterion",
  "dirs",
@@ -1070,6 +1069,7 @@ dependencies = [
  "procfs",
  "rand 0.10.1",
  "ratatui",
+ "ring",
  "rtnetlink",
  "rustables",
  "secp256k1 0.30.0",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ ratatui = "0.30"
 secp256k1 = { version = "0.30", features = ["rand", "global-context"] }
 sha2 = "0.10"
 hkdf = "0.12"
-chacha20poly1305 = "0.10"
+ring = "0.17"
 rand = "0.10.1"
 thiserror = "2.0"
 bech32 = "0.11"
--- a/src/node/session.rs
+++ b/src/node/session.rs
@@ -14,6 +14,15 @@ use crate::noise::{HandshakeState, NoiseSession};
 use secp256k1::PublicKey;
 
 /// State machine for an end-to-end session.
+///
+/// `Established` is intentionally larger than the handshake variants:
+/// `NoiseSession` carries ring's `LessSafeKey` (×2, send + recv), each of
+/// which embeds the precomputed Poly1305 key + per-implementation AEAD
+/// state. That precomputation is exactly the win — it lets the per-packet
+/// AEAD skip key derivation and dispatch straight to NEON / AVX. Boxing
+/// the variant would add an allocation per session and double-indirection
+/// on every encrypt/decrypt, working against that win.
+#[allow(clippy::large_enum_variant)]
 pub(crate) enum EndToEndState {
     /// We initiated: sent SessionSetup with Noise XK msg1, awaiting SessionAck.
     Initiating(HandshakeState),
--- a/src/noise/mod.rs
+++ b/src/noise/mod.rs
@@ -39,10 +39,7 @@ mod handshake;
 mod replay;
 mod session;
 
-use chacha20poly1305::{
-    ChaCha20Poly1305, Nonce,
-    aead::{Aead, KeyInit, Payload},
-};
+use ring::aead::{Aad, CHACHA20_POLY1305, LessSafeKey, Nonce, UnboundKey};
 use std::fmt;
 use thiserror::Error;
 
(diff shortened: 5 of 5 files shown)


jmcorgan:
Reviewed the diff. Wire-format invariance, the `Nonce::assume_unique_for_key` contract, and the `large_enum_variant` allow on `EndToEndState` are all correct. A few follow-ups before merge.

**The `ring` dependency is the bigger decision than the diff size suggests.** A dependency change of this scope (vendored BoringSSL asm, genuinely net-new audit surface that we don't currently pull in transitively) is riskier than other changes in this PR set and warrants more attention than a perf swap normally would. At minimum I'd want a fresh dependency audit pass against the post-swap dep tree to surface any open advisories or unmaintained transitives that come along with `ring`. Have you considered `chacha20poly1305 = "0.11"` (RustCrypto's current version with newer SIMD work)? It might close a meaningful chunk of the gap on aarch64 without changing the dep surface, worth at least a side-by-side bench before committing to the `ring` swap.

**Cross-platform build verification, especially OpenWrt MIPS.** `ring` has historically had spotty MIPS support, and we cross-compile OpenWrt ipk packages for `mipsel-unknown-linux-musl`. Our packaging matrix needs more than a contributor can self-run, so I've started CI on this PR; results will land in the checks tab. The Linux/Debian, macOS, and Windows builds also need to come back clean, but OpenWrt is the canary I'm most concerned about given the asm dispatch story for `ring` is x86_64 + aarch64.

**Testing considerations.** A full local CI sweep (not just `cargo test --lib`) on your end before this is ready would help. Existing rekey integration tests should exercise the new `initialize_key` → `build_cipher` path under realistic conditions, so anything pathological with the cached cipher rebuild would surface there.

mmalmi:
Thanks for the detailed pass — addressing each of the three concerns with evidence below. TL;DR: dep audit is clean for ring; a side-by-side bench against the RustCrypto alternatives shows ring still ~2× faster on aarch64; OpenWrt MIPS works for ring itself but there are pre-existing blockers elsewhere in the tree that are independent of this swap.

### 1. Dependency audit on the post-swap tree

`cargo tree -p ring`:
```
ring v0.17.14
├── cfg-if v1.0.4
├── getrandom v0.2.17
├── libc v0.2.180
└── untrusted v0.9.0
[build-dependencies]
└── cc v1.2.54
```
Every one of those is already in the tree pre-swap (via tokio/hyper/etc).

`cargo audit` on the post-swap lockfile reports 1 vulnerability + 4 warnings, **none introduced by ring**:
- `bytes 1.11.0` — RUSTSEC-2026-0007 (via tokio/hyper)
- `instant 0.1.13` — unmaintained (via ratatui-termwiz)
- `paste 1.0.15` — unmaintained
- `rand 0.8.5` / `rand 0.10.0` — RUSTSEC-2026-0097

So: ring brings zero new audit surface.

### 2. Side-by-side vs `chacha20poly1305 = "0.11"` (and a third configuration worth measuring)

Worth noting up front: `chacha20poly1305 = "0.11.0"` isn't released yet — crates.io has `0.11.0-rc.3` (Feb 2026). Pinning a security-sensitive AEAD to an RC is itself a non-trivial dep-policy call.

Also: the "newer SIMD work" 0.11 brings is mostly the underlying `chacha20 0.10.0` which makes NEON on-by-default. But NEON has been there since `chacha20 0.9.1` (April 2023), gated behind `--cfg chacha20_force_neon` — meaning the 0.10 dep we're shipping today *can* get NEON without changing the dep, just by setting that cfg flag. So the more honest A/B is three-way.

Microbench (aarch64-apple-darwin, 1280-byte MTU, encrypt+decrypt round-trips, single thread, cipher built once = matches the cached `CipherState` fast path, 1M iters):

| Backend | ns/op | Throughput | vs ring |
|---|---|---|---|
| **ring 0.17 (BoringSSL NEON)** | 778.9 | **6573 Mbps** | 1.00× |
| chacha20poly1305 0.10 (default — soft) | 2050.7 | 2497 Mbps | 0.38× |
| chacha20poly1305 0.10 + `--cfg chacha20_force_neon` | 1629.2 | 3143 Mbps | 0.48× |
| chacha20poly1305 0.11.0-rc.3 (NEON default-on) | 1638.3 | 3125 Mbps | 0.48× |

The cfg-flag and 0.11.0-rc.3 land within 0.5% of each other — confirming what 0.11 actually changes for us is removing the cfg gate. NEON-RustCrypto closes ~26% of the gap; ring is still ~2.1× faster. The PR's downstream Docker e2e showed 2.5× — the microbench predicts that fairly well.

Bench harness (standalone, feature-gated per backend) available if you want to reproduce locally.

### 3. OpenWrt MIPS

Drove this end-to-end on a Linux/x86_64 box (since macOS/zig hits a libc gap for plain `mipsel-linux-musl`). Result is layered:

| Layer | Status on `mipsel-unknown-linux-musl` |
|---|---|
| `ring 0.17.14` | ✅ Compiles cleanly |
| `portable-atomic 1.13.1` (32-bit AtomicU64 fallback) | ✅ Compiles cleanly |
| fips's own atomics | ✅ Already routed through `portable_atomic` |
| `nostr-relay-pool 0.44.0` | ⚠️ Blocked: uses `std::sync::atomic::AtomicU64` directly in 3 files. Verified fixable with a 6-line portable-atomic patch via a local `[patch.crates-io]` shim. |
| `rustables 0.8.7` | ⚠️ Bindgen needs `BINDGEN_EXTRA_CLANG_ARGS_mipsel-unknown-linux-musl=-I/usr/include -I/usr/include/x86_64-linux-gnu` for netlink headers. Works under zigbuild path; cross-rs's mipsel container has Ubuntu 18.04 kernel headers (~kernel 4.18) that pre-date `NFTA_CHAIN_FLAGS`. |
| **All Rust crates → `.rlib`** | ✅ Every one compiled successfully under the patched zigbuild path |
| Final ELF link | ❌ Toolchain story: zig 0.16 doesn't ship libc for `mipsel-linux-musl` (only `musleabi`/`musleabihf` variants); cross-rs's container has stale kernel headers blocking rustables |

**Direct answer to the concern:** ring 0.17 is clean on MIPS. The remaining blockers are entirely pre-existing — `nostr-relay-pool` would need an upstream portable-atomic patch (or a `[patch.crates-io]` shim until that ships), and the OpenWrt cross-build tooling needs either a switch from zig to the OpenWrt SDK or a refreshed cross-rs base image. None of those land or unblock with this PR.

Pushed a small commit (`5aaafb8`) updating the comment in `.github/workflows/package-openwrt.yml` to locate the actual blocker (`nostr-relay-pool`) instead of the stale "lacks AtomicU64" framing, so the next person looking at re-enabling the matrix has the right starting point. Re-enabling the matrix itself is a follow-up.

### 4. Full local sweep + rekey path

`cargo test --release --lib` on PR head: **1131 passed, 0 failed, 1 ignored** in 77.82s. Covers `noise::tests` (IK/XK roundtrip, replay window, multi-message nonce sequencing), the 100-node session/routing/spanning-tree stress, TCP reconnection. The `initialize_key → build_cipher` path you specifically called out is exercised by both the rekey unit tests and the 100-node session — which lines up with the three integration `rekey` suites already green on this PR's CI.

jmcorgan:
Landed on master via cherry-pick as `5cda4a9` (ring AEAD swap, was `ae2828e`) and `9b1016f` (OpenWrt comment fix, was `5aaafb8`). CHANGELOG entry committed as `77ecfda` under [Unreleased] § Changed. Forward-merged to `next` as `e5071d7`. Per-commit authorship preserved as `Martti Malmi <sirius@iki.fi>`; committer line is the maintainer since cherry-pick rewrites committer SHA. Shipping in v0.3.0. Thanks for the thorough work, Martti.

PR #81, opened 2026-05-09
node: hot-path drain batching + recvmmsg + eager pubkey_full (3.5× single-stream)

mmalmi:
## Summary

Three independent perf changes that together took TCP single-stream from ~120 Mbps to ~423 Mbps on the Docker e2e bench (before any AEAD changes — see #80 separately for that). Each is small and self-contained, but they're related and easier to review together. Happy to split into three PRs if preferred.

### Commit 1: `identity: eagerly precompute pubkey_full in PeerIdentity::from_pubkey`

`PeerIdentity::pubkey_full()` was lazily deriving the full PublicKey from the x-only key on every call (`self.pubkey.public_key(Parity::Even)`), which is a secp256k1 EC point parse — `fe_sqrt` + `fe_mul` + `ge_set_xo_var`. On the bulk-data send path, perf showed ~6% of CPU per packet on this for a value that never changes after construction. The same EC point parse already runs at construction inside `NodeAddr::from_pubkey`, so the cost is paid once where it would be paid anyway.

### Commit 2: `udp: batched recvmmsg receive on Linux (32-pkt bursts)`

The UDP recv loop drained the kernel queue one packet per `recvmsg(2)`. Each call paid full per-syscall + per-task-wakeup overhead (~50 µs avg including a futex-based scheduler hop), so under sustained load the loop ran at ~one rx event per scheduler quantum — the dominant cap on inbound packet rate.

On Linux, switch the steady-state path to `recvmmsg(2)` with a 32-packet batch. A single readable() wakeup drains up to 32 datagrams in one syscall before yielding back to the reactor. Stack-allocated `mmsghdr` arrays sized to a module-level `BATCH_SIZE` constant; SO_RXQ_OVFL parsing is skipped on the batch path (the diagnostic `recv_from` path retains it).

macOS / Windows fall through to the per-packet `recv_from` loop — recvmmsg is Linux-specific and the per-packet API is fast enough on those platforms for now (recvmsg_x for Darwin can be added later).

A `send_batch` / sendmmsg counterpart is added on the sync layer for a future change to the node task; not yet wired into `run_rx_loop`.

### Commit 3: `node: drain packet_rx / tun_outbound_rx in batches in run_rx_loop`

Even with `recvmmsg` filling `packet_rx` in bursts, the `select!` macro in `run_rx_loop` was still costing one scheduler hop per packet. After the `await` on `packet_rx.recv()` / `tun_outbound_rx.recv()` fires, drain up to 256 additional ready items via `try_recv()` in a tight inner loop before yielding back to `select!`. `biased` gives the data-plane branches priority over tick / control / DNS under sustained load.

The 256 cap is empirically tuned to keep the worker on a busy stream between yield points (a contiguous burst of ~256 MTU-sized packets ≈ 400 KB of contiguous traffic) while still bounding the inner loop so a flood on one branch can't starve the periodic tick or control socket. Lower caps (64) left perf on the table; higher caps (1024+) delayed tick handling visibly.

## Bench numbers

Docker e2e from a downstream consumer, DURATION=10, identical hardware before/after, aarch64 Linux on Apple Silicon. These are the full set landed across the three commits in this PR; the numbers were taken before the AEAD swap in #80, so they isolate the rx-path effect from the crypto effect:

| | before | after this PR | factor |
|---|---|---|---|
| 2-node TCP 1-stream | 120 Mbps | 423 Mbps | 3.5× |
| 2-node UDP @200 Mbit | 178 / 11% loss | 200 / 0.0005% loss | clean |
| 2-node ping under load | 0.84 ms | 0.71 ms | wash |

Stacked on top of the ring AEAD swap (#80), the same bench reaches 1097 Mbps single-stream, lossless UDP at line rate, and the relay-path ping-tail collapses from 215ms to 3.6ms.

## Test plan
- [x] `cargo test --release --lib` — 1129 passed, 0 failed (full noise / replay / handshake / 100-node session / TCP integration suite)
- [x] `cargo clippy` clean on the touched files; pre-existing master clippy warnings in unrelated areas (`firewall_state.rs`, `tun.rs`, `listening.rs`, `ethernet/socket_macos.rs`) are not introduced by this PR
- [x] Docker e2e bench at downstream consumer (numbers above)

### DIFF
--- a/src/identity/peer.rs
+++ b/src/identity/peer.rs
@@ -24,12 +24,22 @@ impl PeerIdentity {
     ///
     /// Note: When only the x-only key is available, the full public key
     /// will be derived assuming even parity for ECDH operations.
+    ///
+    /// Precomputes the even-parity full pubkey eagerly so `pubkey_full()`
+    /// is a constant-time field load. Without this, every send-side
+    /// hot-path caller (per-packet) re-derived the full key, spending
+    /// ~6% of CPU on a secp256k1 EC point parse (`fe_sqrt` + `fe_mul` +
+    /// `ge_set_xo_var`) for what should be a memoized lookup. The same
+    /// EC point parse already runs at construction inside
+    /// `NodeAddr::from_pubkey`, so the cost is paid where it would be
+    /// paid anyway.
     pub fn from_pubkey(pubkey: XOnlyPublicKey) -> Self {
         let node_addr = NodeAddr::from_pubkey(&pubkey);
         let address = FipsAddress::from_node_addr(&node_addr);
+        let pubkey_full = pubkey.public_key(Parity::Even);
         Self {
             pubkey,
-            pubkey_full: None,
+            pubkey_full: Some(pubkey_full),
             node_addr,
             address,
         }
--- a/src/node/handlers/rx_loop.rs
+++ b/src/node/handlers/rx_loop.rs
@@ -82,14 +82,42 @@ impl Node {
 
         loop {
             tokio::select! {
+                biased;
                 packet = packet_rx.recv() => {
                     match packet {
                         Some(p) => self.process_packet(p).await,
                         None => break, // channel closed
                     }
+                    // Drain remaining ready inbound packets in a tight loop
+                    // before yielding back to select! — every yield is a
+                    // futex hop on tokio's multi-thread scheduler, and at
+                    // line rate the kernel UDP queue typically has several
+                    // datagrams available per wake. Caps at a batch
+                    // boundary so other branches (tick, control) eventually
+                    // get a turn even under sustained load.
+                    let mut drained = 0;
+                    while drained < 256 {
+                        match packet_rx.try_recv() {
+                            Ok(p) => {
+                                self.process_packet(p).await;
+                                drained += 1;
+                            }
+                            Err(_) => break,
+                        }
+                    }
                 }
                 Some(ipv6_packet) = tun_outbound_rx.recv() => {
                     self.handle_tun_outbound(ipv6_packet).await;
+                    let mut drained = 0;
+                    while drained < 256 {
+                        match tun_outbound_rx.try_recv() {
+                            Ok(p) => {
+                                self.handle_tun_outbound(p).await;
+                                drained += 1;
+                            }
+                            Err(_) => break,
+                        }
+                    }
                 }
                 Some(identity) = dns_identity_rx.recv() => {
                     debug!(
--- a/src/transport/udp/mod.rs
+++ b/src/transport/udp/mod.rs
@@ -412,6 +412,11 @@ impl Drop for UdpTransport {
 }
 
 /// UDP receive loop - runs as a spawned task.
+///
+/// On Linux, drains the kernel UDP queue in 32-packet bursts via `recvmmsg`
+/// to amortise the per-syscall + per-task-wakeup overhead. macOS / Windows
+/// fall through to single-packet `recv_from`. Either way every datagram
+/// is forwarded to `packet_tx` in arrival order.
 async fn udp_receive_loop(
     socket: AsyncUdpSocket,
     transport_id: TransportId,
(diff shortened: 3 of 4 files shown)


jmcorgan:
The pubkey eager-precompute is correct and effectively covers all three `PeerIdentity` constructors with a single change (`from_npub` calls `from_pubkey`, and `from_pubkey_full` already stored Some). A few follow-ups before merge.

**Unsafe surface in commit 2 has a stable-Rust replacement.** The slice-array construction in `udp_receive_loop` (the `MaybeUninit::uninit().assume_init()` + `transmute` pair) is technically sound, but on edition 2024 the same shape can be written without unsafe by drawing from a shared iterator into `array::from_fn`:

```rust
let mut bufs: [&mut [u8]; BATCH] = {
    let mut iter = backing.iter_mut();
    std::array::from_fn(|_| iter.next().unwrap().as_mut_slice())
};
```

Same stack array, same disjoint mutable borrows, no unsafe block, no transmute. (`MaybeUninit::array_assume_init` is still nightly-only, otherwise that would be even cleaner.) Not blocking, but if you're touching the code anyway it's worth doing.

**`SO_RXQ_OVFL` parsing dropped on the batch path is a bigger deal than the comment makes it sound.** `kernel_drops` isn't just operator-visible; it feeds `sample_transport_congestion()` in `node/handlers/forwarding.rs`, which runs every 1s from the tick path and sets a per-transport `dropping` flag based on whether new drops occurred since the last sample. That flag is the actual congestion-aware signal consumed downstream (and the per-event `congestion.kernel_drop_event` stat is what surfaces in operator visibility). With the batch path active on Linux under sustained load, which is exactly when this signal matters most, the flag stops firing and the congestion-detection path goes dark. I don't have a clean answer for what the right shape is here (parse cmsg per-msg in the batch path, sample via `recv_from` periodically, expose the counter via a different syscall, etc.), but the choice needs to be made deliberately rather than left as a comment-acknowledged speed trade. Thoughts?

**256-cap fairness in commit 3.** The cap looks fine in shape; chaos under sustained load is what would surface any starvation pathology, covered in the testing note below.

**Perf numbers lack x86_64 specifics.** The bench table is all aarch64 Linux on Apple Silicon. Follow-up benchmarking on x86_64 might be needed to confirm the wins translate.

**Testing considerations.** For changes touching the rx hot path like commit 3's drain-batching, running the full local CI sweep (not just `cargo test --lib`) plus 10 to 20 minutes each of the `maelstrom` and `maelstrom-sparse` chaos scenarios is worth doing on your end if you can. Sustained-load fairness and scheduler-hop behavior under contention are exactly what those scenarios stress, and any pathology would surface there before merge.

mmalmi:
Thanks for the careful read. Pushed an update with both follow-ups folded into the recvmmsg commit:

**`SO_RXQ_OVFL` on the batch path.** Right call — congestion-aware `dropping` flag going dark under sustained load is exactly the wrong direction. Wired a small cmsg buffer to `msgs[0]` only, walking the chain after each `recvmmsg`. The counter is socket-wide and monotonic, so a single sample per batch (one per up-to-32 datagrams) keeps the 1Hz `sample_transport_congestion()` detector well-fed under load. Cost is one stack-allocated `CMSG_SPACE(4)` buffer + one `CMSG_FIRSTHDR` walk per batch syscall — strictly cheaper than the per-packet cmsg parse the old `recv_from` path was doing. `AsyncUdpSocket::recv_batch` now returns `(count, drops)` and `udp_receive_loop` calls `stats.set_kernel_drops(drops as u64)` once per batch.

**Drop the `unsafe` block.** Replaced with your snippet — `std::array::from_fn` over a single `backing.iter_mut()`. Same disjoint borrows, no `MaybeUninit`/`transmute`.

Both changes folded into `9a95151` (the recvmmsg commit) so the PR remains a clean three-commit sequence; commit message updated to describe the cmsg sampling design.

Still on the to-do list: x86_64 bench numbers (no aarch64-only claim should stand for a perf change like this) and 10–20 min runs of `maelstrom` + `maelstrom-sparse`. Will follow up here with both.

mmalmi:
Re-ran the bench properly with a before/after split. The single-version numbers in my earlier draft of this comment didn't tell us anything about whether the PR's wins translate to x86_64 — fixing that.

Setup: ubuntu-dev (Linux 7.0, 8 vCPU x86_64), docker bridge, `DURATION=10`, identical hardware. Two scenarios: a 2-node direct mesh (`scripts/perf-docker.sh`) and a 3-node relay topology where every A↔B byte transits node C (`scripts/perf-docker-relay.sh`). "Before" is fips at `a4f9f37` (parent of this PR's first commit); "after" is `c231248` (PR 81 commits 1+2+3 plus the two related follow-up tweaks — `BATCH_SIZE` hoist + `endpoint_event_rx` drain + drain cap 64→256 — that the same downstream consumer landed alongside this PR on its fork).

**2-node direct:**

| | before | after |
|---|---|---|
| TCP 1-stream | 969 Mbps | 989 Mbps |
| TCP 4-stream sum | 997 Mbps | 1008 Mbps |
| TCP 8-stream sum | 1004 Mbps | 996 Mbps |
| UDP @1 Gbit recv | 1000 Mbps / 0% loss | 1000 Mbps / 0% loss |
| ping under load | 0.147 ms avg | 0.172 ms avg |

**3-node relay (A → C → B):**

| | before | after |
|---|---|---|
| TCP 1-stream | 270 Mbps (82 retr) | 273 Mbps (419 retr) |
| TCP 4-stream sum | 275 Mbps | 273 Mbps |
| TCP 8-stream sum | 280 Mbps | 277 Mbps |
| UDP @1 Gbit recv | 355 Mbps / 64% loss | 329 Mbps / 67% loss |
| ping (transit) | 0.582 ms avg | 0.553 ms avg |

So on x86_64 / 8 vCPU bare metal, **PR 81 is essentially a wash** — the wins don't translate. Direct mesh is line-rate before and after (saturating the docker bridge / iperf3 generation, not the recvmmsg syscall ceiling). Relay path caps at ~270 Mbps on both — the bottleneck has moved off the recv syscall (probably to AEAD CPU-per-packet given the topology), so reducing per-syscall overhead doesn't change the cap. The 3.5× aarch64 win was about the per-syscall + per-task-wakeup cost on slower-per-core silicon dominating throughput; x86_64 with 8 cores has enough headroom that the rx loop never gets behind in this Docker setup.

Not a regression though, and the 2-node direct numbers track within run-to-run noise. The clean signal that *did* show up is in chaos: the cmsg-on-msgs[0] path is alive end-to-end (`Congestion events: 46/106` across maelstrom + maelstrom-sparse), `Kernel drop events: 0/0` on both, **0 panics 0 errors** across 20 minutes / 70 nodes / 177 edges of mixed netem + churn + topology mutation. So:

|  | maelstrom | maelstrom-sparse |
|---|---|---|
| nodes / edges / duration | 20 / 67 / 600s | 50 / 110 / 600s |
| panics / errors | 0 / 0 | 0 / 0 |
| sessions / promotions / parent switches | 59 / 312 / 136 | 49 / 354 / 218 |
| congestion events | 46 | 106 |
| kernel drop events | 0 | 0 |
| iperf3 runs landed | 18 | 21 |

Net: the PR is value-neutral on x86_64 / 8 vCPU and a strong win on slower-per-core silicon. Worth merging, just don't expect every consumer's bench to show the aarch64-shaped delta.

PR #83, opened 2026-05-09
node: inherit primary UDP config when adopting NAT-traversal sockets

mmalmi:
## Summary

`Node::adopt_established_traversal` was constructing the adopted UDP transport with `UdpConfig::default()` (MTU 1280, default buffer sizes, default flags), regardless of what the operator had configured on the primary `[transports.udp]` listener. If the primary was set to a higher MTU on a path where larger packets are viable (e.g. 1500), full-sized tunnel datagrams sent over the NAT-traversed link would exceed the adopted socket's MTU and get **dropped at the socket layer with no log**.

Inherit the primary UDP config (MTU + recv/send buffer sizes + accept / advertise flags) and clear the bind / external-address fields since the adopted socket is already bound. Restores symmetric behaviour: any MTU value the operator picks for the primary listener applies to NAT-traversed links too.

## How it manifested

Discovered in a downstream consumer where `MESH_TUNNEL_MTU=1320` / encrypted wire ~1426 B produced silent packet drop on every session that had been promoted onto a NAT-traversed link. Any session that hadn't been promoted (still on the primary listener) was fine. Tail of the symptoms: TCP retransmits + degraded throughput on NAT-traversed paths only.

## Test plan
- [x] `cargo test --release --lib` — 1129 passed, 0 failed (no regressions)
- [x] `cargo clippy` clean on the touched file (`src/node/lifecycle.rs`)

### DIFF
--- a/src/node/lifecycle.rs
+++ b/src/node/lifecycle.rs
@@ -1824,17 +1824,39 @@ impl Node {
         self.register_identity(peer_node_addr, peer_identity.pubkey_full());
 
         let transport_id = self.allocate_transport_id();
-        // Adopted ephemeral UDP transports use UdpConfig::default() when the
-        // bootstrap runtime doesn't pass an override. Default MTU resolves to
-        // 1280 (IPv6 minimum), which is the only value guaranteed to survive
-        // arbitrary NAT-traversal middlebox paths. Inheriting from the named
-        // [transports.udp] config (Option 3 in ISSUE-2026-0013) would track
-        // operator config more closely but risks regressions on hostile paths;
-        // accepted as-is until a concrete use case justifies the change.
+        // Adopted ephemeral UDP transports inherit MTU + socket-buffer sizing
+        // (and accept_connections / advertise flags) from the operator's
+        // configured [transports.udp] when the bootstrap runtime doesn't
+        // pass an explicit override. Lookup tries `transport_name` first
+        // (covers the `Named` multi-listener variant) and falls back to the
+        // unnamed `Single` listener, so single- and named-listener configs
+        // both inherit cleanly.
+        //
+        // Tradeoff: `UdpConfig::default()` sets MTU 1280 (IPv6 minimum), the
+        // only value guaranteed to survive arbitrary middlebox paths.
+        // Inheriting a higher operator-chosen MTU means NAT-traversed flows
+        // initially attempt that MTU and may black-hole on tighter paths
+        // until reactive `MtuExceeded` recovery kicks in. Operators who
+        // raise the primary MTU based on known-clean topology accept that
+        // tradeoff; the silent drop on a too-low default was strictly
+        // worse for the common case where the primary MTU is reachable.
+        //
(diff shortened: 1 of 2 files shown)


jmcorgan:
Thanks for these, Martti. The use case is real and your fix is going in the right direction. Couple of things I'd like worked out before this lands, plus a small docs touch.

The biggest one: operators with multiple named UDP listeners don't pick up the inheritance. `lookup_udp_config(None)` only finds the config when there's exactly one UDP listener (the `Single` variant). If someone has named listeners (the `Named` variant), the lookup misses and the adopted transport drops back to the 1280 default. The `EstablishedTraversal` already carries `transport_name`, so passing that through first and falling back to `None` only when there's no named match would close the gap.

Second, there's no test that pins the new behavior. All three of the existing adoption tests in `src/node/tests/bootstrap.rs` pass an explicit config in the `EstablishedTraversal`, so none of them exercise the path you're changing. Extending one (or adding a sibling) to set `mtu: Some(1500)` on the primary config and asserting the adopted transport's link reports 1500 would lock the new behavior in against later regressions. Quick to add.

Last thing, just one sentence in the comment. The previous default of 1280 was deliberately the IPv6 minimum, which is the only MTU guaranteed to survive arbitrary middlebox paths. With your change, operators who set their primary listener higher (based on known-clean LAN topology) will have NAT-traversed flows initially attempting that higher MTU and possibly black-holing on tighter paths until reactive `MtuExceeded` recovery kicks in. That's the right call for your scenario, but worth saying out loud in the comment so a future reader understands why the conservative default went away.

The test and comment are quick. The named-listener gap is the one I'd want sorted out before merge. Appreciate the contribution.

mmalmi:
Thanks for the review. Pushed an update with all three fixes folded into the same commit:

- **Named-listener gap.** `adopt_established_traversal` now looks up the inherited config with `traversal.transport_name.as_deref()` first and falls back to the unnamed `Single` listener only when there's no named match. Operators with multi-listener configs pick up inheritance from the matching `[transports.udp.<name>]`.
- **Test coverage.** Added two sibling tests in `src/node/tests/bootstrap.rs`: `test_adopted_udp_inherits_mtu_from_single_primary_config` and `test_adopted_udp_inherits_mtu_from_named_primary_config`. Both set `mtu: Some(1500)` on the primary config (Single / Named respectively), adopt without an explicit `transport_config`, and assert the adopted transport's MTU reports 1500.
- **Comment.** Expanded the inline comment to call out the deliberate-1280-as-IPv6-minimum tradeoff and that operators raising the primary MTU on known-clean topology accept the initial-attempt-then-`MtuExceeded`-recovery behaviour on tighter NAT-traversed paths.

Also rebased on the format-check fix.

jmcorgan:
Thanks Martti. All three review follow-ups (named-listener lookup, sibling regression tests, hostile-path tradeoff comment) addressed cleanly. Cherry-picked onto master as f0bb29f with you as author. Closing the PR.

PR #86, opened 2026-05-12
ci: fix AUR publish for fips, add fips-git publish workflow

Origami74:
## Summary

Two AUR-related workflow changes:

### `aur-publish.yml` (stable `fips`)
- Fixes AUR publish failure where the source tarball was committed into the AUR repo and rejected by AUR's 488 KiB max-blob hook. Root cause: `updpkgsums: true` caused `makepkg` to download `fips-<ver>.tar.gz` into the AUR working tree, where it was then staged by the deploy action.
- Pre-computes the source b2sum in the workflow (fetch + `b2sum`), patches `pkgver` and the `b2sums=('SKIP'…)` placeholder in-place, and publishes with `updpkgsums: false` so the AUR clone stays metadata-only.
- Adds a `workflow_dispatch` `tag` input so historical tags can be re-published manually (used successfully to publish [v0.3.0](https://aur.archlinux.org/packages/fips) → `0.3.0-1`).
- Drops `continue-on-error: true` so future regressions surface in CI.

### `aur-publish-git.yml` (new — VCS `fips-git`)
- Adds a workflow that publishes the `fips-git` VCS PKGBUILD to AUR. Triggered only on master pushes that touch `PKGBUILD-git` / its companion files, plus `workflow_dispatch`. Not tied to release tags — pkgver is computed at build time by the PKGBUILD's `pkgver()` function.
- Reuses the same `AUR_EMAIL` / `AUR_SSH_PRIVATE_KEY` secrets; `pkgname: fips-git` selects the right AUR repo.

## Test plan
- [x] Manually dispatched against `v0.3.0` from this branch — AUR now shows `fips 0.3.0-1`.
- [ ] After merge, manually dispatch `aur-publish-git.yml` to publish `fips-git` for the first time.
- [ ] Next real release tag push verifies the automatic path for stable `fips` end-to-end.

### DIFF
--- a/.github/workflows/aur-publish-git.yml
+++ b/.github/workflows/aur-publish-git.yml
@@ -10,7 +10,6 @@ on:
       - 'packaging/aur/fips.sysusers'
       - 'packaging/aur/fips.tmpfiles'
       - 'packaging/aur/fips.install'
-      - '.github/workflows/aur-publish-git.yml'
 
 jobs:
   aur-publish-fips-git:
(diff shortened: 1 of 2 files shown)


Origami74:
v0.3.0 resulted in:
https://aur.archlinux.org/packages/fips

jmcorgan:
Thanks for the fix and the v0.3.0 end-to-end validation.

Landed this as a single squash commit on `maint` (9bf9701) with you preserved as author, then forward-merged into master (aa8f276) and next (195b3ba). All three branches pushed to origin and ngit; CI is in progress.

Targeted `maint` rather than master because maint owns the v0.3.x patch line as of v0.3.0, so the fix is in place before the next maint-cut tag triggers AUR publish.

PR #87, opened 2026-05-14
nostr: ignore stale traversal for active peers

mmalmi:
## Summary
- ignore stale established NAT traversal handoffs once the peer is already connected or already handshaking
- skip stale traversal failures for active/connecting peers before recording cooldown state or scheduling fallback retries
- add regression coverage for connected peers, pending handshakes, and adopted traversal handoffs

## Tests
- cargo test test_try_peer_addresses_skips -- --nocapture
- cargo test test_adopted_traversal_skips_already_connected_peer -- --nocapture
- cargo test test_nostr_traversal_failure_skips_connected_peer -- --nocapture
- cargo test node::tests::unit -- --nocapture
- cargo test node::tests::bootstrap -- --nocapture
- cargo fmt --check
- git diff --check
- cargo test --lib

### DIFF
--- a/src/discovery/nostr/runtime.rs
+++ b/src/discovery/nostr/runtime.rs
@@ -1537,4 +1537,10 @@ impl NostrDiscovery {
         let mut cache = self.advert_cache.write().await;
         cache.insert(npub, advert);
     }
+
+    /// Queue a bootstrap event directly for lifecycle tests without live relays
+    /// or a running traversal task.
+    pub(crate) fn push_event_for_test(&self, event: BootstrapEvent) {
+        let _ = self.event_tx.send(event);
+    }
 }
--- a/src/node/lifecycle.rs
+++ b/src/node/lifecycle.rs
@@ -122,12 +122,7 @@ impl Node {
         }
 
         // Check if connection already in progress to this peer
-        let already_connecting = self.connections.values().any(|conn| {
-            conn.expected_identity()
-                .map(|id| id.node_addr() == &peer_node_addr)
-                .unwrap_or(false)
-        });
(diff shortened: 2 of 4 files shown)


jmcorgan:
Hi Martti,

Nice fix. The four-test layout made it really easy to follow what each piece is protecting against.

I'd like to land this on the `maint` branch instead of `master` so it ships with the next 0.3.x patch release. It's a real bug fix (stale traversal events were unfairly knocking healthy peers around in the cooldown state), and as a bonus it closes a small window where someone could lob fake failure events to push an active peer's cooldown around. Both effects apply equally to 0.3.x. From maint I'll carry it forward through master to next afterward.

I've put the work on a `fix/nostr-skip-stale-traversal` branch with two small follow-ups on top of your commit:

- A `CHANGELOG.md` entry under `[Unreleased] / Fixed` so operators can tell what changed if they notice some debug-log noise disappearing.
- A mirror test for the `Established` arm to match your `test_nostr_traversal_failure_skips_connected_peer`. Same shape, same `push_event_for_test` helper, just exercising the other event path. Pairs nicely with the inner `adopt_established_traversal` guard you already had covered.

CI's running now ([here](https://github.com/jmcorgan/fips/actions/runs/25922967251)). If it's green I'll squash the three commits down to one (with your authorship on it) and merge to maint.

Thanks again for the patch.

jmcorgan:
Landed on `maint` as [`87d1af0`](https://github.com/jmcorgan/fips/commit/87d1af0) with your authorship preserved on the squashed commit. CI was green on the pre-squash branch; the squash didn't touch tree content, just collapsed history.

PR #88, opened 2026-05-14
testing: add boringtun throughput benchmark

mmalmi:
## Summary
- add a two-container Cloudflare BoringTun userspace WireGuard throughput benchmark harness
- generate local WireGuard keys through the benchmark image so host wireguard-tools are not required
- add `testing/static/scripts/iperf-compare-refs.sh` to build two git refs into separate `fips-test:*` images and run the same static iperf topology against both
- allow the static iperf harness to tune `DURATION`, `PARALLEL`, `SETTLE_SECONDS`, `IPERF_TIMEOUT`, and `RUNS`, and bound per-path iperf hangs
- print per-run and aggregate avg/min/max static iperf summaries when `RUNS` is greater than one
- allow `testing/static/docker-compose.yml` to select the image under test with `FIPS_TEST_IMAGE`
- make `testing/scripts/build.sh` respect `CARGO_TARGET_DIR` when copying cross-compiled binaries

## Validation
- `bash -n testing/scripts/build.sh testing/static/scripts/iperf-compare-refs.sh testing/static/scripts/iperf-test.sh testing/boringtun/entrypoint.sh testing/boringtun/scripts/bench.sh testing/boringtun/scripts/generate-keys.sh`
- `bash -n testing/static/scripts/iperf-compare-refs.sh`
- `git diff --check`
- `docker compose -f testing/boringtun/docker-compose.yml config --quiet`
- `FIPS_TEST_IMAGE=fips-test:compare-test docker compose -f testing/static/docker-compose.yml --profile mesh config --quiet`
- `./testing/scripts/build.sh --no-docker` with the local global `CARGO_TARGET_DIR` set
- `testing/boringtun/scripts/generate-keys.sh && docker compose -f testing/boringtun/docker-compose.yml up -d --force-recreate && docker exec bt-bob ping -c 3 10.99.0.1 && testing/boringtun/scripts/bench.sh`
- `RUNS=3 DURATION=5 PARALLEL=8 SETTLE_SECONDS=5 IPERF_TIMEOUT=45 testing/static/scripts/iperf-compare-refs.sh d9ab58a285d1cd99efc366d239948edb3f1c4797 codex/zero-copy-rx mesh` on `ubuntu-dev`

## Benchmark result
On this Docker Desktop aarch64 host, the clean rerun produced:

```text
boringtun bob -> alice : 1000.13 Mbits/sec
```

The static FIPS ref-compare harness was also used on `ubuntu-dev` for PR #90 with `RUNS=3`, which exercised the repeated-run reporting path and produced aggregate avg/min/max rows for each path/ref.

The benchmark containers were shut down with `docker compose -f testing/boringtun/docker-compose.yml down`. A deliberately tiny `DURATION=1 PARALLEL=1` static ref-compare smoke built both images and launched the mesh, but the existing static FIPS topology is not stable at that duration, so it is not reported as a performance number.

### DIFF
--- a/testing/boringtun/.gitignore
+++ b/testing/boringtun/.gitignore
@@ -1,0 +1,1 @@
+generated/
--- a/testing/boringtun/Dockerfile
+++ b/testing/boringtun/Dockerfile
@@ -1,0 +1,28 @@
+# Minimal boringtun-cli image for throughput benchmarking against FIPS.
+# Userspace WireGuard (boringtun) + wireguard-tools + iperf3 in one
+# image. Run two containers on a docker bridge, configure a WG peer
+# between them, then iperf3 client→server across the tunnel.
+
+FROM rust:1-bookworm AS build
+# 0.6.0 (crates.io HEAD) is from 2022 and turns in ~1.7 Mbps in
+# this harness — clearly broken. Pull the boringtun-cli binary from
+# the cloudflare HEAD instead; it gets the post-0.6 perf fixes and
+# the multi-threaded recv loop.
+RUN git clone --depth 1 https://github.com/cloudflare/boringtun /usr/src/boringtun \
+    && cargo install --path /usr/src/boringtun/boringtun-cli
+
+FROM debian:bookworm-slim
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        iperf3 \
+        iproute2 \
+        wireguard-tools \
+        iputils-ping \
+        netcat-openbsd \
+        ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
+COPY --from=build /usr/local/cargo/bin/boringtun-cli /usr/local/bin/boringtun-cli
+COPY entrypoint.sh /entrypoint.sh
+RUN chmod +x /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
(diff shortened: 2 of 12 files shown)


jmcorgan:
Hi Martti, thanks for this. Really nice piece of test tooling, and the comparison script is going to be useful well beyond this PR. Planning to land it as-is. Two small suggestions as light follow-ups, no need to block the merge:

1. Pin the cloudflare/boringtun clone to a specific commit in the Dockerfile. Right now the build does `git clone --depth 1 ...` and `cargo install`, so every rebuild gets whatever HEAD happens to be at the time. That makes the "1000 Mbits/sec" baseline drift as the upstream repo gets new commits, even though nothing in this PR changed. Adding a `git -C /usr/src/boringtun checkout <sha>` line after the clone would keep the number reproducible. Whatever commit was current when you ran the benchmark is the natural choice.

2. A short note in `testing/boringtun/README.md` about what the harness is measuring would help anyone citing the number later. The BoringTun side here is a single Docker bridge with two containers and a direct WG tunnel between them. The FIPS side of the comparison runs the full mesh routing stack. Both are useful numbers but they aren't quite measuring the same thing, and it's easy to read them as if they are.

Optional third nice-to-have if you're touching the file anyway: a line in `testing/README.md` pointing at the new `boringtun/` directory, so it shows up in the top-level index.

Thanks again, this is going to be a useful piece of infrastructure to have around.

jmcorgan:
Landed this on master as a single squashed commit preserving you as author:

```
b05c80e testing: add boringtun throughput benchmark and iperf ref-compare harness
```

PR #89, opened 2026-05-14
udp: batch macOS receive with recvmsg_x

mmalmi:
## Summary

- Add a macOS `recvmsg_x(2)` UDP receive batch path matching the existing Linux `recvmmsg(2)` contract.
- Route macOS through the existing `udp_receive_loop` batch drain so one readiness wake can reap up to 32 datagrams.
- Add a burst-receive regression test and an ignored loopback amortization benchmark.
- Gate the Linux-only `IpAddr` import in `control::listening` so macOS test builds are warning-clean.

## Benchmark

aarch64-apple-darwin, 100-byte UDP loopback receive throughput over 3 s windows:

```text
senders=1: baseline=348308 pps  batched=382982 pps  speedup=1.10×
senders=2: baseline=322229 pps  batched=521407 pps  speedup=1.62×
senders=4: baseline=333902 pps  batched=525601 pps  speedup=1.57×
senders=8: baseline=347476 pps  batched=445765 pps  speedup=1.28×
```

The win appears once the kernel receive queue is deep enough for `recvmsg_x` to amortize wakeup + syscall cost across a batch.

## Validation

`cargo fmt --check`, `git diff --check`, full lib suite, transport::udp suite (25 passed, 1 ignored), and the burst-recv regression test all clean. The ignored amortization bench above is reproducible via `cargo test --release --lib transport::udp::socket::tests::bench_udp_recv_amortization -- --ignored --nocapture`.

### DIFF
--- a/src/control/listening.rs
+++ b/src/control/listening.rs
@@ -15,7 +15,9 @@
 //! See `docs/design/fips-security.md` for the operator-side narrative
 //! that motivates the panel.
 
-use std::net::{IpAddr, Ipv6Addr};
+#[cfg(target_os = "linux")]
+use std::net::IpAddr;
+use std::net::Ipv6Addr;
 
 /// Transport protocol of a listening socket.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
--- a/src/transport/udp/mod.rs
+++ b/src/transport/udp/mod.rs
@@ -413,10 +413,10 @@ impl Drop for UdpTransport {
 
 /// UDP receive loop - runs as a spawned task.
 ///
-/// On Linux, drains the kernel UDP queue in 32-packet bursts via `recvmmsg`
-/// to amortise the per-syscall + per-task-wakeup overhead. macOS / Windows
-/// fall through to single-packet `recv_from`. Either way every datagram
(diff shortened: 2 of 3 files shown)


jmcorgan:
Landed on `master` as [`59225cc`](https://github.com/jmcorgan/fips/commit/59225ccfe1a056e5680ab4450c8558e26fbfa562). Authorship preserved.

The four commits were rebased onto the post-#87/#88 master tip (`b05c80e`) and collapsed into one via soft-reset. Pre- and post-squash trees are byte-identical across `src/control/listening.rs`, `src/transport/udp/mod.rs`, and `src/transport/udp/socket.rs`, so nothing was rewritten in content. The two fallout commits (`control: gate linux-only IpAddr import` and `udp: fix recv benchmark clippy lint`) folded cleanly since both were mechanical consequences of the cfg widening in the substance commits.

CI ran on the squashed commit before merge: 42/42 green, including `Unit tests (macOS)` exercising `test_burst_recv_batch` on the `macos-latest` Apple Silicon runner through the new `recvmsg_x` path.

Closing manually since the squash rewrote the head SHAs and GitHub's patch-equivalence auto-close didn't fire. Thanks!

PR #90, opened 2026-05-14
rx: avoid copies in receive hot paths

mmalmi:
## Summary

- Add a borrowed `SessionDatagramRef` decoder and use it in the forwarding handler so local delivery and coordinate-cache warming don't allocate/copy the session payload.
- Keep forwarding behavior unchanged by materializing an owned `SessionDatagram` only when the packet must be re-encoded for the next hop.
- On Linux UDP receive, move the filled `recvmmsg` slot buffer into `ReceivedPacket` instead of `to_vec()`-ing it; format socket addresses directly into `TransportAddr` without an intermediate `String`.
- Keep the Linux-only listening-socket import scoped so macOS builds stay warning-free.

## Performance

Targeted decode benchmark, release profile:

```
SessionDatagram decode: ref=1.6 ns/op  owned=34.7 ns/op  speedup=21.44×  iters=300000  payload=1200 B
```

End-to-end static-mesh iperf3 comparison (`testing/static/scripts/iperf-compare-refs.sh`, 3 × 5 s × 8 streams on ubuntu-dev):

| Path | origin/master avg/min/max | this PR avg/min/max |
|---|---|---|
| A→D direct  | 1410 / 1390 / 1430 | 1353 / 1280 / 1410 |
| A→E direct  | 1380 / 1330 / 1410 | 1377 / 1350 / 1400 |
| A→B multi-hop |  873 /  820 /  901 |  860 /  855 /  866 |
| A→C multi-hop | 1167 / 1130 / 1220 | 1180 / 1170 / 1190 |
| E→A direct  | 1377 / 1360 / 1390 | 1387 / 1380 / 1390 |

(Mbit/s.) End-to-end change is effectively neutral — the PR removes known per-packet copies on the receive hot path; the focused decode bench shows the 21× win on the specific operation, but the iperf path is dominated by other costs at this throughput.

## Validation

`cargo fmt --check`, `cargo test --lib` (1140 passed / 2 ignored / 0 failed), plus the targeted decode bench above.

### DIFF
--- a/src/control/listening.rs
+++ b/src/control/listening.rs
@@ -15,7 +15,9 @@
 //! See `docs/design/fips-security.md` for the operator-side narrative
 //! that motivates the panel.
 
-use std::net::{IpAddr, Ipv6Addr};
+#[cfg(target_os = "linux")]
+use std::net::IpAddr;
+use std::net::Ipv6Addr;
 
 /// Transport protocol of a listening socket.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
--- a/src/node/handlers/forwarding.rs
+++ b/src/node/handlers/forwarding.rs
@@ -12,7 +12,8 @@ use crate::node::session_wire::{
 };
 use crate::node::{Node, NodeError};
 use crate::protocol::{
-    CoordsRequired, MtuExceeded, PathBroken, SessionAck, SessionDatagram, SessionSetup,
+    CoordsRequired, MtuExceeded, PathBroken, SessionAck, SessionDatagram, SessionDatagramRef,
+    SessionSetup,
 };
 use std::time::{Duration, Instant};
 use tracing::{debug, warn};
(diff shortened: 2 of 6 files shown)


jmcorgan:
**Landed on master as [`b1af151`](https://github.com/jmcorgan/fips/commit/b1af151aefc17c47b18c2921c1b578ac3448cbe7).**

Rebased `b097b0c` onto post-PR-#89 master (`59225cc`), then soft-reset-squashed to a single commit preserving your authorship (`Martti Malmi <sirius@iki.fi>`). The rebase was effectively clean: the `control/listening.rs` `IpAddr` cfg-gate hunk was a textual no-op since PR #89 had already landed the same 3-line change, and the `transport/udp/mod.rs` body lines you modified sit unchanged inside the `recvmmsg` block PR #89 widened to cover macOS.

Two small maintainer-side adjustments folded into the squashed commit:

1. CHANGELOG entry under `[Unreleased]` `### Changed` describing the rx hot-path copy avoidance, including the focused-decode `21.4x` number from your bench.
2. Refreshed the slot-buffer ownership comment in `udp/mod.rs` from `"One Vec per recvmmsg slot ..."` to `"One Vec per recvmmsg / recvmsg_x slot ..."` since after rebase this block now runs both syscalls (Linux + macOS). Net effect of the rebase landing on top of PR #89: your zero-copy slot-move path now applies to the macOS receive path as well as Linux. The decoder/borrowed-view side was already platform-independent and benefits the forwarding handler on every host.

The squashed commit message body covers the four call-site changes individually and the rebase-induced cfg-scope widening.

PR #91, opened 2026-05-15
Data-plane perf overhaul (+~2× single-stream TCP)

mmalmi:
# Data-plane perf overhaul: off-task encrypt + decrypt, GSO, connected UDP

## TL;DR

Moves both AEAD (Authenticated Encryption with Associated Data — ChaCha20-Poly1305 in FIPS, one round per layer per packet) layers plus the `sendmsg` syscall off the rx_loop task onto a per-shard worker pool, adds per-peer `connect(2)`-ed UDP with `SO_REUSEPORT`, and uses Linux UDP GSO (Generic Segmentation Offload — the kernel splits one large super-skb into N on-the-wire datagrams in a single trip through the TX stack) when packets in a batch are uniform-size. GSO is the same kernel primitive WireGuard's in-kernel module and Cloudflare's BoringTun userspace tunnel use to hit 2.5–3.2 Gbps single-stream.

**Single TCP stream: ~+1.9× across every static-peer path.**

| Path | Before this PR (Mbps) | With this PR (Mbps) | Speedup | RTT Δ |
|---|---|---|---|---|
| A→D | 1379 | **2708** | **1.96×** | +0.12 ms |
| A→E | 1394 | **2663** | **1.91×** | +0.11 ms |
| E→A | 1406 | **2624** | **1.87×** | +0.19 ms |

`testing/static/scripts/bench-multirun.sh`, 5 × 15 s × 1 stream medians on a Linux x86_64 docker-bridge mesh (8 vCPU / 24 GiB, kernel 7.0). All CoV < 3 %, 0 outliers (Δ > 20 % from median), 0 % ICMP loss on both branches. The +100–200 µs RTT is the worker queue handoff added by moving AEAD off rx_loop.

### Topology

5 nodes on one docker-bridge subnet (`testing/static/configs/topologies/mesh.yaml`). Static `peers:`:

```
A: D, E       B: C        C: B, D, E
D: A, C, E    E: A, C, D
```

## Diff shape

**30 files, +6173 / -101.** ~85 % of new code is in six new files that drop in alongside the existing transport stack:

| File | Lines | Role |
|---|---|---|
| `src/node/encrypt_worker.rs` | +2238 | FMP+FSP AEAD-seal + sendmmsg / UDP_GSO worker pool |
| `src/node/decrypt_worker.rs` | +693 | shard-owned receive AEAD-open + replay window |
| `src/transport/udp/peer_drain.rs` | +457 | per-peer recv drain thread |
| `src/transport/udp/connected_peer.rs` | +436 | per-peer `connect(2)`-ed UDP socket |
| `src/perf_profile.rs` | +405 | optional per-stage timing reporter |
| `src/transport/udp/darwin_sockopts.rs` | +197 | macOS UDP tuning |

The rest is integration glue. All `#[cfg(unix)]`-gated; Windows continues on the existing tokio-based send/recv.

## What's in scope

- **Off-task encrypt** (`encrypt_worker`) — `std::thread` + `crossbeam_channel`; hash-by-destination dispatch pins a TCP flow to one worker so wire ordering is preserved; per-worker `sendmmsg(2)` batching up to 32; Linux uses `sendmsg(2)+UDP_SEGMENT` when packets in a group are uniform-size.
- **Off-task decrypt** (`decrypt_worker`) — receive-side mirror. Each shard owns its session's recv cipher + replay window in a thread-local `HashMap` (no shared `RwLock`/`Mutex`). Sessions are handed off at `promote_connection` and re-registered on K-bit flip / rekey cutover.
- **FSP+FMP pipelined send** — both AEAD layers seal in-place in the worker on a single wire-buffer allocation; no intermediate `inner_plaintext` / `fsp_payload` Vecs.
- **Per-peer connected UDP** (Linux + macOS) — `SO_REUSEPORT` so the per-peer connected socket can bind to the same wildcard port the listen socket holds; the worker sends with `msg_name = NULL` and the kernel uses its cached 5-tuple (skips per-packet route + neighbor lookup). Tick-driven activation, idempotent.
- **Receive zero-copy** — `mem::replace` the recvmmsg backing buffer instead of `to_vec()` per packet; `SessionDatagramRef` zero-copy view for local delivery; `TransportAddr::from_socket_addr` collapses two allocs to one.
- **rx_loop ordering** — fallback drain promoted ahead of `packet_rx` in the `select!`; interleaved fallback drain every 32 packets inside the rx burst loop so TCP ACKs don't pile up behind a 256-packet inbound burst.
- **Worker pool sizing** — both default to `num_cpus`, overridable via `FIPS_ENCRYPT_WORKERS=N` / `FIPS_DECRYPT_WORKERS=N`.
- **`FIPS_PERF=1`** — optional per-stage timing reporter. Off by default, zero overhead when disabled.
- **Bench harness** (`testing/static/scripts/bench-multirun.sh`) — N reruns (default 5), median / min / max / **CoV %** / per-run outlier flag, avg ping RTT, ICMP loss %, TCP retransmit total. Pre-bench peer-convergence check + per-path route verification via `stats.bytes_sent` deltas — fails fast if traffic exits via a non-static-peer link.

## Test plan

- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean on Linux (what CI runs)
- [x] `cargo nextest run --all --profile ci` — 1228 passed / 4 skipped / **0 failed**
- [x] Session-specific tests 127 / 127, rekey-specific 9 / 9
- [x] `bash -n` on both modified shell scripts
- [ ] CI to confirm macOS + Windows builds

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,2 +1,24 @@
 [profile.ci]
 junit = { path = "junit.xml" }
+# Synthetic node tests build 250-edge meshes with one-shot UDP
+# handshakes; on shared CI runners the localhost stack still drops the
+# occasional msg1 under burst load even with the per-edge repair loop.
+# Allow a retry rather than failing the whole CI run on a single
+# dropped packet.
+retries = 2
+
+[test-groups]
+node-synthetic = { max-threads = 1 }
+
+# nextest runs each test in a separate process, so in-process Tokio mutexes
+# can't serialize the synthetic localhost UDP node tests on CI. Those tests
+# send one-shot handshakes without production reconnect timers; under runner
+# load even small topologies drop the lone msg1. Group all node tests so
+# they run mutually exclusive — slower CI, reliable assertions.
+[[profile.default.overrides]]
+filter = 'test(node::tests::)'
+test-group = 'node-synthetic'
+
+[[profile.ci.overrides]]
+filter = 'test(node::tests::)'
+test-group = 'node-synthetic'
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -662,6 +662,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "82b8f8f868b36967f9606790d1903570de9ceaf870a7bf9fbbd3016d636a2cb2"
+dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
 name = "crossbeam-deque"
 version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1058,6 +1067,7 @@ dependencies = [
  "bluer",
  "clap",
  "criterion",
+ "crossbeam-channel",
  "dirs",
  "futures",
  "hex",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,5 @@
+[workspace]
+
 [package]
 name = "fips"
 version = "0.4.0-dev"
@@ -15,6 +17,7 @@ sha2 = "0.10"
 hkdf = "0.12"
 ring = "0.17"
 rand = "0.10.1"
+crossbeam-channel = "0.5"
 thiserror = "2.0"
 bech32 = "0.11"
 serde = { version = "1.0", features = ["derive"] }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,6 +15,7 @@ pub mod mmp;
 pub mod node;
 pub mod noise;
 pub mod peer;
+pub mod perf_profile;
 pub mod protocol;
 pub mod transport;
 pub mod tree;
--- a/src/node/decrypt_worker.rs
+++ b/src/node/decrypt_worker.rs
@@ -1,0 +1,693 @@
+//! Off-task FMP + FSP decrypt + delivery worker.
+//!
+//! First incremental step of the data-plane shard restructure (per the
+//! architectural plan): each worker now **owns its session state
+//! directly** in a local `HashMap`, with no `Arc<RwLock<HashMap>>`
+//! cache on the Node side and no `Arc<Mutex<ReplayWindow>>` shared
+//! with the rx_loop. The worker is the sole authority over the replay
+//! window and the recv-side ciphers for every session it owns.
+//!
+//! Dispatch is **deterministic by session key**: rx_loop computes
+//! `worker_idx = hash(cache_key) % N` and routes both
+//! `RegisterSession` control messages and per-packet `Job` messages
+//! through the same hash, so a session always lands on the same shard.
+//!
+//! Three message types travel through the per-worker `crossbeam_channel`:
+//!
+//! - **`RegisterSession`** — sent once on the first successful legacy
+//!   decrypt for a session. Hands the worker an owned snapshot of the
+//!   recv cipher + replay window for both FMP and FSP layers.
+//! - **`Job`** — per-packet bulk decrypt + deliver. The worker looks
+//!   up the session in its local HashMap; if absent (registration
+//!   hasn't arrived yet, or session was unregistered), the packet is
+//!   bounced back to rx_loop via the fallback channel.
+//! - **`UnregisterSession`** — sent on rekey / peer drop so the worker
+//!   releases the owned cipher + replay state.
+//!
+//! Only the **bulk-data** path (FMP DataPacket → FSP EndpointData) is
+//! handled by the worker. Anything else (handshakes, MMP reports,
+//! routing errors, IPv6-shim packets going to TUN) is bounced back to
+//! the rx_loop via a fallback channel so the existing slow paths
+//! continue to work.
+
+// **Unix only at the call sites.** On Windows nothing constructs an
+// `OwnedSessionState` or spawns the pool (see `lifecycle.rs`), so
+// every field + function in here becomes dead. Silence the warnings
+// rather than gate them individually.
+#![cfg_attr(not(unix), allow(dead_code))]
+
+use crate::NodeAddr;
+use crate::transport::{TransportAddr, TransportId};
+use crossbeam_channel::{Receiver, Sender, TrySendError, bounded};
+use ring::aead::{Aad, LessSafeKey, Nonce};
+use std::collections::HashMap;
+use std::sync::Arc;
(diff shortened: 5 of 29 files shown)


jmcorgan:
This PR has gone DIRTY against current master. Three landings arrived after it opened: #89 macOS recvmsg_x at 59225cc, #90 rx zero-copy at b1af151, and the coord cache surgical-invalidation merge wave at f51dde6. A rebase onto current master is needed before this can land.

Once you've rebased, I'll do a detailed pass and follow up. Thanks.

mmalmi:
Rebased onto current master ([2b6d402](https://github.com/mmalmi/fips/commit/2b6d4025d783af1e3a571fae5feacc6178c8b1c3)). Status is now MERGEABLE / CLEAN.

Conflicts resolved:
- `src/protocol/link.rs` — master's #90 already added `SessionDatagramRef` (b1af151); dropped the PR's redundant duplicate definition, kept master's slightly cleaner `SessionDatagram::decode → SessionDatagramRef::decode + into_owned()` flow.
- `src/node/handlers/forwarding.rs` — same situation; kept master's ref-based `handle_session_datagram` (uses `into_owned()`; both branches independently arrived at the same zero-copy ref pattern).
- `src/transport/mod.rs`, `src/transport/udp/mod.rs` — comment-only conflicts; kept master's comment in `TransportAddr::from_socket_addr`, kept the PR's mem::replace explanation in the recv loop.

Verified on Linux x86_64 (matches CI target):
- `cargo fmt --check` clean
- `cargo clippy --all-targets --all-features -- -D warnings` clean
- `cargo nextest run --all --profile ci`: **1247 passed / 6 skipped / 0 failed** (222s)

Also built on Windows — succeeds, same pre-existing Windows-only dead-code warnings as before the rebase (`process_authentic_fmp_plaintext` only referenced from `cfg(unix)` worker code; `ESTABLISHED_HEADER_SIZE` import unused on non-unix). Not regressions from the rebase, but happy to gate them on the next pass if you'd like.

jmcorgan:
Squash-merged locally onto master as `0a5c367e` with author preserved as `Martti Malmi <sirius@iki.fi>`. Thanks for the contribution.

The squash inherits five maintainer follow-up commits that were authored on top of the rebased PR head `2b6d4025` before the squash. Two of them cancel out in the net diff and are not visible in the final commit, so they are recorded here:

**1. Decrypt worker session unregistration on rekey drain and peer removal** (visible in net diff).
The `DecryptWorkerPool::unregister_session` API and the matching `WorkerMsg::UnregisterSession` worker-thread handler shipped in the PR but had no callers. Without them, `Node::decrypt_registered_sessions` and each worker's session `HashMap` grew monotonically per rekey on long-lived peers. A new `Node::unregister_decrypt_worker_session` helper is now called at two sites that already iterate `peers_by_index`: the rekey drain-completion block in `handlers/rekey.rs` (drops the worker entry for the old `our_index` once the drain window has expired and the cache_key is unreachable to any in-flight OLD-K packet), and `remove_active_peer` in `handlers/dispatch.rs` (drops the worker entry for each of the four index slots: current, rekey, pending, previous). `unregister_session` is fire-and-forget for missing entries, so calling unconditionally on all four slots is correct and bounds the cleanup without per-slot accounting. The `#[allow(dead_code)]` on `unregister_session` was removed. Two unit tests in `decrypt_worker.rs::tests` cover the worker-side `UnregisterSession` message handling: one asserts the worker drops the entry from its session `HashMap`, the other asserts the call is idempotent on never-seen cache_keys (the safety invariant the unconditional unregister at all four index slots relies on).

**2. Drop empty `[workspace]` table from `Cargo.toml`** (cancels out in net diff).
fips is not a workspace and has no members; the empty `[workspace]` table that suppressed the cargo "implicit workspace from parent directory" warning in a different checkout layout was removed before the squash.

**3. Drop `NVPN_PIPELINE_TRACE` env var from `perf_profile.rs`** (cancels out in net diff).
The env-var name from the original Nostr-VPN project this module was cherry-picked from was removed from the module doc comment and the `enabled()` probe array. `FIPS_PERF` and `FIPS_PIPELINE_TRACE` cover the same enable path.

**4. Round-trip test for the pipelined FSP+FMP wire layout** (visible in net diff).
The existing `fsp_preseal_runs_before_outer_fmp_seal` test covers the seal-ordering invariant with synthetic headers but does not exercise the wire-layout invariant: that `try_send_session_data_pipelined`'s hand-rolled byte offsets (`fsp_aad_offset`, `fsp_plaintext_offset`) match what the canonical receive-side decoders expect. The new `pipelined_send_wire_layout_roundtrips_canonical_decoders` test mirrors the encoder geometry (no coords, the common established-session path), runs the worker's real seal + send via `flush_direct_batch_sync`, and decodes the resulting wire packet using only canonical decoders: `EncryptedHeader::parse` for the FMP outer header, `noise::open` with the parsed header bytes as AAD, 4-byte link-ts and msg_type byte parsing, `SessionDatagramRef::decode` for the SessionDatagram body, FSP header byte-for-byte round-trip check, and `noise::open` on the FSP inner using the parsed FSP header as AAD. Any divergence between the encoder offsets and the decoders fails at one of the parse / open / decode steps before the inner-plaintext assertion fires.

**5. `CHANGELOG.md` `[Unreleased] ### Changed` entry** (visible in net diff).
Describes the worker-pool threading model, hash-by-destination dispatch, `sendmmsg` + UDP_GSO, per-peer connected UDP, the operator-facing env vars (`FIPS_ENCRYPT_WORKERS`, `FIPS_DECRYPT_WORKERS`, `FIPS_CONNECTED_UDP`, `FIPS_PERF`, `FIPS_PIPELINE_TRACE`, plus the macOS-specific knobs), the `FIPS_DECRYPT_WORKERS=0` fallback escape hatch, the bench numbers from the PR body, and the Windows-continues-as-before note.

Two follow-up items have been filed as separate issues:

- #101 Documentation gap for the new env vars. A `docs/how-to/tune-worker-pools.md` companion to the existing `docs/how-to/tune-udp-buffers.md`, covering worker pool sizing, per-peer connected UDP, the perf profiler, and the macOS-specific tuning knobs. The two docs cover adjacent concerns (UDP socket buffers vs. per-stage worker queues) and operators tuning one will want to find the other.
- #100 macOS per-peer drain `recvmmsg` batching gap. The new `peer_drain.rs` recv path uses `recvmmsg` on Linux but falls back to single-packet `recv` on macOS, so macOS established peers get the kernel 5-tuple caching win from the new per-peer connected socket but not the receive-side batching.

Master CI watch on the new tip in flight. Thanks again.

PR #92, opened 2026-05-16
Improve discovery refresh and retry recovery

mmalmi:
## Summary

- start Nostr discovery relay connect/subscribe/publish work in background so node startup is not blocked by slow relays
- filter unroutable direct overlay adverts and bound retry/discovery work per tick to avoid public discovery spam or local queue overload
- add runtime peer-list refresh with recency-ranked candidates and non-disruptive alternate-path handshakes for active peers
- clear stale endpoint sessions when peer restart epochs change, including recovery rekeys
- clean current clippy warnings on macOS/Linux-gated helper code

## Tests

- cargo test --lib --tests --no-fail-fast
- cargo clippy --all-targets -- -D warnings

### DIFF
--- a/src/config/peer.rs
+++ b/src/config/peer.rs
@@ -45,7 +45,27 @@ pub struct PeerAddress {
     /// are tried first.
     #[serde(default = "default_priority")]
     pub priority: u8,
+
+    /// Wall-clock observation timestamp (Unix ms) for ranking by recency.
+    ///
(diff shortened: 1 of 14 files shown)


mmalmi:
Superseded by smaller PRs with more detailed problem/fix descriptions: #93 for platform warning cleanup, #94 for nonblocking Nostr discovery startup, #95 for public advert endpoint filtering, and #96 for runtime peer path refresh/retry recovery.

PR #94, opened 2026-05-17
nostr: start discovery without blocking node startup

mmalmi:
## What was happening

When Nostr overlay discovery was enabled, `Node::start` performed relay connect/subscribe/publish work inline. A slow, unreachable, or overloaded relay could hold node startup hostage even though the local transports and mesh node were otherwise ready to run. In applications embedding FIPS, that made turning the VPN on feel hung and also delayed local/LAN paths that did not need the relay yet.

## What changed

- Split construction of the Nostr discovery runtime from its relay I/O startup.
- Start relay connection/subscription/publish work in a background task so `Node::start` can finish once local transports are ready.
- Keep the initial advert/subscription behavior; it just no longer sits on the critical startup path.
- Add a regression test with deliberately unreachable local relay URLs to prove startup returns promptly.

## How the fix works

The node now installs the Nostr discovery handle immediately and lets the runtime own its relay warmup asynchronously. Later lifecycle ticks can still refresh adverts and drain discovery events after the background startup has progressed. If relays are slow, the mesh can still establish configured direct/LAN links and the relay path catches up separately.

## Tests

- `cargo fmt --check`
- `cargo test test_node_start_does_not_wait_for_nostr_relay_startup -- --nocapture`

This is split out of the broader discovery/path-refresh PR so the startup behavior can be reviewed independently.

### DIFF
--- a/src/discovery/nostr/runtime.rs
+++ b/src/discovery/nostr/runtime.rs
@@ -11,7 +11,7 @@ use nostr::prelude::{
 };
 use nostr_sdk::{Client, ClientOptions, prelude::RelayPoolNotification};
 use serde::Serialize;
-use tokio::sync::{Mutex, RwLock, Semaphore, broadcast, mpsc, oneshot};
+use tokio::sync::{Mutex, Notify, RwLock, Semaphore, broadcast, mpsc, oneshot};
 use tokio::task::JoinHandle;
 use tracing::{debug, info, trace, warn};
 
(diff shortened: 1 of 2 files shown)


jmcorgan:
Rebased onto current master and squash-merged locally with a CHANGELOG `[Unreleased]` entry folded in. New master tip: `c0ccedb4`. Thanks for the contribution.

PR #95, opened 2026-05-17
nostr: filter unroutable direct advert endpoints

mmalmi:
## What was happening

Overlay adverts are public bootstrap hints. Before this change, a malformed or overly optimistic advert could contain direct endpoints that are not usable by remote peers, such as wildcard, loopback, unspecified, link-local, or private addresses. Consumers would cache and attempt those addresses, which can waste connection attempts and, in mixed LAN/VPN/NAT environments, prefer a misleading one-way private path over the intended `udp:nat` bootstrap path.

## What changed

- Treat direct TCP/UDP advert endpoints as publicly usable only when they parse as concrete socket addresses with routable IPs and nonzero ports.
- Keep `udp:nat` adverts valid, since those are rendezvous instructions rather than direct public socket addresses.
- Apply the same filtering when validating received adverts and before publishing the local advert.
- Reject adverts that have no usable endpoints left after filtering.

## How the fix works

`validate_overlay_advert` now normalizes an advert by retaining only usable endpoint hints. Public consumers never learn private LAN candidates through this path; LAN candidates belong in encrypted/direct mechanisms, not public Nostr adverts. Publishing goes through the same endpoint predicate, so local misconfiguration does not leak unroutable direct endpoints either.

## Tests

- `cargo fmt --check`
- `cargo test validate_overlay_advert -- --nocapture`

This is split out from the broader discovery refresh work because it is a self-contained advert hygiene and privacy fix.

### DIFF
--- a/src/discovery/nostr/runtime.rs
+++ b/src/discovery/nostr/runtime.rs
@@ -57,6 +57,67 @@ fn endpoint_summary(endpoints: &[OverlayEndpointAdvert]) -> String {
         .join(",")
 }
 
+fn is_unroutable_direct_advert_ip(ip: std::net::IpAddr) -> bool {
+    match ip {
+        std::net::IpAddr::V4(v4) => {
+            v4.is_private()
+                || v4.is_loopback()
+                || v4.is_link_local()
+                || v4.is_unspecified()
+                || v4.is_multicast()
+                || v4.is_broadcast()
+                || v4.is_documentation()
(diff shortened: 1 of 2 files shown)


jmcorgan:
Landed on the maint branch as a bug-fix with a CHANGELOG `[Unreleased] ### Fixed` entry folded in. New maint tip: `d4181060`. The change will reach master via the standard forward-merge from maint. Thanks for the contribution.

PR #96, opened 2026-05-17
node: refresh peer paths without dropping active links

mmalmi:
## What was happening

Embedders can learn better peer paths after a node is already running: refreshed static hints, recently observed LAN/overlay addresses, or a newly fetched Nostr advert. The old connection path was too static for that environment. Once a peer was already active, refreshed addresses were mostly ignored until the current link died. If a stale UDP port or NAT tuple was still in the peer config, retries could keep hitting that old tuple before trying fresher overlay data. In practice this can leave a peer looking offline or stuck on an inferior path until a daemon restart clears the in-memory state.

A separate but related failure mode appears after a peer restarts: the link layer can promote a fresh connection while an old end-to-end session still exists. If that stale session is left in place, later traffic can fail decrypt/rekey paths even though the link has recovered.

## What changed

- Add a runtime `update_peers` path that diffs the configured peer list, updates retry state, registers identities, and reports added/updated/removed/unchanged counts.
- Add an optional runtime-only `seen_at_ms` timestamp on `PeerAddress`, ignored by config equality, so callers can rank fresh hints without making config diffs noisy.
- Merge static addresses with freshly fetched overlay-advert candidates and sort by recency so current hints are tried before stale ones.
- For active peers, start alternate-path handshakes without tearing down the current link; the old path remains live until the new handshake authenticates and promotes.
- Bound parallel path racing per peer so a refreshed candidate set cannot create unbounded outbound handshakes.
- Refresh stale same-path discoveries when an active peer has gone idle, and keep retry processing paced.
- Clear stale end-to-end sessions when a promoted peer restart/recovery changes the link epoch, so session state is rebuilt on the fresh link.

## How the fix works

The important safety property is that path refresh is additive until authentication succeeds. A new candidate creates a normal Noise handshake on a separate link. Cross-connection handling still decides which side wins, and promotion performs the actual transport tuple switch. If the candidate fails, the existing peer remains in place and traffic continues over the previous usable path.

For offline peers, retry state is updated with the new peer config and the dialer races the freshest concrete candidates in one bounded pass. `udp:nat` can still start the Nostr traversal path, but it no longer prevents direct/static/LAN alternatives in the same candidate list from being attempted.

## Tests

- `cargo fmt --check`
- `cargo test update_peers -- --nocapture`

This replaces the path-refresh portion of the earlier broad PR; Nostr startup and public-advert filtering are now separate PRs.

### DIFF
--- a/src/config/peer.rs
+++ b/src/config/peer.rs
@@ -45,7 +45,27 @@ pub struct PeerAddress {
     /// are tried first.
     #[serde(default = "default_priority")]
     pub priority: u8,
+
+    /// Wall-clock observation timestamp (Unix ms) for ranking by recency.
+    ///
+    /// `None` means "no freshness signal", typically an operator-edited
+    /// static config. The dialer sorts candidates by this field descending
+    /// so freshly observed overlay or runtime hints can be tried before stale
+    /// static addresses. This field is runtime-only and is ignored when
+    /// comparing peer-address lists for config changes.
+    #[serde(default, skip_serializing_if = "Option::is_none", skip_deserializing)]
+    pub seen_at_ms: Option<u64>,
 }
+
+impl PartialEq for PeerAddress {
+    fn eq(&self, other: &Self) -> bool {
+        self.transport == other.transport
+            && self.addr == other.addr
+            && self.priority == other.priority
+    }
+}
+
+impl Eq for PeerAddress {}
 
 fn default_priority() -> u8 {
     100
@@ -62,6 +82,7 @@ impl PeerAddress {
             transport: transport.into(),
             addr: addr.into(),
             priority: default_priority(),
+            seen_at_ms: None,
         }
     }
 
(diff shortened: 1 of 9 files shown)


jmcorgan:
Thanks for this, @mmalmi. This is landed.

I folded it onto both master and next as a single squashed commit that preserves your authorship (`da0d9d39` on master, carried into next), with a CHANGELOG entry covering the new surface and behavior: `Node::update_peers` / `UpdatePeersOutcome`, `PeerAddress::seen_at_ms`, the recency-based address-candidate sort, the per-tick connection budgets, and clearing the stale FSP session on peer restart.

One note on next: because next uses the Noise XX handshake (identity arrives in msg3 rather than msg1), the peer-restart epoch handling had to be re-authored against the XX rekey path. `complete_rekey_msg2` there now also returns the remote startup epoch, and the stale session is cleared once the rekey completes after msg3 is sent. The behavior matches the master-side intent.

Local CI and GitHub CI are green on both branches. Closing as landed. Thanks again.

PR #97, opened 2026-05-17
discovery: add scoped mDNS LAN discovery

mmalmi:
## What was happening

Same-LAN peers currently have to discover each other through the slower overlay path: Nostr advert lookup, STUN/NAT traversal, then a UDP punch before the normal Noise/FIPS handshake can run. That works for public or NAT-separated peers, but it adds seconds of delay for the common case where two nodes are already on the same broadcast domain and directly reachable.

Upstream master did not have mDNS/DNS-SD wired yet: the UDP transport still comments that multicast/DNS-SD discovery is not implemented. This PR adds that local-link discovery path without changing the Nostr path.

## What changed

- Added a LAN discovery module that advertises and browses `_fips._udp.local.` with `mdns-sd`.
- Each advert carries only routing hints: `npub`, protocol version, and optional LAN scope. The observed address/port is used to initiate the existing Noise XX handshake.
- Added `node.discovery.lan` config with `enabled`, `service_type`, and optional `scope`. It defaults on, and the node falls back to the existing Nostr `app` value as a scope when no LAN-only scope is configured.
- Started the mDNS responder/browser from `Node::start`, polled it from the normal RX tick, and shut it down in `Node::stop`.
- Added UDP socket-family selection so an mDNS IPv4 address uses an IPv4 UDP transport and IPv6 uses IPv6, instead of blindly taking the first UDP transport.

## Security / privacy model

The mDNS packet is unauthenticated and is treated only as a hint. A spoofed advert can make us attempt a handshake, but it cannot create a peer unless the remote side proves the advertised npub in the existing Noise exchange.

This is link-local mDNS/DNS-SD. It is not published to Nostr relays. The LAN advert is scoped by service type and optional `scope` TXT so unrelated FIPS meshes sharing a physical LAN do not cross-feed peers.

## Tests

- `cargo fmt --check`
- `cargo check`
- `cargo test discovery::lan::tests -- --nocapture`

The matched-scope mDNS roundtrip test is present but ignored by default because same-process mDNS multicast loopback behavior differs across OSes. The default test run covers TXT parsing, scoped IPv4/IPv6 address handling, service-type isolation, and scope filtering.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1063,6 +1063,7 @@ dependencies = [
  "hex",
  "hkdf",
  "libc",
+ "mdns-sd",
  "nostr",
  "nostr-sdk",
  "portable-atomic",
@@ -1095,6 +1096,17 @@ name = "fixedbitset"
 version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0ce7134b9999ecaf8bcd65542e436736ef32ddca1b3e06094cb6ec5755203b80"
+
+[[package]]
+name = "flume"
+version = "0.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "da0e4dd2a88388a1f4ccc7c9ce104604dab68d9f408dc34cd45823d5a9069095"
+dependencies = [
+ "futures-core",
+ "futures-sink",
+ "spin",
+]
 
 [[package]]
 name = "fnv"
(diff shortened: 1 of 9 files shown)


jmcorgan:
Thanks, @mmalmi. This is landed on both master and next as a squashed commit preserving your authorship (`7d7b551c` on master, carried into next), with a CHANGELOG entry.

I made one deliberate change while folding it in: the feature ships default-off, enabled via `node.discovery.lan.enabled`, rather than on by default. The opt-in path you built is unchanged; only the default flips. The reasoning is least-surprise on upgrade: existing nodes should not begin emitting and listening for multicast on their LANs automatically when they update, and on shared or untrusted networks an operator may not want mDNS advertising the node without opting in. I know you made the case for on-by-default, and the discovery itself is solid; this is purely about the safe default for an in-place upgrade. Anyone who wants it flips one config key.

I also folded in a fix to the startup advertised-port selection: it now skips bootstrap transports and picks deterministically (stable lowest id) instead of iterating the transport map in arbitrary order, so the advertised port is stable and never a bootstrap transport's port.

Local CI and GitHub CI are green on both branches. Closing as landed. Thanks for the contribution.

PR #108, opened 2026-06-07
Reject stale MMP receiver reports

mmalmi:
## What happened

MMP computes RTT from timestamp echoes in `ReceiverReport`s. A delayed duplicate or out-of-order report could be processed after a newer report and treated as a fresh RTT sample. That could poison SRTT even though the path itself had not actually changed.

There was also a dwell-time edge case: receiver dwell is encoded as `u16`, but old code truncated larger dwell values with `as u16`, so an old report could carry an echo with an invalid dwell value.

## Fix

- Ignore duplicate/regressed `ReceiverReport` counter snapshots before updating RTT, loss, goodput, or ETX state.
- Use checked timestamp arithmetic for RTT samples.
- Suppress the RTT echo when receiver dwell no longer fits in the wire field, while preserving the report counters.

This intentionally does not use an arbitrary RTT ceiling; stale reports are rejected by freshness/encoding invariants instead.

## Tests

- Duplicate receiver report does not move SRTT.
- Out-of-order receiver report does not move SRTT.
- Wrapped timestamp sample is ignored.
- Dwell overflow suppresses timestamp echo.

Ran:

```sh
CARGO_TARGET_DIR=/tmp/fips-target-upstream-pr cargo test mmp::
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- MMP sender metrics now ignore duplicate or regressed receiver reports
+  before updating RTT, loss, goodput, or ETX. Receiver reports also
+  suppress timestamp echo when dwell time overflows, so stale reports
+  cannot inflate SRTT.
+
 ### Added
 
 - `pool_inbound` and `pool_outbound` counters on the TCP and Tor
--- a/docs/design/fips-mmp.md
+++ b/docs/design/fips-mmp.md
@@ -100,6 +100,11 @@ inter-frame processing delays inflate spin bit RTT measurements
 unpredictably. Timestamp-echo from ReceiverReports (with dwell-time
 compensation) is the sole SRTT source.
 
+Duplicate or regressed ReceiverReports are ignored before any RTT, loss,
+goodput, or ETX update. If receiver-side dwell time exceeds the wire
+field, the report keeps its counters but sends a zero timestamp echo so
+the sender cannot form an invalid RTT sample.
+
 The spin bit lives in the link-layer FMP inner header, so this
 mechanism applies to link-layer MMP only. Session-layer MMP carries
 its spin bit in the FSP encrypted inner header but uses it the same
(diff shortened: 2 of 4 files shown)


jmcorgan:
Hi Martti,

Nice catch. I've verified the issue and your fix.

I'd like to do this on the `maint` branch rather than `master` so it ships with the next 0.3.x patch release, then carry it forward through master to next afterward. It's a clean bug fix with no wire-format change, so it belongs on the maintenance line.

There are a few small loose ends I'll tidy up while folding it in. None of them block anything, but if you'd rather take any yourself, just say so and I'll leave it for you:

- A `CHANGELOG.md` entry under `[Unreleased] / Fixed` so operators know stale reports now get dropped and the echo is suppressed on dwell overflow.
- A line in `docs/design/fips-mmp.md` (the RTT Measurement section) spelling out that duplicate or regressed reports are ignored before any metric update, and that an out-of-range dwell sends a zero echo.
- A couple of test additions: one with a future-dated echo so it exercises the `checked_sub` side (the wrapped-sample test only hits `checked_add` today), and a check that loss and goodput stay put after a duplicate, next to the SRTT assertion you already have.
- A one-line comment by the guard noting the drop is safe because a report only gets built when the interval has data, so a fresh one always carries advanced counters.

Happy to fold all of these in myself during the merge; otherwise push whatever you'd like to the branch and I'll pick it up from there. Thanks again for the patch.

mmalmi:
Thanks — I rebased the branch onto `maint` and pushed the review follow-ups.

Added:
- CHANGELOG `Fixed` entry
- design note for duplicate/regressed reports and zero echo on dwell overflow
- checked_sub coverage with a future-dated echo sample
- duplicate-report coverage for loss and goodput staying unchanged alongside SRTT
- guard comment explaining why dropping stale reports is safe

Focused checks passed locally:
- `cargo fmt --check`
- `cargo test mmp::metrics -- --nocapture`
- `cargo test mmp::receiver -- --nocapture`
- `git diff --check`

jmcorgan:
Landed on `maint` as d548add with your authorship preserved, plus the CHANGELOG entry, the fips-mmp.md design note, the added duplicate/out-of-order/wrapped/future-dated sample tests and dwell-overflow echo-suppression test, and the guard comment. Thanks for the clean fix and for folding all of that in. It will forward-merge to master and next in the normal flow. Closing as landed.

PR #110, opened 2026-06-11
Add Nym mixnet transport

oleksky:
## Summary

I worked on this during the SEC-07 cohort and believe that it is good to have many transports in FIPS.
Mixnet is a good alternative to the TOR. But we shouldn't stop on this. 

This PR add the mixnet transport only. In a next PR I'll add the sidecar example for testing the transport in docker in the same way as other examples in a repo.

Since the SEC cohort there were few buts in transport and example. They were fixed and tested.

BR,
oleksky

===

Adds a **Nym mixnet transport** — a new transport instance type that tunnels FIPS peer connections through a local `nym-socks5-client` SOCKS5 proxy into the [Nym](https://nym.com/) mixnet, providing Sphinx-routed anonymity and timing obfuscation for peer links.

It is structurally identical to the existing **Tor SOCKS5 transport**: outbound-only, FMP-v0 framing reused from the TCP transport (`tcp::stream::read_fmp_packet`), so an existing TCP-listening peer is reachable as-is over Nym from a node configured with a `transports.nym` block. The `nym-socks5-client` runs separately (sidecar process or container); FIPS only ever sees a SOCKS5 endpoint.

## What's in it

**New files**
- `src/transport/nym/mod.rs` — `NymTransport`: SOCKS5 connection pool, connect-on-send, exponential-backoff readiness probe, FMP framing.
- `src/transport/nym/stats.rs` — `NymStats` atomic counters, serialized into the control-query JSON surface.

**Wiring (additive — no existing behavior changed)**
- `src/transport/mod.rs` — `pub mod nym`, `TransportType::NYM`, `TransportHandle::Nym(..)` variant + dispatch arms. Not platform-gated (works wherever the SOCKS5 client runs).
- `src/config/transport.rs` — `NymConfig` (`socks5_addr`, `connect_timeout_ms`, `mtu`, `startup_timeout_secs`) + `TransportsConfig` merge/empty handling.
- `src/config/mod.rs`, `src/lib.rs` — `pub use NymConfig`.
- `src/node/mod.rs` — instantiate Nym transports at node startup (after the Tor block).
- `src/bin/fipstop/ui/transports.rs` — `"nym"` arm in the transport-detail pane.

## Config

```yaml
transports:
  nym:
    socks5_addr: "127.0.0.1:1080"   # nym-socks5-client SOCKS5 endpoint
    startup_timeout_secs: 120
peers:
  - npub: "<peer-npub>"
    addresses:
      - transport: nym
        addr: "<peer-host>:<tcp-port>"   # peer's TCP endpoint; Nym tunnels TCP
    connect_policy: auto_connect
```

## Testing

- `cargo build --release`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo fmt --check` all green (Linux/amd64 and Linux/arm64).
- Verified end-to-end against a public TCP-listening peer: the node connects through a live Nym service provider, exchanges tree/bloom routing, and carries the data plane (ICMPv6 + sessions) at mixnet latency (~1–2 s RTT vs ~50 ms direct), with the direct route firewalled off to prove traffic crosses the mixnet.

A single-container demo (FIPS + nym-socks5-client + strfry relay) exists on the fork but is intentionally **not** included here to keep this PR focused on the transport; happy to open it separately if useful.

### DIFF
--- a/src/bin/fipstop/ui/transports.rs
+++ b/src/bin/fipstop/ui/transports.rs
@@ -478,6 +478,30 @@ fn draw_transport_detail(frame: &mut Frame, app: &App, area: Rect, t: &serde_jso
                     &helpers::nested_u64(t, "stats", "connect_refused"),
                 ));
             }
+            "nym" => {
+                lines.push(helpers::kv_line(
+                    "MTU Exceeded",
+                    &helpers::nested_u64(t, "stats", "mtu_exceeded"),
+                ));
+                lines.push(helpers::kv_line(
+                    "SOCKS5 Errors",
+                    &helpers::nested_u64(t, "stats", "socks5_errors"),
+                ));
+                lines.push(Line::from(""));
+                lines.push(helpers::section_header("Connections"));
+                lines.push(helpers::kv_line(
+                    "Established",
+                    &helpers::nested_u64(t, "stats", "connections_established"),
+                ));
+                lines.push(helpers::kv_line(
+                    "Timeouts",
+                    &helpers::nested_u64(t, "stats", "connect_timeouts"),
+                ));
+                lines.push(helpers::kv_line(
+                    "Refused",
+                    &helpers::nested_u64(t, "stats", "connect_refused"),
+                ));
+            }
             "ethernet" => {
                 lines.push(Line::from(""));
                 lines.push(helpers::section_header("Beacons"));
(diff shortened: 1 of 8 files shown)


oleksky:
Superset PR with the single-container demo example added on top of this transport: #111. Merge whichever scope you prefer — I'll rebase the other accordingly.

oleksky:
I added a transport + example in PR #111

PR #111, opened 2026-06-11
Nym mixnet transport + single-container FIPS-over-mixnet demo

oleksky:
## Summary

I worked on this during the SEC-07 cohort and believe that it is good to have many transports in FIPS.
Mixnet is a good alternative to the TOR. But we shouldn't stop on this.

This PR adds both mixnet transport and sidecar example for testing the mixnet in docker in the same way as other examples in a repo.

Since the SEC cohort there were few bugs in transport and example. They were fixed and tested.

BR,
oleksky

===

The **full version** of the Nym mixnet work: the transport (same code as #110) **plus** a single-container demo that shows FIPS peering through the mixnet end-to-end.

This PR is a **superset of #110** — if you'd rather take the transport alone, merge #110 and I'll rebase this down to just the example. Two commits:

1. **Add Nym mixnet transport** (src-only, identical to #110) — `NymTransport` over a local `nym-socks5-client` SOCKS5 proxy, structurally mirroring the existing Tor SOCKS5 transport. FMP-v0 framing reused from the TCP transport; outbound-only; not platform-gated. See #110 for the per-file breakdown.

2. **`examples/sidecar-nostr-mixnet-relay/`** — an isolated, reproducible demo. **Everything runs in ONE container**: the FIPS daemon, the mixnet proxy (`nym-socks5-client`), a [strfry](https://github.com/hoytech/strfry) Nostr relay behind nginx, and dnsmasq.

## What the example demonstrates

- A single env var, `FIPS_PEER_TRANSPORT=nym|tcp|udp`, switches the peer link between the **mixnet** and a direct connection — and gates whether the mixnet proxy runs at all (the two stay in sync).
- In `nym` mode the entrypoint starts the SOCKS5 proxy **before** FIPS dials, and iptables **DROPs the direct route to the peer**, so a completed handshake proves the traffic crossed the mixnet.
- The latency delta is the visible proof: ~50–300 ms direct vs ~1–2 s through the Sphinx hops, same peer.

## Notable implementation details

- The image builds **native** for the host: FIPS must not run under emulation — Rosetta/qemu mis-translate the ring/BoringSSL ChaCha20-Poly1305 asm and silently corrupt larger frames (bloom announces), which breaks multi-hop routing while small control traffic still works. Only the prebuilt amd64 `nym-socks5-client` is emulated on arm64 (with its x86-64 loader/glibc copied in), which it tolerates. **This also explains node-test flakiness when building FIPS in amd64 containers on Apple Silicon — worth a docs note upstream regardless of this PR.**
- The example's builder needs `libdbus-1-dev libnftnl-dev libclang-dev clang` (bluer + rustables are unconditional glibc-Linux deps); the current `examples/sidecar-nostr-relay/` Dockerfile predates that and no longer builds — fixable as a separate small PR if useful.

## Testing

- `cargo build` / `clippy --all-targets --all-features -D warnings` / `fmt --check` green on Linux amd64 + arm64.
- Verified from a **fresh clone**: `docker compose up` (only `FIPS_NSEC` set) connects to a public peer through a live Nym service provider, receives bloom routing, and answers multi-hop `ping6 <npub>.fips` at mixnet latency, with the direct route firewalled off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,12 @@ and [testing/README.md](testing/README.md).
   reachable exclusively over the FIPS mesh. The relay container
   shares the FIPS sidecar's network namespace and is isolated from
   the host network.
+- **[examples/sidecar-nostr-mixnet-relay/](examples/sidecar-nostr-mixnet-relay/)** —
+  Single-container demo of FIPS peering through a **mixnet**
+  (implemented with [Nym](https://nym.com/)): the FIPS daemon, the mixnet
+  proxy, and a strfry Nostr relay all in one isolated container, with
+  the direct route to the peer firewalled off so traffic provably
+  crosses the mixnet.
 - **[examples/k8s-sidecar/](examples/k8s-sidecar/)** — Run FIPS as
   a Kubernetes Pod sidecar. The sidecar creates `fips0` in the
   Pod's shared network namespace so every other container in the
--- a/examples/sidecar-nostr-mixnet-relay/.env
+++ b/examples/sidecar-nostr-mixnet-relay/.env
@@ -1,0 +1,37 @@
+# FIPS-over-Nym-mixnet demo configuration.
+# Override these values or create a .env.local file.
+
+# Node identity — generate with: fipsctl keygen
+# Must be set before running: export FIPS_NSEC=<your-nsec>
+FIPS_NSEC=
+
+# Peer configuration (leave FIPS_PEER_NPUB empty for standalone operation).
+# The peer MUST expose a TCP endpoint in nym mode — the Nym SOCKS5 proxy
+# tunnels TCP streams. Find more public peers at https://join.fips.network/
+# Default peer test-us03 exposes tcp:54.183.70.180:443 and udp:…:2121.
+# For udp mode (see FIPS_PEER_TRANSPORT below) change FIPS_PEER_ADDR to
+# 54.183.70.180:2121.
+# The alias doubles as the peer's .fips hostname (<alias>.fips), so it
+# must be a plain hostname label — no dots.
+FIPS_PEER_NPUB=npub136yqae6na688fs75g95ppps3lxe07fvxefj77938zf47uhm6074sxw8ctm
+FIPS_PEER_ADDR=54.183.70.180:443
+FIPS_PEER_ALIAS=test-us03
+
+# Transport — THE switch that selects mixnet vs. direct: nym | tcp | udp
+#   nym : peer traffic goes through the Nym mixnet via the in-container
+#         nym-socks5-client (started automatically, before FIPS). DEFAULT.
+#   tcp : direct TCP to FIPS_PEER_ADDR; the nym client is NOT started.
+#   udp : direct UDP — also set FIPS_PEER_ADDR to the peer's :2121 endpoint.
+# To go back to a direct link, just set this to tcp (or udp) and re-run
+# `docker compose up`. Nothing else needs to change for tcp.
+FIPS_PEER_TRANSPORT=nym
+
+# ----- Nym mixnet (only used when FIPS_PEER_TRANSPORT=nym) -----
+# Network-requester service provider. Leave empty to auto-discover the
+# best-scored provider from https://harbourmaster.nymtech.net/ at startup.
+NYM_SERVICE_PROVIDER=
+NYM_CLIENT_ID=fips-nym-client
+FIPS_NYM_SOCKS5_ADDR=127.0.0.1:1080
+
+# Logging
+RUST_LOG=info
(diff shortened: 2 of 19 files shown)


jmcorgan:
Thanks for this, and for taking it all the way to a working single-container demo instead of just the transport on its own. I read through the whole thing and it looks solid. The Nym transport is basically the existing Tor SOCKS5 transport with the onion and inbound-listener pieces removed, which is exactly how I'd have wanted it done: reuse the outbound path that already works instead of building a parallel one. The dispatch wiring is complete across all the `TransportHandle` methods, the config follows the standard instance pattern, the framing reuses TCP's FMP as-is, and there are no new crate dependencies. CI is green across the board.

Since this was written before the v0.3.0 release, I also went through it convention by convention against where the Tor and TCP transports are today, because a fair amount shifted in the transport layer in between. Most of it held up: the error idioms, the receive hot path, the stats snapshot and how it feeds the control-socket JSON, the constructor signature, and the config all match current mainline. A few things moved underneath the code after you first wrote it, though, and those are what need a pass. None of them are structural. I've listed each with the file and the symbol so they're quick to find.

**Updates needed**

1. Move the stat counters to `portable_atomic`. In `src/transport/nym/stats.rs`, swap `use std::sync::atomic::{AtomicU64, Ordering};` for `use portable_atomic::{AtomicU64, Ordering};`. The other transports (tor, tcp, udp, ethernet, ble) all did this so their 64-bit counters keep compiling on 32-bit targets like the MIPS OpenWrt builds, which have no native 64-bit atomics. CI is green here only because every runner is 64-bit, so it doesn't catch this. The `Ordering::Relaxed` calls don't change, just the import.

2. Drop the per-connection logs from `info!` to `debug!`. A log-hygiene pass (974e146) moved routine per-connection events to `debug!` and kept `info!` for transport start/stop and significant state changes. These five in `src/transport/nym/mod.rs` are still at `info!`: "Connecting via Nym mixnet SOCKS5 proxy", "Initiating background Nym SOCKS5 connect", "Nym SOCKS5 CONNECT starting", "Nym SOCKS5 CONNECT succeeded", and "Nym mixnet connection established via SOCKS5". Tor logs the same events at `debug!` now. The transport "started"/"stopped" and SOCKS5-readiness lines are fine at `info!`.

3. Add a few unit tests for the address parsing. `parse_target_addr` and `validate_host_port` in `src/transport/nym/mod.rs` are pure functions and the easiest place for a small slip to hide, so a `#[cfg(test)]` module covering IP vs hostname, a missing or non-numeric port, and the MTU-exceeded rejection would be good to have; the other transports all came in with unit coverage. One case worth pinning while you're there: `validate_host_port` splits on the last colon, so a bracketed IPv6 address like `[::1]:1080` leaves a stray bracket on the host. The default proxy is loopback so it doesn't bite, and Tor has the same gap, but a test would document it. Tor's `mock_socks5` helper is there to copy if you want to cover the connect path too.

4. Wire up or drop the `connect_refused` counter. `NymStats::record_connect_refused()` exists and fipstop shows a "Refused" line for it (`src/bin/fipstop/ui/transports.rs`), but nothing calls it: the SOCKS5 refusal path calls `record_socks5_error()` and returns `ConnectionRefused`, so "Refused" always reads 0. This one's inherited from Tor, not something you introduced, but a fresh module's a clean place to either call `record_connect_refused()` on that path or remove the counter and its UI line.

5. Small one: the node startup block in `src/node/mod.rs` calls `crate::transport::nym::NymTransport::new(...)` fully qualified inline, while the other transports are `use`-imported at the top and called bare. A `use crate::transport::nym::NymTransport;` lines it up with the rest.

Once these are in I'll squash-merge it. One last thing, unrelated to the merge: the note you included about emulation corrupting the AEAD assembly and silently mangling larger frames is a really useful find. I've documented it on my end and will follow up on it later. Thanks again.

oleksky:
Thanks for your feedback. I will return with the fixes

oleksky:
Done — pushed a commit addressing all five: `portable_atomic` for the counters, the five per-connection logs demoted to `debug!`, address-parse + mock-SOCKS5 connect-path unit tests, removed the dead `connect_refused` counter and its fipstop line, and the bare `NymTransport` import. Ready for re-review — thanks!

jmcorgan:
Thanks for the updates.

It went in as a single squash commit with your authorship preserved (`4e43cb8`), and I added a CHANGELOG entry for the new transport at fold-in time. It is also forward-merged onto the `next` branch. Since `next` uses the XX handshake with a different FMP frame size and version, I adapted the end-to-end SOCKS5 test's frame builder so it passes there too, with `master` keeping your original form. Nothing in the transport itself changed.

PR #113, opened 2026-06-15
Add a Nix flake for reproducible from-source builds

dskvr:
## Summary

Adds a Nix flake at the project root so FIPS can be built and run on Nix /
NixOS systems straight from source, joining the existing packaging formats
(deb, AUR, systemd tarball, OpenWrt, macOS, Windows).

The flake builds all four binaries — `fips`, `fipsctl`, `fips-gateway`,
`fipstop` — and pins the **exact** toolchain declared in
`rust-toolchain.toml` (Rust 1.94.1 + rustfmt/clippy, edition 2024) via
[fenix](https://github.com/nix-community/fenix), so Nix builds match CI and
the AUR/Debian packaging rather than drifting to whatever `rustc` nixpkgs
happens to ship.

## What's included

- **`flake.nix` / `flake.lock`** — pinned inputs (nixpkgs, flake-utils, fenix).
- **Native deps wired up** to match the source tree's build requirements:
  - `pkg-config` + `bindgenHook` (libclang) for the `rustables` / `libdbus-sys`
    bindgen step.
  - `dbus` for `bluer`'s BLE support.
  - `autoPatchelfHook` to rewrite binary RPATHs so the daemon resolves
    `libdbus-1.so.3` and `libgcc_s.so.1` from the Nix store at runtime.
- **Outputs:**
  - `packages.{default,fips}`
  - `apps.{fips,fipsctl,fips-gateway,fipstop}`
  - `checks.fips`
  - `devShells.default` (pinned toolchain + `cargo-edit`)
- **Docs** — a Nix/NixOS section added to `packaging/README.md` in the same
  per-format style as the other entries.

## Design notes

- **Why `autoPatchelfHook`:** without it, `fips` builds fine but fails at
  runtime on NixOS with `libdbus-1.so.3: cannot open shared object file`,
  since NixOS has no global `/usr/lib`. The hook gives a *build-time*
  guarantee that every dynamic dependency resolves.
- **Tests are skipped in the package build** (`doCheck = false`): the suite
  exercises TUN devices, raw sockets, and mDNS that aren't available in the
  build sandbox, mirroring what the AUR and Debian packaging already do.

## Verification

Built and exercised end-to-end in a pure Nix store (`nixos/nix` container,
no global `/usr/lib` — i.e. NixOS-equivalent runtime linking), against the
committed `flake.lock`:

- `nix build .#fips` — all four binaries compile and link.
- `auto-patchelf: 0 dependencies could not be satisfied`.
- All four binaries run and report the correct version (`0.4.0-dev`,
  `x86_64-unknown-linux-gnu`).
- `nix flake check` — **all checks passed**.
- `nix develop` exposes exactly `rustc/cargo 1.94.1` and `clippy 0.1.94`.

## Notes

- The `darwin` platform is declared in `meta.platforms` (the source supports
  macOS), but verification here covered `x86_64-linux` / NixOS specifically.
- Might push another update so the flake can support both stable and dev builds,

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,100 @@
+{
+  "nodes": {
+    "fenix": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "rust-analyzer-src": "rust-analyzer-src"
+      },
+      "locked": {
+        "lastModified": 1781527054,
+        "narHash": "sha256-1fX9ev2Fh5QoKQ41G9dYutjo5j/jywu6tZse5Eb1Ck4=",
+        "owner": "nix-community",
+        "repo": "fenix",
+        "rev": "8c2e51dffefc040a21975da7abf6f252c8c9b783",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
+        "repo": "fenix",
+        "type": "github"
+      }
+    },
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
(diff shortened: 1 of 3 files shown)


jmcorgan:
Landed on master as 3733349, thanks @dskvr. Squashed your flake commit together with a set of maintainer doc fold-ins, keeping you as the author with me as co-author: a CHANGELOG entry, a Nix mention in the README from-source section, a "With Nix (flake)" subsection in docs/getting-started.md, a nix develop note in CONTRIBUTING, and a Nix line in the v0.4.0 release notes (plus the byte-identical root RELEASE-NOTES mirror).

The getting-started and CHANGELOG notes spell out that the flake produces binaries only, with NixOS service integration handled through the system configuration rather than the installer. Ships in v0.4.0.

PR #114, opened 2026-06-17
fix(tun): loop back self-addressed packets for local delivery

Origami74:
## Problem

Reaching the local node at its own mesh address — `ping6 <our-addr>`, `ping6 <our-npub>.fips`, or connecting a local app to a service bound on our own mesh IPv6 — failed with 100% loss on macOS, even though the address is bound and the routing table shows a host route to `lo0`.

macOS point-to-point `utun` interfaces egress self-addressed traffic *down the tunnel* into the daemon's TUN reader rather than looping it back via `lo0`. `handle_tun_packet` saw the `fd::/8` prefix and pushed the packet onto the mesh outbound path, where it was dropped for lack of a session/route to ourselves (`drop_no_route`).

## Fix

When an outbound TUN packet's destination equals our own mesh address, hairpin it straight back to the TUN writer for local (inbound) delivery instead of routing it into the mesh.

The check is **unconditional**, not `#[cfg(macos)]`:
- On Linux the kernel already loops self-traffic via `lo` before it reaches the TUN, so the branch simply never fires there — but keeping it unconditional treats self-delivery as a daemon-level invariant (robust against rp_filter/netns/address-assignment quirks) rather than a bet on kernel routing.
- Unit tests run on Linux only in CI; gating to macOS would ship the fix permanently unverified by automation.

## Testing

- New unit tests: `self_addressed_packet_is_hairpinned_to_tun` and `other_fips_packet_goes_to_mesh`.
- Manually verified on macOS: with a patched daemon, `ping6` and an HTTP fetch to our own mesh address (and `<npub>.fips`) now succeed where they previously timed out.
- `cargo fmt --check`, `cargo clippy`, and `cargo test` clean.

### DIFF
--- a/src/upper/tun.rs
+++ b/src/upper/tun.rs
@@ -704,6 +704,19 @@ fn handle_tun_packet(
 
     // Check if destination is a FIPS address (fd::/8 prefix)
     if packet[24] == crate::identity::FIPS_ADDRESS_PREFIX {
+        // Loopback: a packet to our own mesh address must be delivered
+        // locally, not pushed into the mesh (we have no session/route to
+        // ourselves, so it would just be dropped). Hairpin it back to the TUN
+        // writer for inbound delivery. macOS utun egresses self-traffic into
+        // this reader; Linux loops it via `lo` first, so this never fires there.
+        if packet[24..40] == *our_addr.as_bytes() {
+            trace!(name = %name, "Hairpinning self-addressed packet back to TUN (loopback)");
+            if tun_tx.send(packet.to_vec()).is_err() {
+                return false; // Channel closed, shutdown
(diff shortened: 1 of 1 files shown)


jmcorgan:
Landed on master as 9a9e90a, thanks @Origami74. Folded in two maintainer additions on top of your commit (authorship preserved): a CHANGELOG [Unreleased] entry for the fix, and an expanded source comment making explicit that the hairpin branch is reached only on macOS, whereas Linux routes self-traffic via lo before it reaches the TUN so the branch never fires there.

cargo fmt and clippy are clean and the full lib test suite is green, including your two new unit tests. Ships in v0.4.0.

PR #117, opened 2026-06-18
fix(tun): complete L4 checksum on hairpinned self-traffic (macOS)

Origami74:
## Problem

Connecting to a node's own `<npub>.fips` TCP service from the same macOS host half-opens and hangs: the listener stays in `SYN_RCVD`, the client's `connect()` returns but no data flows and `recv()` times out. `ping6` to the same address works fine.

## Root cause

macOS routes self-addressed traffic as loopback (a `LOCAL` route via `lo0`), which defers the transport TX checksum — but the point-to-point `utun` then egresses the packet into the daemon with only the **pseudo-header partial checksum** present. The hairpin path added in 9a9e90a re-injects these verbatim, so the local stack drops every segment whose checksum MSS clamping doesn't happen to rewrite:

- **SYN / SYN-ACK** get through — MSS clamping rewrites the option and recomputes the checksum as a side effect.
- **Bare ACK, data, FIN** keep the bad checksum and are silently dropped.

Confirmed by `tcpdump -vv` on utun0 — the clamped SYN verifies, the ACK/FIN do not:

```
[S]  mss 1143  cksum 0xb5d3 (correct)     <- hairpinned SYN, accepted
[.]  ack ...   cksum 0x8ece (incorrect)   <- ACK, dropped
[F.] ...       cksum 0x8ece (incorrect)   <- FIN, dropped
```

ICMP is unaffected (the kernel computes its checksum in software), which is why `ping6` always worked.

## Fix

Recompute the TCP/UDP checksum for self-addressed packets on the hairpin path before re-injection (`recompute_transport_checksum`), completing the self-delivery 9a9e90a started (which only covered ICMP and the TCP handshake). Linux is unaffected — it loops self-traffic via `lo` before the TUN, so the hairpin branch never fires.

## Testing

- 4 new unit tests (TCP non-SYN, UDP, non-transport no-op, truncated-packet guard); full `upper::` suite green; `fmt` / `clippy` clean.
- Verified live on macOS: a self-connect that previously timed out after 8s now completes in **~9 ms** with payload delivered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -324,9 +324,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   interface egresses self-addressed traffic into the daemon, which
   previously pushed it onto the mesh outbound path where it was dropped
   for lack of a route to self; such packets are now hairpinned back to
-  the TUN for inbound delivery, so `ping6` and connections to a node's
-  own `<npub>.fips` address work. Linux was unaffected (the kernel
-  already loops self-traffic via `lo`).
+  the TUN for inbound delivery, so `ping6` to a node's own
+  `<npub>.fips` address works. Linux was unaffected (the kernel
+  already loops self-traffic via `lo`). (Full TCP/UDP self-delivery is
+  completed by the checksum fix below.)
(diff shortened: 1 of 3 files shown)


Origami74:
Would like another macos user to test this as well

PR #118, opened 2026-06-20
OpenWrt: SDK-free .apk packaging (OpenWrt 25+) + control-socket fix

Origami74:
## What

Adds OpenWrt `.apk` packaging for **OpenWrt 25+** (where apk-tools is the mandatory package manager) alongside the existing `.ipk`, and fixes the OpenWrt control-socket path so `fipsctl`/`fipstop` work without an explicit `--socket`.

### `.apk` packaging (SDK-free)
- New `packaging/openwrt-apk/` (`build-apk.sh`, `apk-version.sh` + test, README) and a `make apk` target. The existing `.ipk` continues to cover OpenWrt 24.x and earlier.
- Assembled with `apk mkpkg` from apk-tools 3.0.5 built from source — **no OpenWrt SDK image**, mirroring how the `.ipk` is hand-assembled. Reuses the `.ipk` cross-compile and the shared installed-filesystem payload under `openwrt-ipk/files`.
- `apk-version.sh` derives an apk-tools-valid version from *structured* inputs (release tag or commit height) rather than rewriting a flattened version string; covered by a case-table test.
- CI: a single `compile-binaries` job cross-compiles + strips each arch **once**, and both the `.ipk` and `.apk` packagers consume the binaries via `--bin-dir` (no double compilation). The `build-apk` job builds apk-tools, packages, and structurally verifies the `.apk` with `apk adbdump`. apk artifacts publish to Blossom + NIP-94 (`format=apk`) and join the GitHub release with checksums, matching the `.ipk` pipeline.
- Packages are unsigned (same posture as the `.ipk`); install with `apk add --allow-untrusted`.

### Control-socket fix (`fix(openwrt)`)
The OpenWrt init script never created `/run/fips`, so the daemon fell back to binding `/tmp/fips-control.sock`. `fips-gateway` then created `/run/fips` for its own socket, and the clients' directory-existence resolver pointed at a control socket the daemon never bound there — hence the `--socket` workaround. `start_service()` now pre-creates `/run/fips`, the procd equivalent of the systemd unit's `RuntimeDirectory=fips` / `fips.tmpfiles` that every other platform already ships. This is a bugfix to shipped v0.4.0 behavior and could alternatively land via `maint`.

## Test
- **OpenWrt Package CI green** on the branch head: both `.ipk` and `.apk` jobs pass; the `.apk` payload is verified via `apk adbdump`.
- `apk-version.test.sh` case table.
- `build-ipk.sh --bin-dir` exercised locally, producing a structurally valid `.ipk`.

### DIFF
--- a/.github/workflows/package-openwrt.yml
+++ b/.github/workflows/package-openwrt.yml
@@ -677,19 +677,25 @@ jobs:
             echo "  PASS: installed-size >= 1MB"
           fi
 
-          # The adbdump paths: block is hierarchical — directory entries carry
-          # the full relative path, files are basenames nested beneath them.
-          # Reconstruct full paths so we can assert exact payload locations.
+          # The adbdump paths: block is hierarchical. Each directory is a
+          # top-level list item "- name: <full relative dir>"; its files are
+          # "- name: <basename>" nested one indent level deeper under "files:".
+          # (There are no "path:" keys.) Reconstruct full file paths by keying
+          # off the indentation of the directory-level list items.
           RECON=$(awk '
-            /^paths:/ {p=1; next}
-            p && /^[^[:space:]-]/ {p=0}        # a new top-level key ends paths:
+            /^paths:/ {p=1; diri=-1; next}
+            p && /^[^ #-]/ {p=0}                 # a new top-level key ends paths:
             !p {next}
-            /^[[:space:]]*-?[[:space:]]*path:[[:space:]]/ {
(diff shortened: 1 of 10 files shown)


Origami74:
Sanity checked on both:

- comfast CF-WR632AX with openwrt 25
- GL.iNet MT3000 with openwrt 24

Tested gateway as well. All looks good 👍

jmcorgan:
Landed on master as 0f1fd18 (squashed, your authorship preserved). Thanks!

PR #119, opened 2026-06-22
feat(mobile): gate ethernet + system-tun behind default-on features

Origami74:
## What

Introduces two default-on Cargo features so the crate can cross-compile for
mobile (Android/iOS) via `default-features = false`, while leaving desktop
builds completely unchanged.

- **`ethernet`** — gates the raw-socket Ethernet transport (Linux AF_PACKET /
  macOS BPF): the module, the `TransportHandle::Ethernet` variant and all its
  match arms, node construction, and `resolve_ethernet_addr`.
- **`system-tun`** — gates the real TUN platform ops. When off, a no-op
  platform stub stands in and the TUN is app-owned by the embedder (e.g. an
  Android `VpnService`); FIPS performs no system-TUN operations.
- **dns** — `ipi6_ifindex` is `i32` on Android vs `u32` on macOS; added a cast.

`default = [ethernet, system-tun]`, so desktop builds pull both in exactly as
before.

## Why

This is the foundation for the mobile/BLE work. Splitting it out keeps the
plumbing — which is behavior-preserving and mostly mechanical `#[cfg]`
gating — reviewable on its own, and shrinks the follow-up BLE PR to just the
BLE-specific logic.

## Risk / behavior change

None for existing users: both features are on by default, so every desktop
and server build is identical to before. The change is almost entirely
repetitive `#[cfg(unix)]` → `#[cfg(all(unix, feature = "ethernet"))]` edits
across the `TransportHandle` match arms.

## Note for reviewers

The `system-tun`-off path (the no-op TUN stub) has no in-tree consumer yet —
its first user is the app-owned TUN seam that lands in the follow-up `ble-v2`
stack (embedder provides the fd, FIPS talks to it over channels). This PR is
the prerequisite plumbing; the off path is exercised by the mobile embedder.

## Test plan

- [ ] `cargo build` (default features) — desktop unchanged
- [ ] `cargo build --no-default-features` — confirms the mobile gating compiles
- [ ] `cargo clippy` / `cargo fmt --check`

### DIFF
--- a/src/transport/mod.rs
+++ b/src/transport/mod.rs
@@ -11,7 +11,7 @@ pub mod tcp;
 pub mod tor;
 pub mod udp;
 
-#[cfg(unix)]
+#[cfg(all(unix, feature = "ethernet"))]
 pub mod ethernet;
 
 #[cfg(target_os = "linux")]
@@ -19,7 +19,7 @@ pub mod ble;
 
 #[cfg(target_os = "linux")]
 use ble::DefaultBleTransport;
-#[cfg(unix)]
+#[cfg(all(unix, feature = "ethernet"))]
 use ethernet::EthernetTransport;
 #[cfg(test)]
 use loopback::LoopbackTransport;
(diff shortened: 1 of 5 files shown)


src/transport/mod.rs
```diff
@@ -1011,7 +1011,7 @@ impl TransportHandle {
     pub fn state(&self) -> TransportState {
         match self {
             TransportHandle::Udp(t) => t.state(),
-            #[cfg(unix)]
+            #[cfg(all(unix, feature = "ethernet"))]
```
Origami74, line 1014: @jmcorgan possibly better to only feature gate it instead of both unix + ethernet? Thoughts?

PR #121, opened 2026-06-25
packaging(openwrt): default .apk WAN port to DSA name 'wan'

Origami74:
The shared fips.yaml ships ethernet.wan.interface: "eth0", the default WAN port on OpenWrt 24 and earlier. OpenWrt 25 (DSA) boards, which the .apk package targets, name the WAN port "wan" instead. Rewrite the staged copy at build time so the as-installed config binds the Ethernet transport to the right port out of the box, without maintaining a second copy of the config file. The .ipk package keeps "eth0".

Update the openwrt README to document eth0 (24) vs wan (25/DSA) and add a CHANGELOG entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- openwrt: the `.apk` package now defaults `ethernet.wan` to the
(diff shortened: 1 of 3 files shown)


jmcorgan:
Thanks Arjen. Folded into master as 22a5b3e, with the CHANGELOG entry moved up into the [0.4.0] section so it ships in that release. Closing as landed.

PR #123, opened 2026-07-02
feat(openwrt): 802.11s open-mesh backhaul support

Origami74:
Router-to-router radio backhaul over an open 802.11s mesh, with FIPS providing all encryption (Noise IK), authentication, and routing on top of bare L2 neighbor links. Field-tested on real hardware (two OpenWrt routers peering over `fips-mesh0`).

## Design choices

- **Open mesh (no SAE):** FIPS's Noise handshake already authenticates and encrypts every peer link; SAE would duplicate it, add credential provisioning, and force ath10k into raw Tx/Rx mode. Strangers can peer at L2 but die at the handshake — same model as mDNS/BLE discovery.
- **`mesh_fwding 0`:** 802.11s HWMP stays off; FIPS is the routing layer. Each mesh link is a plain neighbor link, and discovery beacons reach direct neighbors only.
- **Radio setup is opt-in, transport binding is default-on:** `fips-mesh-setup <radio>` creates the interface (a package must not commandeer radios on install); the shipped `fips.yaml` carries `mesh0`/`mesh1` Ethernet-transport entries enabled by default — each inert until its interface exists, since a transport whose interface is missing at startup is skipped with a warning.
- **Dual-band:** run the helper once per radio (`radio0` → `fips-mesh0`, `radio1` → `fips-mesh1`); two bands give every router pair complementary paths (2.4 GHz reach, 5 GHz capacity) that the FIPS spanning tree treats as redundant links with automatic failover.

## Contents

- `fips-mesh-setup`: per-radio UCI helper (create/remove, collision guard), prints band/channel, warns loudly on `channel 'auto'` — mesh points only peer on the same channel, and auto-selection is the classic silent-failure cause (hit during hardware testing).
- Shipped `fips.yaml`: default-enabled `mesh0`/`mesh1` transport entries + regression test parsing the shipped config through the real `Config` deserializer.
- Packaging: helper installed in ipk/apk/buildroot (three synced copies); CI structural checks and shellcheck targets extended.
- `docs/how-to/set-up-80211s-mesh-backhaul.md`: full guide including the radio-first-then-restart ordering constraint and an ordered no-peers triage (channel mismatch → on-air scan → DFS CAC → regdomain).

## Testing

`cargo test` green (new regression test included), `cargo fmt` clean, shell scripts `sh -n` checked; mesh formation and channel-mismatch triage verified on two physical routers.

### DIFF
--- a/.github/workflows/package-openwrt.yml
+++ b/.github/workflows/package-openwrt.yml
@@ -285,6 +285,7 @@ jobs:
             "$FILES_DIR/etc/fips/firewall.sh"
             "$FILES_DIR/etc/hotplug.d/net/99-fips"
             "$FILES_DIR/etc/uci-defaults/90-fips-setup"
+            "$FILES_DIR/usr/bin/fips-mesh-setup"
           )
           fail=0
           for f in "${TARGETS[@]}"; do
@@ -404,6 +405,7 @@ jobs:
             ./usr/bin/fipsctl
             ./usr/bin/fipstop
             ./usr/bin/fips-gateway
+            ./usr/bin/fips-mesh-setup
             ./etc/init.d/fips
             ./etc/init.d/fips-gateway
             ./etc/fips/fips.yaml
@@ -717,6 +719,7 @@ jobs:
 
           for path in \
             usr/bin/fips usr/bin/fipsctl usr/bin/fipstop usr/bin/fips-gateway \
+            usr/bin/fips-mesh-setup \
             etc/init.d/fips etc/init.d/fips-gateway \
             etc/fips/fips.yaml etc/fips/firewall.sh etc/dnsmasq.d/fips.conf \
             etc/sysctl.d/fips-gateway.conf etc/sysctl.d/fips-bridge.conf \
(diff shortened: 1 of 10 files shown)


Origami74:
Opened #125 Not blocking for this PR.

Other than that testing worked, between an openwrt 24 and an openwrt 25 router.

jmcorgan:
Two small things, neither blocking:

- The default-on `mesh0`/`mesh1` entries will log a per-boot bind warning on installs that never create those interfaces. Is that the intended flash-and-drop-in default, or would you rather ship them commented out and have `fips-mesh-setup` uncomment them when it creates an interface?

- Both the create and remove paths call `wifi reload`, which briefly drops client APs on all radios. Might be worth a one-line note in the how-to so users expect the blip.

Origami74:
Addressed both points in 4ceeb0e.

**1. Per-boot bind warning** — went with the commented-out default. `mesh0`/`mesh1` now ship commented in `/etc/fips/fips.yaml`, so a stock install that never creates `fips-mesh*` logs nothing. `fips-mesh-setup <radio>` uncomments the matching `meshN` block when it creates the interface and re-comments it on `remove`, so the flash-and-drop-in flow still needs no manual config edit. The shipped-config regression test now checks both states parse (as shipped = inactive; after the uncomment the helper performs = valid `mesh0`/`mesh1`).

**2. `wifi reload` blip** — noted in the how-to (Step 1) that both create and remove run `wifi reload`, which briefly drops client APs on all radios.

Two implementation details worth flagging on the toggle logic: the block rewrite forces `chmod 600` before replacing the file (it is installed 0600 and may hold an inline `nsec`), and the awk uses `[ \t]` rather than `[[:space:]]` for BusyBox awk.

Caveat: **not yet validated on hardware.** The comment/uncomment round-trip and 0600 preservation pass locally (byte-identical round-trip, mode held), and `sh -n` / shellcheck / `cargo test` / `cargo fmt` are green, but the toggle has not been exercised on a real OpenWrt/BusyBox device yet. Will confirm on the two test routers before this merges.

jmcorgan:
Thanks @Origami74. Squashed the seven commits into one, 054d17aa, preserving your authorship, and rebased onto current master.

PR #124, opened 2026-07-06
Enable Android fips and fipsctl builds

dskvr:
## Summary
- make host-only mDNS, raw Ethernet, and TUN setup compile as explicit unsupported Android paths
- avoid Android getifaddrs linker failures by excluding mdns-sd on Android
- fix Android DNS pktinfo type conversion without tripping host Clippy so fips/fipsctl build for x86_64 Android

## Verification
- cargo fmt --check && git diff --check
- cargo clippy --all-targets --all-features -- -D warnings
- cargo test
- cargo build --target x86_64-linux-android --bin fips --bin fipsctl
- cargo build --release --target x86_64-linux-android --bin fips --bin fipsctl
- MeshDrop installed APK smoke used the release-built binaries and reported android-native-fipsctl with rustCore=true

## Known gaps
- ARM64 Android cross-build not run in this branch
- Android raw Ethernet, host TUN management, and LAN discovery remain intentionally unsupported

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,6 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 tokio = { version = "1", features = ["rt", "macros", "signal", "sync", "net", "time", "process", "io-util"] }
 futures = "0.3"
 simple-dns = "0.11.2"
-mdns-sd = "0.19"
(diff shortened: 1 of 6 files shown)


dskvr:
My agent was gungho and jumped the gun on this one.

@Origami74 if you see something worth keeping by all means steal it :smile:

PR #129, opened 2026-07-25
feat(freebsd): FreeBSD support — daemon, TUN datapath, .fips DNS, and native pkg packaging

fr34aky:
## Summary

Ports FIPS to FreeBSD: the daemon, `fipsctl`, and `fipstop` build, run,
and package natively, including mDNS LAN discovery. Validated end-to-end
on real FreeBSD 15 hardware — build, `pkg` install, rc.d start/stop,
system-wide `.fips` DNS through `local_unbound`, and a mesh ping to a
remote npub over the mesh TUN interface. `fips-gateway` remains
Linux-only (its NAT backend is nftables); a pf-based backend is possible
follow-up work.

Two concerns, one commit each — `feat(freebsd)` (the platform arm,
details below) and `packaging(freebsd)` (native `.pkg`, rc.d services,
`.fips` resolver integration, CI workflow) — plus a `fix(freebsd)` /
`packaging(freebsd)` pair addressing the review findings.

No vendored or patched dependencies.

## FreeBSD reuses the macOS arm wherever possible

macOS is the closest relative, so instead of a parallel FreeBSD
implementation the port extends the existing macOS cfg gates to
`cfg(any(target_os = "macos", target_os = "freebsd"))`:

- **One shared `mod platform`** in `tun.rs` (`ifconfig`/`route`
  shell-outs, shared `interface_exists`/`configure_interface`/
  `delete_interface`), with FreeBSD-specific arms only where kernel
  behavior differs.
- **Kernel-assigned device names**, as with `utun` on macOS: the `tun`
  crate can't name devices on FreeBSD, so the device keeps its `tunN`
  name — no rename, no name collisions, no startup cleanup.
- **Identical teardown**: down the interface, signal the shutdown pipe,
  join the reader. Like `utun`, a devfs-cloned tun device is destroyed
  by the kernel when its last fd closes, so no `SIOCIFDESTROY`
  special-casing is needed.
- **Shared 4-byte address-family framing helpers.** The writer builds
  the prefix with `utun_af_inet6_header()` and `TunDevice::read_packet`
  strips inbound frames with `parse_utun_af_prefix()` on FreeBSD, so
  the documented "returns a raw IP packet" contract holds on all
  platforms and neither reader loop knows about TUNSIFHEAD. The value
  derives from `libc::AF_INET6` (30 on Darwin, 28 on FreeBSD); the
  round-trip test module runs on both platforms and pins the value
  per OS.
- **Shared config/run paths** (see cross-platform notes below).

### What FreeBSD deliberately does not get

- **The batched UDP receive path and the connected-UDP fast path stay
  Linux/macOS.** FreeBSD has `recvmmsg(2)`, but it is a libc wrapper
  looping over `recvmsg` — not a kernel batch like Linux's syscall or
  Darwin's `recvmsg_x` — so compiling the batched arm would gain
  nothing over the portable single-packet loop FreeBSD takes. The
  connected-UDP fast path relies on Linux `SO_REUSEPORT` steering
  semantics; enabling it on FreeBSD needs on-target validation first
  and is possible follow-up work. Both facts are noted in
  `packaging/freebsd/README.md`.
- **Ethernet and BLE transports are not compiled** (Ethernet is
  `any(linux, macos)`, BLE needs BlueZ). The README support matrix now
  has a FreeBSD column saying so, and a configured `ethernet:` block on
  a platform without the transport now warns instead of being silently
  ignored.

## FreeBSD kernel behaviors the port had to handle

- **Without `TUNSIFHEAD`, `tunoutput()` rejects every non-IPv4 frame
  with `EAFNOSUPPORT`** — IPv6 over tun requires it. It's enabled at
  create time (the ioctl constant `_IOW('t', 96, int)` isn't in the
  `libc` crate), and every frame then carries the 4-byte network-order
  address-family prefix described above.
- **ND6 `IFDISABLED` must be cleared** or the kernel silently drops all
  IPv6 on the interface.

## LAN discovery (mDNS) works on FreeBSD

`mdns-sd` is bumped 0.19 → 0.20, which picks up `socket-pktinfo` 0.4.1
— the first release that builds on FreeBSD (native `IP_RECVDSTADDR` +
`IP_RECVIF` in place of the Linux-style `IP_PKTINFO`; upstreamed in
pixsper/socket-pktinfo#6 and verified on FreeBSD 15 hardware as part of
this work). No FIPS code changes were needed; `src/mdns` is identical
to master.

On-wire note for the bump (it applies to every platform): between 0.19
and 0.20.3 the crate changed mDNS *timing*, not record formats — RFC
6762 §6 multicast rate limiting (identical responses at most once per
second per interface), PTR response delays narrowed from 20–120 ms to
10–50 ms, initial-query jitter likewise, wider announcement
retransmission spacing, and a fix that reports instead of panicking on
labels over 63 bytes. All are compliance improvements; packet formats
are unchanged.

## Cross-platform changes (all intentional)

- **macOS default paths now match its package.** The macOS `.pkg`
  stages everything under `/usr/local/etc/fips`
  (`packaging/macos/build-pkg.sh`), but the hosts / `peers.allow` /
  `peers.deny` / `fipsctl keygen` defaults were hardcoded to
  `/etc/fips`, leaving `.fips` hosts aliases silently unresolvable on
  package installs — the same bug found and fixed on FreeBSD. Because a
  hand-created ACL file stranded at the old path would fail open, the
  daemon now warns at startup when a file exists at `/etc/fips` but not
  at the new default, `fipsctl keygen` prints the equivalent note, and
  the move has a `### Changed` CHANGELOG entry. The
  `/usr/local/etc/fips` config-search-path entry and the
  `/var/run/fips` control-socket location (neither OS has `/run`) are
  shared cfg arms with FreeBSD.
- `mdns-sd` 0.19 → 0.20 (all platforms; see the on-wire note above).
- Daemon logs disable ANSI color when stdout is not a terminal (all
  platforms) — under daemon(8)/systemd redirection the log files were
  full of escape sequences.
- `ENXIO` is suppressed during shutdown reads on macOS as well as
  FreeBSD (same benign race, same fix).

Everything else is `#[cfg]`-gated; Linux and Windows builds are
otherwise untouched.

## Packaging

`packaging/freebsd/build-pkg.sh` produces `deploy/fips-<version>.pkg`
(`pkg create`, version `-` mapped to `.` per pkg rules). The rc.d script
runs the daemon under daemon(8); `fips_dns` is a oneshot that forwards
the `fips.` zone to the daemon's `[::1]:5354` responder, trying base
`local_unbound`, pkg `unbound`, then `dnsmasq` (readiness is detected
via the responder socket, since the TUN name is kernel-assigned). The
unbound drop-in carries three settings that field debugging showed are
each individually fatal if missing (`do-not-query-localhost: no`,
`do-ip6: yes`, `domain-insecure: "fips."`), and the setup script warns
about the resolver pitfalls hit during validation (no loopback
nameserver in resolv.conf, EDNS-broken CPE forwarders as upstream,
self-forwarding loops). `packaging/freebsd/README.md` documents
install, host-resolver setup, and debugging.

Security and lifecycle, matching the Debian/macOS packages:

- `fips.yaml` installs `0600 root:wheel` (it may hold the node's
  `nsec:`); the sample is staged `0600` too.
- post-install creates a `fips` group, the rc script makes
  `/var/run/fips` `root:fips 0750` (falling back to `0755` for source
  builds without the group), and the daemon already chowns the control
  socket to the group — so `fipsctl`/`fipstop` work without root for
  group members.
- `pkg upgrade` runs the old package's deinstall scripts, so
  pre-deinstall stops both services unconditionally (the daemon binary
  is never replaced under a running process) and confines the `.fips`
  resolver teardown and sample-config cleanup to real removals
  (`PKG_UPGRADE` guard, verified against `pkg-script(5)` on pkg 2.7.5 /
  FreeBSD 15.1); post-install restarts whatever is enabled when
  upgrading.

A `package-freebsd.yml` workflow mirrors the Linux/macOS package
workflows (same version derivation, sha256 handoff verification,
release upload on tags). With no GitHub-hosted FreeBSD runners, it
builds inside a KVM-accelerated FreeBSD 15.1 VM (`vmactions/freebsd-vm`)
on `ubuntu-latest` — rustup honors the `rust-toolchain.toml` pin — runs
`cargo test` in the VM (the only place the FreeBSD cfg arms' unit tests
execute in CI), and smoke-installs the produced package with `pkg add`
(binaries present and fully linked, rc.d scripts in place, sample-config
install-if-absent semantics, `fips.yaml` mode `0600`, and the `fips`
group). The artifact is named `fips-<version>-freebsd-<arch>.pkg` to
stay distinct from the macOS `.pkg` release assets.

## Testing

- All four CI gates green on FreeBSD **and** Linux (`fmt`, `build`,
  `clippy -D warnings`, `test`); `testing/check-log-strings.py` clean.
- Config-path tests extended for the shared macOS/FreeBSD locations.
- The AF-prefix framing is covered by the shared round-trip test module
  (both platforms, per-OS value pins), which now also runs on FreeBSD
  in CI via the package workflow's in-VM `cargo test`.
- The Docker integration harness is Linux-hosted and unaffected by the
  cfg-gated changes.

No associated issue exists for FreeBSD support, so nothing to close.


fr34aky:
Two design notes on the TUN framing, since they're the least obvious
part of the diff:

**Why the `TUNSIFHEAD` ioctl is the one FreeBSD-only step in the create
path.** macOS `utun` and FreeBSD `tun(4)` use the same wire framing —
each frame prefixed with a 4-byte network-order address family — but
they differ in their defaults. `utun` is permanently in that mode; there
is no ioctl to enable or disable it. FreeBSD's `tun(4)` instead defaults
to headerless IPv4-only mode: per tun(4), without `TUNSIFHEAD` every
`write(2)` "is assumed to be of type `AF_INET`", and on the kernel→
daemon side `tunoutput()` rejects every non-IPv4 frame with
`EAFNOSUPPORT` — observed exactly that way during hardware validation
(no IPv6 passes at all). The one-time ioctl at create switches the
device into "multi-af" mode, i.e. the mode utun is always in. After
that, the framing code (writer prepend / reader strip) is shared
between the two platforms.

**Why the AF prefix is built from `libc::AF_INET6` on both platforms.**
The prefix must carry the local kernel's `AF_INET6` value, which
differs: 30 on Darwin, 28 on FreeBSD. The `libc` crate resolves
`AF_INET6` per target, so a single
`(libc::AF_INET6 as u32).to_be_bytes()` replaces what used to be two
cfg-gated lines, with byte-identical output on macOS
(`[0, 0, 0, 30]`). The macOS test module pins this:
`assert_eq!(libc::AF_INET6 as u32, UTUN_AF_INET6)` alongside the
existing wire-format round-trip tests, so the equivalence is enforced
rather than assumed.

jmcorgan:
Merged on `master` as a single commit with your authorship, and carried up to `next`.

The landed diff differs from yours in one area. A separate change had already moved the macOS config, ACL, hosts and keygen paths to `/usr/local/etc/fips` and flowed up from `maint`, touching the same constants. The two were reconciled into one implementation: predicates widened to cover FreeBSD alongside macOS, one copy of the legacy-path warning, and keygen reading the shared `SYSTEM_CONFIG_DIR` rather than a literal. FreeBSD behaviour is as you specified.

The identity key point from the follow-up is fixed, in that other change rather than here. The relocation comes from adding a second system config directory to the search path, so it was live independently of this PR.

Two review notes are filed as follow-up rather than done: the TUN shutdown pipe still uses raw libc calls from the node lifecycle, which this widens to a second platform, and route setup still keys on `route`'s stderr wording, now across two vendors.

The `mdns-sd` bump to 0.20 is the only part of this that changes behaviour on Linux and macOS. It will be reviewed with the other dependency updates rather than as part of the port.

PR #132, opened 2026-08-02
 fix: follow the macOS install layout for config, ACL, and identity paths

sh1ftred:
## Summary

   The macOS packaging (`packaging/macos/`) installs config under
   `/usr/local/etc/fips/` (Homebrew-style prefix, wired through the launchd
   plist and `build-pkg.sh`), but several default-path constants and the config
   search path were hardcoded to `/etc/fips/` for all Unix. On macOS the daemon
   and `fipsctl` looked in a directory that does not exist: the ACL and
   host-map loaders hit their `NotFound` no-op arm and silently returned empty
   state, `fipsctl keygen` wrote identity keys where the daemon never reads
   them, and a bare `fips` run (no `--config`) skipped the installed system
   config and derived identity key paths from the wrong base.

   **Root cause:** every affected path used a bare `#[cfg(unix)]` arm returning
   `/etc/fips` with no macOS case, ignoring that the macOS packaging uses
   `/usr/local/etc/fips/`.

   ## What changed

   Each surface is split into `target_os = "macos"` → `/usr/local/etc/fips` and
   other Unix → `/etc/fips`, with Windows behavior preserved exactly:

   - **Peer ACL** (`src/node/acl.rs`): `DEFAULT_PEERS_ALLOW_PATH` /
     `DEFAULT_PEERS_DENY_PATH`. Pre-fix, a populated `peers.deny` reported
     `effective_mode: "default_open"` and `enforcement_active: false`; denied
     peers connected regardless.
   - **Hosts file** (`src/upper/hosts.rs`): `DEFAULT_HOSTS_PATH`. Host-file
     aliases went unloaded on macOS.
   - **System config search path** (`src/config/mod.rs`): `Config::search_paths()`
     now uses a new shared `SYSTEM_CONFIG_DIR` constant (`/usr/local/etc/fips` on
     macOS, `/etc/fips` otherwise). This also fixes the `resolve_identity`
     fallback, which derives `fips.key` / `fips.pub` from the first search path.
   - **Keygen output dir** (`src/bin/fipsctl.rs`): `default_key_dir()` returns
     `SYSTEM_CONFIG_DIR` on Unix, so `fipsctl keygen` writes keys where the
     daemon actually reads them.

   `SYSTEM_CONFIG_DIR` is shared between the search path and keygen so the two
   cannot drift independently — which is how keygen got out of sync with the
   daemon in the first place. CHANGELOG updated with one entry per surface.

   ## Tests

   - Platform-gated pin tests assert the literal path on each Unix variant
     (`macos` + `all(unix, not(macos))`). The macOS pins run on the
     `macos-latest` CI job; the Linux pins run on the Linux matrix.
   - `test_search_paths_includes_expected` asserts against `SYSTEM_CONFIG_DIR`
     on every platform (wiring, not a third literal copy of the path).
   - The `snapshot_show_acl` snapshot test in `control/queries.rs` was
     inadvertently relying on the bug: `build_test_node` constructs a `Node`
     via `with_identity`, which reads the host's real ACL files. On a macOS dev
     machine with a populated `peers.deny` the snapshot flipped
     `effective_mode` to `denylist`. A new `isolate_peer_acl_for_test` helper
     repoints the reloader at non-existent temp paths so the snapshot is
     deterministic regardless of host state.

   ## Notes

   - **Windows is unchanged.** The ACL constants were previously unconditional
     `/etc/fips` for every platform and remain so on Windows; the hosts
     constant already had a `#[cfg(unix)]` / `#[cfg(windows)]` split that is
     preserved.
   - **The launchd-installed daemon was unaffected** by the search-path bug —
     it always passes `--config /usr/local/etc/fips/fips.yaml` — so only
     bare/manual `fips` runs were skipping the installed config.
   - **No integration suite** specifically exercises macOS path resolution;
     coverage is the platform-gated unit pins plus the `test-macos` CI job
     running the full library suite.
   - Operator docs in `docs/` are largely Linux-first; the reference tables
     (`configuration.md`, `cli-fips.md`, `cli-fipsctl.md`) are updated with the
     macOS paths. A broader docs pass for macOS-specific prose is left as a
     follow-up — it's outside this change's footprint.

PS: There are some sloppy comments again in the code, we can remove them after you review this.

### DIFF
--- a/docs/how-to/persistent-identity.md
+++ b/docs/how-to/persistent-identity.md
@@ -72,6 +72,37 @@ present after the first successful daemon start. If the daemon never
 came up cleanly (config error, permission problem), the key files
 will be missing.
 
+### macOS note
+
+The macOS package (`.pkg`) installs config and keys under
+`/usr/local/etc/fips/` instead of `/etc/fips/`. The paths above become:
+
+| Linux / other Unix | macOS |
+| --- | --- |
+| `/etc/fips/fips.yaml` | `/usr/local/etc/fips/fips.yaml` |
+| `/etc/fips/fips.key` | `/usr/local/etc/fips/fips.key` |
+| `/etc/fips/fips.pub` | `/usr/local/etc/fips/fips.pub` |
+
+`fipsctl keygen` writes to `/usr/local/etc/fips/` by default on macOS.
+The daemon still probes `/etc/fips/fips.yaml` as a fallback (so an
+existing install is not broken by an upgrade), but the macOS packaging
+only installs files under `/usr/local/etc/fips/`.
+
+If you have files in `/etc/fips/` from a manual install, move them:
+
+```sh
+sudo mv /etc/fips/fips.yaml /usr/local/etc/fips/fips.yaml
+sudo mv /etc/fips/fips.key /usr/local/etc/fips/fips.key
+sudo mv /etc/fips/fips.pub /usr/local/etc/fips/fips.pub
+sudo mv /etc/fips/peers.allow /usr/local/etc/fips/peers.allow 2>/dev/null || true
+sudo mv /etc/fips/peers.deny /usr/local/etc/fips/peers.deny 2>/dev/null || true
+sudo mv /etc/fips/hosts /usr/local/etc/fips/hosts 2>/dev/null || true
+```
+
+The daemon logs a warning at startup if any of `fips.yaml`, `peers.allow`,
+`peers.deny`, or `hosts` exist at the old `/etc/fips/` path but not at
+`/usr/local/etc/fips/`.
+
 ### File layout and permissions
 
 | Path | Mode | Owner | Contents |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   keep `/etc/fips`, Windows is unchanged, and the values are pinned by
   platform-gated unit tests.
 
+- macOS: the system-wide config search path is now
+  `/usr/local/etc/fips/fips.yaml`, matching the install layout the macOS
+  packaging ships. It was hardcoded to `/etc/fips/fips.yaml`, so a bare
+  `fips` run without `--config` probed a directory that does not exist and
+  derived identity key paths from it. The search path and `fipsctl keygen`
+  now share one platform constant for the config dir, so the two cannot
+  drift apart again. The launchd-installed daemon was unaffected (it always
+  passes `--config`). Linux and Windows behavior is unchanged.
+
 - Nostr NAT traversal no longer breaks after the host suspends. The traversal
   clock cached a Unix timestamp once at startup and advanced it with a
   monotonic `Instant`, which does not tick while a machine is asleep, so after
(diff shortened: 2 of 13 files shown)


jmcorgan:
Thanks @sh1ftred. The diagnosis is right, and sharing one constant between keygen and the search path is the right fix for those two drifting apart.

The main ask: keep reading `/etc/fips`. Add `/usr/local/etc/fips` to the search path instead of replacing `/etc/fips` with it, and warn once at startup when a file sits at the old path and not the new one. That covers the config, the ACL files and the hosts file together.

Why it matters is that both failures are silent. A missing config isn't an error, so anyone with a config at `/etc/fips` loses every setting in it, and if that file had `persistent: true` the node comes back with a new npub on each restart. A `peers.deny` there was being read before this change and isn't after, and a missing deny list isn't an error either, so peers it used to refuse now get in.

PR #129 moves these same paths for macOS and FreeBSD and already does both of the above, including a `warn_on_legacy_config_paths` you can lift. Worth syncing the two, since one has to merge into the other.

The changelog says the constants use `%ProgramData%\fips\` on Windows. Only the hosts path does; both ACL paths stay at `/etc/fips` everywhere but macOS.

Please also add a changelog line telling Mac users to move existing files, and update `docs/how-to/persistent-identity.md`. The wider docs pass can wait.

The comment-trimming commit can fold into the keygen commit it edits, and two messages reference `016aec4` and `9608a3b`, which aren't in the branch any more.

No need to strip the comments you mentioned. They're fine as they are.

fr34aky:
Hi @sh1ftred , 
author of #129 here. Happy to adjust my code accordingly and config gate macos + freebsd whenever possible after your merge to maint. So that we have a clear base for macos and the code can in most cases be reused for freebsd. If you have questions concerning #129 feel free to ask me.

docs/how-to/persistent-identity.md
```diff
@@ -72,6 +72,37 @@ present after the first successful daemon start. If the daemon never
...
+
+`fipsctl keygen` writes to `/usr/local/etc/fips/` by default on macOS.
+The daemon still probes `/etc/fips/fips.yaml` as a fallback (so an
+existing install is not broken by an upgrade), but the macOS packaging
+only installs files under `/usr/local/etc/fips/`.
+
+If you have files in `/etc/fips/` from a manual install, move them:
+
+```sh
+sudo mv /etc/fips/fips.yaml /usr/local/etc/fips/fips.yaml
```
sh1ftred, line 94: Hey @jmcorgan, which location is considered the canonical location? Do I have /usr/local/etc/fips/ only becuase I installed long back? If yes, these move commands should be opposite.

sh1ftred:
Thank you both of you for the feedback :)
I think I've addressed all your points. It works fine on my mac. Let me know if I can help with something else. 

@jmcorgan @fr34aky

jmcorgan:
Merged on `maint` as a single commit with your authorship, and carried up through `master` and `next`.

Three changes folded in on landing. `search_paths()` reads `SYSTEM_CONFIG_DIR` rather than repeating the literal. The persistent-identity how-to no longer claims the daemon warns about a stranded `fips.yaml`; it does not, because both directories stay on the search path.

The third is behavioural. The daemon derives the identity key directory from whichever config file loaded last, so adding `/usr/local/etc/fips/fips.yaml` moves it: a host with a config at both locations resolved `fips.key` to the new directory, found none, and under `persistent: true` generated a fresh keypair, changing its npub, routing address and mesh IPv6 with nothing in the log. It now adopts a key at the old path and warns instead. The fallback applies only to keys resolved from the system config directory.

PR #133, opened 2026-08-09
transport: make address clones allocation-free

mmalmi:
## What

`TransportAddr` is cloned in receive, handshake, and peer-bookkeeping paths. Its private `Vec<u8>` backing makes every clone allocate and copy the address.

Store the immutable bytes in `Arc<[u8]>` instead. Public construction, value equality and hashing, wire bytes, display, and `no_std + alloc` compatibility are unchanged.

## Measurement

The committed benchmark compares the former `Vec<u8>::clone` operation with the actual public `TransportAddr::clone`:

```sh
cargo bench --bench transport_addr_clone
```

Apple M4 Pro, macOS 26.6.1, Rust 1.94.1:

| Address bytes | Former | Shared | Change |
| ---: | ---: | ---: | ---: |
| 6 | 18.93 ns | 3.75 ns | -80.2% |
| 18 | 20.04 ns | 3.70 ns | -81.5% |
| 58 | 22.02 ns | 3.63 ns | -83.5% |

A counting pass measured 1 -> 0 allocations per clone. A separate lifecycle pass measured construct + one clone at 2 -> 1 allocations and 42.7-45.5 ns -> 23.7-25.0 ns.

As an end-to-end check, upstream's real-daemon Docker benchmark ran three ABBA blocks with six 10-second single-stream legs per mode. Throughput was 2,689 -> 2,675 Mbps (-0.5%) and combined endpoint CPU was 1.655 -> 1.660 CPU-sec/Gbit (+0.3%), with mixed block directions. The full-daemon effect is therefore neutral within noise; the claim is limited to removing clone allocation and cost.

## Testing

- `cargo bench --bench transport_addr_clone`
- `cargo test --lib` (1,645 passed, 4 ignored)
- `cargo check --all-targets`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo fmt --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `TransportAddr` clones now share immutable address bytes instead of
+  reallocating and copying them.
+
 - The Ethernet transport's per-interface `discovery` flag was renamed to
   `listen` (`transports.ethernet.*`) to match the symmetric `announce`
   (transmit) / `listen` (receive) neighbor-beacon vocabulary. The old
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -124,3 +124,8 @@ path = "src/bin/fipstop/main.rs"
 name = "routing_next_hop"
 path = "benches/routing_next_hop.rs"
 harness = false
+
+[[bench]]
+name = "transport_addr_clone"
+path = "benches/transport_addr_clone.rs"
+harness = false
--- a/benches/transport_addr_clone.rs
+++ b/benches/transport_addr_clone.rs
@@ -1,0 +1,37 @@
+//! Clone-cost comparison for transport address backing storage.
+//!
+//! The `Vec<u8>` case preserves the former `TransportAddr` clone operation;
+//! the shared case calls the current public `TransportAddr::clone` directly.
+
+use std::hint::black_box;
+
+use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+use fips::TransportAddr;
+
+const ADDRESS_BYTES: [usize; 3] = [6, 18, 58];
+
+fn bench_transport_addr_clone(c: &mut Criterion) {
+    let mut group = c.benchmark_group("transport_addr_clone");
+
+    for len in ADDRESS_BYTES {
+        let bytes = vec![0x5a; len];
+        let former_vec_backed = bytes.clone();
+        let current_shared = TransportAddr::from_bytes(&bytes);
+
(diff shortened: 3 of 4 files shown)


jmcorgan:
Thanks for the contribution. The implementation is correct: the wire format is untouched, hashing and `Debug` render identically in either representation so map-key behaviour and log output are unchanged, and both new tests discriminate. One change is needed before it lands.

**`from_socket_addr` gets slower, and it is the constructor on the receive path.** At `src/transport/mod.rs:343-348` it formats into a `Vec` and passes it to `new`. On master `new` is a move, so the call is one allocation; with this change it is `bytes.into()`, and `Vec<u8>` to `Arc<[u8]>` cannot reuse the buffer because `Arc` needs its refcount header ahead of the data, so it allocates again and memcpys. That runs once per received packet at `src/transport/udp/mod.rs:517` and `:569`, and at `src/transport/udp/io/connected/drain.rs:154`. Per packet the path goes from one construct plus one clone allocation to two construct plus zero. A wash, not a win, and invisible to the benchmark because that constructs with `from_bytes`, where both versions allocate once. It is also a plausible mechanism for the -0.5% end-to-end you measured.

A stack buffer makes it a genuine one allocation, better than master rather than level:

```rust
pub fn from_socket_addr(addr: std::net::SocketAddr) -> Self {
    use std::io::Write;
    let mut buf = [0u8; 64];
    let mut cursor = &mut buf[..];
    write!(&mut cursor, "{addr}").expect("64 bytes is enough for any SocketAddr");
    let len = 64 - cursor.len();
    Self::from_bytes(&buf[..len])
}
```

64 bytes is safe: the longest form a `SocketAddr` can display is 55. I compiled and round-tripped this before suggesting it.

A rebase is needed. Two changes landed on master on 2026-08-10 and the UDP transport was reorganized between them. The conflict is only in `CHANGELOG.md`; I simulated the merge and there is no source conflict.

Two smaller things, neither blocking. `new` quietly went from taking ownership to copying while its rustdoc at `types.rs:101` still reads as though it takes the `Vec`, which is worth a line so a future caller does not reach for it in a receive loop and get a silent per-packet copy. And `types.rs:202` pulls in `std::collections::HashSet` in a file whose header says these types stay free of `std` for the sans-IO cores; only the test escapes, so a comment marking it test-only is enough.

One question: the benchmark is single-threaded, and the drain builds one address and clones it per packet on one thread while another drops them, so a clone-and-drop pair is now an atomic on a shared cacheline where it used to be an independent malloc. I expect that is still well ahead, and your neutral end-to-end numbers agree, but it is the site your change most improves on paper and the one the benchmark says least about. Did you look at the multi-threaded case?

jmcorgan:
Closed due to lack of follow up.

PR #134, opened 2026-08-09
routing: eliminate next-hop allocations (21-24% faster)

mmalmi:
## What

Routing currently collects peer addresses and eligible candidates into temporary vectors before choosing the next hop. Visit borrowed peer entries and perform eligibility checks and `(cost, distance, address)` selection in one pass instead.

This removes address/coordinate clones and temporary allocations while preserving bloom filtering, sendability, strict progress, and tie ordering. Characterization tests cover those semantics.

## Measurement

Uninstrumented Criterion A/B on an Apple M4 Pro running macOS 26.6.1 and Rust 1.94.1, 50 samples:

| Peers | Before | After | Change | Allocations before -> after |
| ---: | ---: | ---: | ---: | ---: |
| 8 | 1.403 us | 1.072 us | -23.6% | 6 -> 0 |
| 32 | 5.587 us | 4.250 us | -23.9% | 20 -> 0 |
| 128 | 22.993 us | 18.117 us | -21.2% | 70 -> 0 |
| 256 | 45.836 us | 36.007 us | -21.4% | 135 -> 0 |

At 256 peers this is about 45.8 -> 36.0 seconds of single-thread benchmark time per million selections.

## Testing

- `cargo test --lib` (1,646 passed, 4 ignored)
- `cargo check --all-targets`
- `cargo clippy --all-targets -- -D warnings`
- `cargo fmt --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Routing next-hop selection now visits borrowed peers and coordinates instead
+  of allocating candidate snapshots for each forwarded packet.
+
 - The Ethernet transport's per-interface `discovery` flag was renamed to
   `listen` (`transports.ethernet.*`) to match the symmetric `announce`
   (transmit) / `listen` (receive) neighbor-beacon vocabulary. The old
--- a/benches/routing_next_hop.rs
+++ b/benches/routing_next_hop.rs
@@ -1,23 +1,18 @@
-//! Micro-benchmark quantifying the per-forwarded-packet heap-allocation cost
-//! of the routing next-hop candidate-assembly path.
+//! Micro-benchmark quantifying the per-forwarded-packet cost of routing
+//! next-hop selection before and after candidate assembly was fused.
 //!
-//! `find_next_hop` runs once per forwarded data packet. Its sans-IO core
-//! assembles a `Vec<Candidate>` by enumerating every peer through the
-//! `RoutingView` seam: `peer_addrs()` materializes a `Vec<NodeAddr>` of all
-//! peers, the survivors are snapshotted (each cloning its `TreeCoordinate`),
-//! and the result is collected into a second `Vec`. This bench measures that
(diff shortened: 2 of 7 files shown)


jmcorgan:
Merged, thanks. It landed on master as `53d80e6` rather than `dc22e540`: master had moved on under it, so the commit was replayed onto the current tip. Your authorship is on it unchanged. The only thing that needed resolving was the CHANGELOG entry, which collided with the macOS receive batching from #135, and both entries are in.

I reviewed it against the previous selector line by line before landing, and I agree it is behaviour-preserving: the filter set, the metric, the strictness of the progress check and the tie-break all come out the same. The tie-break was the part I expected to be a problem and is not. Both versions enumerate the same `HashMap`, so both were already order-dependent in principle, and the comparison is a strict lexicographic minimum over cost, then distance, then address, whose last component is a unique map key. That makes the result independent of enumeration order either way, and the test you added for exactly that property does fail if the address comparison is flipped.

One thing worth flagging, and it is about the evidence rather than the change: the benchmark does not exercise the code either arm claims to compare. Both arms are hand-written reproductions over the public types, and the "current" arm is a hand-inlined loop over the peer map rather than the shipped generic selector with its `for_each_peer` closure, the GAT-bearing peer type and the five trait calls per peer. So it assumes the inlining it is meant to demonstrate, and the 21-24% is a measurement of the reproduction rather than of the selector that ships. The scenario setup itself is honest: peer counts, the 50% bloom survivor rate and the coordinate depth are all unchanged from the baseline, so nothing is tilted in the change's favour, and the baseline arm does faithfully reproduce the old per-predicate map lookups.

I would still expect a real improvement here, since removing a per-packet allocation and a coordinate clone is a win on its own terms. If you want the figure to mean what it says, pointing both arms at the actual `select_best_candidate` would do it. Not a blocker and nothing needed from you.

Separately, reviewing this turned up that the shell adapter between the node's peer map and the routing core has no test coverage at all: stubbing out its bloom check, so that filtering is disabled on the forwarding path entirely, leaves the whole library suite green. That predates your change and is not caused by it, but your commit rewrote all six of those methods, so it is on my list to fix.

PR #135, opened 2026-08-09
udp: batch macOS connected-peer receive with recvmsg_x

### LINKED ISSUE #100: transport/macOS: per-peer drain uses single-packet recv instead of recvmmsg batching
PR #91 introduced a per-peer `connect(2)`-ed UDP socket model (`src/transport/udp/connected_peer.rs`) and a dedicated drain thread per established peer (`src/transport/udp/peer_drain.rs`). The drain thread's `recv_drain` loop uses `recvmmsg(2)` on Linux to batch up to 32 datagrams per syscall, but falls back to a single-packet `recv` loop on macOS.

The result is asymmetric on macOS established peers:

- Each datagram still costs one syscall plus one task wakeup, even though the wildcard listen socket got `recvmsg_x(2)` batching in PR #89 / commit `59225ccf`.
- Established peers on Linux get both wins from PR #91: kernel 5-tuple caching from the `connect(2)` (eliminating per-packet route lookup on the send side) AND `recvmmsg` batching on the receive side.
- Established peers on macOS get only the kernel 5-tuple caching; the receive-side batching win is left on the table.

Handshake and NAT-traversal traffic is not affected since those continue to land on the wildcard listen socket, which already uses `recvmsg_x` batching on macOS.

## Suggested fix

Add a Darwin path to `peer_drain.rs::recv_drain` that uses `recvmsg_x` against the per-peer connected socket, mirroring the existing wildcard-path use of `recvmsg_x` in `src/transport/udp/socket.rs`. The `msghdr_x` FFI extern and the batching contract are already in tree; the new piece is wiring them to a connected UDP fd in the drain loop.

## Out of scope

This is not a blocker. macOS established-peer throughput already improves vs. master because of the `connect(2)` + 5-tuple cache win. The gap is "macOS established peers get a smaller per-peer batching win than Linux", not "macOS regresses".

Refs: maintainer review of the PR #91 land at master `0a5c367e`.

mmalmi:
Fixes #100.

## What

The connected-peer drain issues one `recv` syscall per queued datagram on macOS. Reuse the existing Darwin `recvmsg_x` ABI to receive up to 32 connected datagrams per call.

The helper preserves partial-batch and `EINTR` behavior. A native connected-socket test queues 40 indexed datagrams, crossing the batch boundary, and verifies every length and payload in order.

## Measurement

Fixed-work loopback benchmark on an Apple M4 Pro running macOS 26.6.1: one persistent connected receiver, one sender, 100-byte datagrams, three ABBA blocks with six 1,000,000-packet legs per mode. Receiver CPU is the thread's user + system time from macOS `THREAD_BASIC_INFO`, excluding sender work.

| Mode | Receiver CPU / million packets | Receive calls / million packets | Wall time / million packets |
| --- | ---: | ---: | ---: |
| `recv` | 1.362 s | 1,285,531 | 2.305 s |
| `recvmsg_x` | 1.260 s | 358,524 | 2.320 s |

Receiver-thread CPU falls 7.5% and receive calls fall 72.1%. Wall time and separate packet-rate tests were neutral within loopback variance, so this is a CPU/syscall-amortization improvement rather than a throughput claim.

## Testing

- `cargo test` (all unit and binary tests passed)
- connected burst test repeated 200 times
- `cargo build`
- `cargo clippy --all-targets -- -D warnings`
- `cargo fmt --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Connected UDP peer drains now batch macOS receives with `recvmsg_x(2)`,
+  matching the wildcard UDP receive path instead of issuing one `recv(2)`
+  syscall per queued datagram.
+
 - The Ethernet transport's per-interface `discovery` flag was renamed to
   `listen` (`transports.ethernet.*`) to match the symmetric `announce`
   (transmit) / `listen` (receive) neighbor-beacon vocabulary. The old
--- a/src/peer/connected_udp/drain.rs
+++ b/src/peer/connected_udp/drain.rs
@@ -9,7 +9,7 @@
 //!
 //! This module owns the drain side: spawn one OS thread per connected
 //! socket, drain into a fixed-size batch (`recvmmsg(2)` on Linux,
-//! repeated nonblocking `recv(2)` on Darwin), push each packet into
+//! `recvmsg_x(2)` on Darwin), push each packet into
 //! the existing `packet_tx` (the same channel that the wildcard listen
 //! socket feeds), and exit cleanly when the parent signals shutdown
 //! via a self-pipe.
@@ -334,9 +334,9 @@ fn drain_packets(fd: RawFd, backing: &mut [Vec<u8>], lens: &mut [usize]) -> io::
     recvmmsg_drain(fd, backing, lens)
 }
 
-#[cfg(not(target_os = "linux"))]
+#[cfg(target_os = "macos")]
 fn drain_packets(fd: RawFd, backing: &mut [Vec<u8>], lens: &mut [usize]) -> io::Result<usize> {
-    recv_drain(fd, backing, lens)
+    crate::transport::udp::io::recvmsg_x_drain(fd, backing, lens)
 }
 
 /// One-shot `recvmmsg(2)` on a non-blocking fd. Returns the number of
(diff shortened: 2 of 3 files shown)


jmcorgan:
Merged as-is, thanks.

I took the opportunity to finish a refactor of the UDP transport I had been part way through, so there is one more commit on top of yours. It is behaviour-neutral, but it moves the file your change lives in, so worth a heads-up if you have anything else in flight here.

`src/transport/udp/io.rs` becomes a directory:

- `io/unix.rs` holds what every Unix target shares: socket setup, buffer sizing, the synchronous calls, and the `AsyncFd` wrapper.
- `io/linux.rs`, `io/macos.rs` and `io/unix_other.rs` hold only the three things that genuinely differ, behind identical signatures: enabling the kernel drop counter, the ancillary control buffer and its reader, and batched receive. `unix.rs` picks one of them once at the top, so its bodies carry no `cfg` at all. Your `recvmsg_x_drain` and the `msghdr_x` FFI now live in `io/macos.rs`.
- `io/windows.rs` stays a separate backend over `tokio::net::UdpSocket`, sharing nothing with the Unix side but the two exported type names.
- `io/connected/` gathers the whole connected-socket fast path, which had been split across two trees: `fd.rs` builds the fd, `socket.rs` is the owning handle and `drain.rs` is the drain thread. The last two came over from `src/peer/connected_udp/`, which had no dependency on `peer` at all.
- `sockopts_macos.rs` moves in as `io/macos_sockopts.rs`, since that module is now its only consumer.

Nothing about the syscall sequences changed, and the paths you would reach for from outside are the same.

PR #136, opened 2026-08-10
feat(node): publish the DNS responder's bound address for embedders

Origami74:
Adds `Node::dns_local_addr() -> Option<SocketAddr>`, the DNS companion to the app-owned TUN seam already on `master`.

## Why

An embedder that owns the TUN fd generally has no system DNS socket to aim at the built-in `.fips` responder. On Android, `VpnService.Builder.addDnsServer()` takes an address with no port — the resolver always uses 53, which an unprivileged app UID cannot bind — and it points the resolver *into* the tunnel, so `.fips` queries arrive as IPv6/UDP packets on the app's own fd rather than at any socket FIPS holds. The responder ends up listening on a port nothing on the device will ever query.

The app can still use it: lift the DNS payload out of the packet it read, send it to the responder over an ordinary UDP socket of its own, and splice the answer back into a reply packet. That matters because answering a `<npub>.fips` query is what registers the peer's pubkey in the identity cache, and a FIPS address is SHA-256(pubkey) truncated twice — the key is not recoverable from the IPv6 address. An app that resolves the AAAA itself leaves the cache empty and the first packet to the resolved name is rejected with ICMPv6 Destination Unreachable. Direct neighbours mask it, since their identity arrives with the Noise handshake.

## What it does

- **`Node::dns_local_addr()`** — a one-shot read taken after `start()` returns and before the node is backgrounded, which is the only window in which an embedder running `run_rx_loop` still holds a `&Node`. Reports the address read back off the bound socket, so `dns.port = 0` yields the kernel-assigned port. `None` when no responder came up — config alone cannot tell you that, since a failed bind only warns and leaves `dns.enabled = true` with `bind_addr` naming a target nothing listens on. Explicitly not a liveness feed.
- **`retract_child_publications`** — clears the published address on `stop()`, and on a runtime child exit. See the disposition note below: the runtime half is dormant and documented as such.
- **Design doc** — an App-Owned DNS Path section in `docs/design/fips-ipv6-adapter.md` beside the App-Owned TUN one it mirrors, plus status rows for both.

No new dependencies. Nothing in the responder's start-up needed changing: `bind_dns_socket` is plain socket2, `lookup_mesh_ifindex` returns `None` with no system TUN so the mesh filter self-disables, and `HostMapReloader` on an absent hosts file settles at a no-op stat.

## Review round 2 — what changed

Rebased onto current `master` (was forked from `83c4e800`, ~28 behind by the time I picked this up; `master` moved twice more during the run, so the head is a fresh rebase). Plus:

- **`dns_local_addr_stays_none_when_the_bind_fails` rewritten.** It named a TEST-NET-3 address, which `net.ipv4.ip_nonlocal_bind = 1` defeats — reds on a maintainer box, green in CI at the default `0`. Exactly the wrong direction. It now forces `EADDRINUSE` against a socket the test holds open, per your suggestion; `bind_dns_socket` sets neither `SO_REUSEADDR` nor `SO_REUSEPORT`, so it is deterministic on Linux and macOS. Verified it discriminates by pointing the config at a free port instead and watching it red.
- **Rustdoc reworded** to a one-shot read after `start()`, with the liveness half explicitly disclaimed as needing a general way to read live state from a backgrounded `run_rx_loop`. The design-doc section's closing sentence had the same overclaim and got the same treatment.
- **`retract_child_publications` documented as dormant for `Dns`**, with the reason: `run_dns_responder` never returns, so nothing produces the event. Kept per your call, so the producer fix doesn't have to rediscover the consuming side. The panicking-responder gap is noted there too.
- **Test renamed** `dns_local_addr_is_retracted_when_the_responder_dies` → `retract_child_publications_clears_the_dns_address`, since no responder dies in it and the rx-loop wiring is covered by nothing. You flagged it without asking; it was a fair hit.
- **App-owned TUN status row** added alongside the DNS one.

## Standing caveats

**No integration suite.** The `testing/` harness runs the daemon in containers; there is no embedder in a container, so the daemon never reaches this path. Covered by three unit tests instead.

**I still cannot run `./testing/ci-local.sh`** — no Docker on this machine. `fmt`, `clippy --all-targets -D warnings`, `cargo test` (1672 passed), the Android `cargo ndk -t arm64-v8a clippy --lib -D warnings` gate, and the doctest are all clean locally.

**Correction to my earlier body:** I cited `testing/check-log-strings.py` as evidence the `dns-resolver` log assertion was safe. It is not — as you noted, it skips files without the literal `docker logs`, and that suite reads via `docker exec`, so the line was never in scope. Your reading is the one that holds: the field name is unchanged, only the value's source moved, and the scenario's default `::1` bind renders identically because `local_addr()` on a loopback IPv6 socket comes back with `sin6_scope_id` zero.

### DIFF
--- a/docs/design/fips-ipv6-adapter.md
+++ b/docs/design/fips-ipv6-adapter.md
@@ -330,6 +330,47 @@ and the [TUN-Side TCP MSS Clamping](#tun-side-tcp-mss-clamping). The embedder is
 therefore responsible for routing only `fd00::/8` to its TUN (so only mesh-bound
 packets arrive) and for clamping TCP MSS on outbound SYNs.
 
+### App-Owned DNS Path (embedded hosts)
+
+An embedded host that owns the TUN fd generally has no system DNS socket to aim
+at the responder either. On Android, `VpnService.Builder.addDnsServer()` takes an
+address with no port — the resolver always uses 53, which an unprivileged app UID
+cannot bind — and it points the resolver *into* the tunnel, so `.fips` queries
+surface as IPv6/UDP packets on the app's own fd rather than at any socket FIPS
+holds. The [DNS responder](#dns-integration) itself needs no changes for this:
+its bind is a plain UDP socket, and with no system TUN the
+[mesh-interface filter](#mesh-interface-query-filter) self-disables because the
+interface name does not resolve.
+
+`Node::dns_local_addr()` closes the gap. It reports the address read back off the
+bound socket — so a `dns.port = 0` config yields the port the kernel assigned —
+and is `None` when no responder came up. The embedder lifts the DNS payload out
+of the packet it read, sends it to that address over an ordinary UDP socket of
+its own, and splices the answer back into a reply packet.
+
+It is a one-shot read taken after `start()` returns and before the node is moved
+into a background task, because `run_rx_loop` then borrows the node exclusively
+for its whole lifetime and no `&Node` remains to call it on. By that point the
+value is settled: the responder is either up for the rest of the node's life or
+it never came up.
+
+Proxying to the responder rather than resolving in the app is what keeps the
+[identity cache](#identity-cache) warm. Answering a `<npub>.fips` query is what
+registers that peer's public key, and a FIPS address is a truncated hash of a
+hash of the pubkey — the key cannot be recovered from the IPv6 address alone. An
+app that resolves the AAAA itself leaves the cache empty, and the first packet to
+the resolved name is rejected with ICMPv6 Destination Unreachable. Direct
+neighbours mask the omission, since their identity arrives with the Noise
+handshake and never needed resolving.
+
+The address is retracted on `stop()`. A retraction hook for a responder that
+exits on its own at runtime is wired on the consuming side, but is dormant:
+`run_dns_responder` never returns, so nothing produces the `Child::Dns` exit
+event it consumes. Watching a responder that dies mid-run is therefore not
+something an embedder can do today, and would in any case need a way to read
+live node state from a backgrounded `run_rx_loop` — a general gap rather than a
+DNS-specific one.
+
 ## Implementation Status
 
 | Feature | Status |
@@ -343,6 +384,8 @@ packets arrive) and for clamping TCP MSS on outbound SYNs.
 | TCP MSS clamping (SYN + SYN-ACK) | **Implemented** |
 | DNS service (.fips domain) | **Implemented** |
 | DNS responder mesh-interface filter | **Implemented** |
+| App-owned TUN (`Node::enable_app_owned_tun`) | **Implemented** |
+| App-owned DNS path (`Node::dns_local_addr`) | **Implemented** |
 | Port-based service multiplexing (port 256) | **Implemented** |
 | IPv6 header compression (format 0x00) | **Implemented** |
 | Per-destination route MTU (netlink) | Planned |
--- a/src/node/dataplane/rx_loop.rs
+++ b/src/node/dataplane/rx_loop.rs
@@ -271,6 +271,11 @@ impl Node {
                 // `PublishState`; other variants are ignored defensively.
                 maybe_child = child_exit_rx.recv() => {
                     if let Some(child) = maybe_child {
+                        // Drop anything the dead child published for embedders
+                        // (e.g. the DNS responder's bound address) before
+                        // republishing health, so nothing outside the node can
+                        // observe an address the listener no longer answers on.
+                        self.retract_child_publications(child);
                         let actions = self
                             .supervisor
                             .fsm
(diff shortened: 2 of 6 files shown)


Origami74:
All four addressed, head is now `93e07a3851dd944a41dc5c0de8b9aab72357b5f3`.

**Bind-failure test.** Took your suggestion exactly: the test now binds a UDP socket to `[::1]:0`, reads its port, and points `dns.bind_addr` / `dns.port` at it, giving `EADDRINUSE` regardless of sysctl. Thanks for tracking that to `ip_nonlocal_bind` rather than just reporting a red — I would not have found it from here, and "passes in CI, reds on the maintainer's box" is the right way to weigh it. I checked it discriminates the way you check things: pointing the config at a free port instead reds it at the same assertion.

**The dormant retraction.** Taking you at your word and leaving it in, with a `**Dormant for `Dns` as written.**` paragraph on the helper stating that `run_dns_responder` never returns and the `Child::Dns` send after it is unreachable, so nothing produces the event. Also recorded the panic-unwinds-past-the-send gap as general rather than DNS-specific. Appreciate you disclosing it instead of letting me ship a hook I believed was live.

**Rustdoc.** Reworded to a one-shot read after `start()` and before backgrounding — the only window a `&Node` exists — with the liveness half explicitly disclaimed as needing a general answer for reading live state out of a backgrounded `run_rx_loop`, not an accessor-by-accessor one. The design-doc closing sentence had the same overclaim and got the same fix.

**Rebase.** Done — and worth flagging that `master` moved twice while I was working, so this is a fresh rebase onto the tip as of now rather than onto the head you reviewed against.

Two smaller ones: the test is renamed to `retract_child_publications_clears_the_dns_address`, since you were right that it was named for a scenario it does not run; and the status table gained the app-owned TUN row next to the DNS one.

One correction of my own, on your point about `check-log-strings.py`: you are right that it is not evidence here, and I should not have cited it as a gate for that assertion. Your explanation of *why* the `dns-resolver` regex still matches is also better than mine — the field name never changed, only the value's source, and the zero `sin6_scope_id` is the reason the rendering is identical. I had the right conclusion for a vaguer reason.

Still no Docker here, so `./testing/ci-local.sh --only dns-resolver` remains unrun on my side. Everything else is clean locally: 1672 tests, fmt, `clippy --all-targets -D warnings`, the Android `cargo ndk clippy` gate, doctest.

jmcorgan:
Merged. This is on `master` as `672d828e`, pushed.

One change at merge, and it reverses something I told you in the review. I said no CHANGELOG entry was needed, on the grounds that the app-owned TUN seam had landed without one. That precedent had already stopped being true when I said it: `26f17105` ("Cover the master-only work since 0.4.0 in the changelog") went back and gave that seam an `[Unreleased] / Added` entry naming `Node::enable_app_owned_tun()`, so it is the sweep rather than the original commit that sets the convention now, and leaving this bare would have put the entry straight back on the list that sweep was clearing. Rather than send you back for one bullet on an otherwise finished pull request, I wrote it and folded it into your commit:

> `Node::dns_local_addr()`, the DNS companion to the app-owned TUN seam above. An embedder whose resolver is pointed into the tunnel has no system socket aimed at the built-in `.fips` responder, so the accessor reports the address read back off the bound socket: `dns.port = 0` therefore yields the kernel-assigned port, and it returns `Some` only while the responder is up. Read it once, after `start()` returns and before the node is moved into a background task; it is not a liveness feed (#136).

It went in as its own bullet rather than as a sentence on the Android-ready-core entry, which is where it fits by subject. That entry carries no pull request reference, so appending `(#136)` to it would have read as attributing the whole `target_os` gating change to this contribution.

Two consequences of the fold worth stating. Your authorship is preserved on the commit, with me as committer. And because amending changed the hash, `672d828e` is not `93e07a38`, so GitHub does not recognize this as merged and will not close the pull request on its own. I am closing it by hand, and that close means merged rather than rejected.

All four asks came back discharged and I checked them by reading rather than by running this time, which the re-review says at each claim. The `EADDRINUSE` rewrite is host-independent for the reason you gave: neither your socket nor `bind_dns_socket` sets a reuse option, and I confirmed the second half in the vendored tokio and mio sources rather than assuming it. The rustdoc rewording is the strongest of the four responses, since you wrote the constraint into the docs instead of deleting the overclaim. Verification at the merged tip: `fmt` clean, `clippy --all-targets -D warnings` clean, and the unit tests at 1749 passed, 0 failed, 4 ignored.

Two things remain open, neither of them yours. The `dns-resolver` scenario is still unrun by either of us against a build of this work, so the regex claim is settled by inspection and the suite's actual state at this commit is not something either of us can assert. And the wiring test for the rx-loop call site is ours to write with the producer fix, not yours: the `Child::Dns` producer is still statically unreachable, so that line cannot be exercised yet, and the test that pins it is recorded against the producer fix on our side. Your helper starts working the moment we land it.

PR #138, opened 2026-08-11
fix(macos): secure control socket runtime path

erskingardner:
Not sure if this is helpful so please close if not - but I'd like to work on making this work better for mac and this came up as the highest value first step. feedback welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -238,6 +238,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- The packaged macOS daemon now recreates and binds its control socket at
+  `/var/run/fips/control.sock` instead of falling through to the shared
+  `/tmp/fips-control.sock` path after boot. The previous fallback also made the
+  root daemon attempt to chown `/tmp` itself to the `fips` group because socket
+  setup unconditionally changed the parent directory. A privileged macOS
+  process now selects the private runtime path before its leaf exists, clients
+  follow it once created, and socket setup changes ownership and mode only for
+  a private parent directory it atomically created.
+
 - Nostr NAT traversal signals are now sent only to relays the client pool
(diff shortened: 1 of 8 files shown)


jmcorgan (changes requested):
This is helpful, don't close it. The `/tmp` chown is a real bug in shipped code and the macOS bootstrap gap is real. One thing blocks: the new ownership rule regresses Linux.

On the packaged Linux installs the socket is in `/run/fips`, and a non-root `fips` group member needs the containing directory group-accessible, not just the socket. systemd creates `/run/fips` via `RuntimeDirectory=fips`, and since none of the units set `User=` or `Group=`, it creates it `root:root 0750`. The tmpfiles fragment that sets the group runs only at install and at boot, and systemd deletes the runtime directory when the unit stops. So the daemon's unconditional chown of the parent is what has been re-establishing group ownership on every start. This PR gates that on the daemon having created the directory, which on Linux it never does. After the first `systemctl restart fips` following a boot, `/run/fips` is `root:root 0750` and `fipsctl` gives `Permission denied` to everyone but root until the next reboot.

Key the decision on which directory it is rather than on who created it. Every candidate the resolver produces ends in a `fips` leaf (`/run/fips`, `/var/run/fips`, `$XDG_RUNTIME_DIR/fips`); the `/tmp` fallback is `/tmp/fips-control.sock`, whose parent is `/tmp`. So secure the parent when its file name is `fips`, regardless of who created it. `ensure_socket_parent` can keep returning whether it created the directory for the log line, it just shouldn't gate the chown on its own. `access_setup_never_changes_an_existing_parent` will need to change with it.

jmcorgan:
Merged. The three commits are rolled up into one, rebased onto current `master`, and landed as `5e3892ed` with you as author. The commit message is rewritten in the project's style, as `CONTRIBUTING.md` describes.

Closing this here rather than letting GitHub do it, since the rollup changed the hash. The close means merged.

PR #139, opened 2026-08-20
feat(ble): advertise and read the L2CAP PSM in the BlueZ backend

fr34aky:
## What

Wires the bluer/BlueZ backend into the PSM seam this branch already carries (`psm.rs`, `ScanAdvert.psm`, learned-PSM dialing):

- `start_advertising(psm)` now emits the listener's PSM as the **Service Data — 16-bit UUID** AD structure specified in `psm.rs` (key `0x9C90`, two bytes little-endian), instead of ignoring the argument.
- The scanner reads a peer's advertised PSM and RSSI into `ScanAdvert`, instead of always yielding `psm: None`.

One file, `src/transport/ble/io.rs` (+45/−16).

## Why

Android (`listenUsingInsecureL2capChannel`) and macOS assign the L2CAP PSM; only BlueZ lets the application choose it. Before this change a BlueZ node was a legacy UUID-only advertiser: it never told mobile peers which PSM to dial, and never learned their OS-assigned one — so Linux↔Android BLE failed in both directions, while the seam built for exactly this problem sat unused on the BlueZ side.

## Trade-off: the local name

Emitting the service data costs the `local_name "fips"`: flags (3) + 128-bit UUID list (18) + service data (6) = 27 of the 31-byte legacy PDU, and a name no longer fits — the same budget `psm.rs` asserts at compile time.

## Compatibility

BlueZ↔BlueZ on the configured PSM keeps working: listeners still bind it, and a peer with no service data still decodes to `psm: None` and is dialed at the configured constant, exactly as before.

## Validation

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, full test suite green.
- Hardware: Linux laptop (BlueZ) ↔ Pixel (Android app embedding fips, OS-assigned PSM) — discovery and L2CAP mesh link established in both directions with this patch; failed both directions without it.

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -684,8 +684,15 @@ entries become no-ops. Communicates with BlueZ via D-Bus through the
 
 **Advertising and scanning.** When `advertise` is enabled, the transport
 advertises the FIPS service UUID continuously so that nearby nodes can
-discover and connect via L2CAP. When `scan` is enabled, the transport
-continuously scans for other FIPS nodes' advertisements. Discovered
+discover and connect via L2CAP, plus the L2CAP PSM its listener actually
+bound, as a service-data structure (see `src/transport/ble/psm.rs` for the
+wire layout and why platforms with OS-assigned PSMs need it). The
+advertisement carries no device name — alongside the PSM a name no longer
+fits the 31-byte legacy PDU, so the node shows up in generic Bluetooth
+scanners as an unnamed device with the FIPS UUID. When `scan` is enabled,
+the transport continuously scans for other FIPS nodes' advertisements and
+learns each peer's advertised PSM; a peer that advertises none is dialled
+at the configured `psm`. Discovered
 peers are probed immediately (L2CAP connect + pubkey exchange) with a
 cooldown (`probe_cooldown_secs`) to prevent rapid re-probing of the same
 address. If two nodes probe each other at the same time (cross-probe),
--- a/src/transport/ble/io.rs
+++ b/src/transport/ble/io.rs
@@ -183,10 +183,12 @@ mod bluer_impl {
         AdapterEvent, AddressType, DiscoveryFilter, DiscoveryTransport, adv::Advertisement,
     };
     use futures::StreamExt;
-    use std::collections::{BTreeSet, HashSet};
+    use std::collections::{BTreeMap, BTreeSet, HashSet};
     use std::pin::Pin;
     use tokio::sync::Mutex;
     use tracing::{debug, trace};
+
+    use super::super::psm;
 
     /// FIPS BLE service UUID.
     ///
@@ -194,6 +196,17 @@ mod bluer_impl {
     /// version/variant bits applied.
     pub const FIPS_SERVICE_UUID: bluer::Uuid =
         bluer::Uuid::from_u128(0x9c90_b790_2cc5_42c0_9f87_c9cc_4064_8f4c);
+
+    /// The PSM service-data key as a whole UUID.
+    ///
+    /// BlueZ speaks in full UUIDs, so [`psm::PSM_SERVICE_DATA_UUID16`] is
+    /// expanded through the Bluetooth base UUID
+    /// (`00009C90-0000-1000-8000-00805F9B34FB`). The controller emits it
+    /// back on the air as the 16-bit Service Data AD structure the wire
+    /// layout in [`psm`] specifies.
+    pub const PSM_SERVICE_DATA_UUID: bluer::Uuid = bluer::Uuid::from_u128(
+        ((psm::PSM_SERVICE_DATA_UUID16 as u128) << 96) | 0x0000_0000_0000_1000_8000_0080_5F9B_34FB,
+    );
 
     /// Map a bluer error to a TransportError.
     fn map_err(context: &str, e: bluer::Error) -> TransportError {
@@ -308,11 +321,12 @@ mod bluer_impl {
     }
 
     impl BleScanner for BluerScanner {
-        /// Yields address-only adverts.
+        /// Yields adverts with the PSM and RSSI when BlueZ can supply them.
         ///
-        /// BlueZ does not yet advertise or read the PSM service data (see
-        /// `super::super::psm`), so `psm` and `rssi` are reported as `None`
-        /// and peers are dialled at the configured PSM, exactly as before.
+        /// The PSM comes out of the peer's Service Data AD structure (see
+        /// `super::super::psm`); a peer that advertises none — a legacy
+        /// UUID-only advertiser — yields `psm: None` and is dialled at the
+        /// configured PSM, exactly as before.
         async fn next(&mut self) -> Option<ScanAdvert> {
             loop {
                 match self.events.next().await {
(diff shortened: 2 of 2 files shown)


jmcorgan:
Thanks for this. I read the change and the seam it plugs into, and I want to land it. One thing has to change before it can merge, and there are four smaller items you can take here or leave for a follow-up.

**On the problem.** The description matches what the code does, which is not always the case with a change this small. Before it, the BlueZ backend took a PSM argument and discarded it, and the scanner returned an advert whose PSM was always `None`. That only bites because of the platform asymmetry `psm.rs` describes: BlueZ lets an application choose its L2CAP PSM, Android and macOS assign one it cannot request, and there is no channel to learn a phone's number on before a connection exists. So the advertisement is the only place it can ride, and the seam for that was already built and already used on the Android side. This is the missing half.

**On compatibility.** I checked this in all four directions against the copy of the BLE transport on `master`, since a node built from a trunk can meet a node built from this branch, and I did not find a break. A `master` node admits a peer on `uuids.contains(FIPS_SERVICE_UUID)` and sets the same UUID in its discovery filter, so neither its scanner nor its filter reads the local name your change drops. A `master` node dials at its configured PSM, and a BlueZ listener on this branch still binds the number it is handed and returns it, with the default `0x0085` identical in `src/config/transport.rs` on both branches. In the other direction, a `master` advertiser carries no service data, so `decode_psm` is never reached, the PSM stays `None`, and the dialer falls back to the configured constant. And `decode_psm` ignoring trailing bytes means a later advertiser can extend the value without breaking a reader built against this version. Worth saying out loud that the visible cost is real even though no code reads it: a BlueZ node stops appearing as `fips` in `bluetoothctl` and in a phone's Bluetooth list, and becomes an unnamed device carrying a UUID.

**The one blocking item.** The commit ends with a `Co-Authored-By: Claude Fable 5` trailer and the pull request body carries a generated-with badge. This project does not take coding-assistant attribution in its history. That needs an amend and a force-push, which is the only reason I am raising it on an otherwise clean single-commit branch.

**The item I would most like your read on.** In `scan_probe_loop`, `learned_psm` is inserted into whenever an advert carries a PSM and is never removed from, and the scanner only produces an advert on `AdapterEvent::DeviceAdded`; every other adapter event, including `DeviceRemoved`, falls through the `Some(_) => continue` arm. This is existing code on the branch and not something your change introduced, but your change is what makes the path live on BlueZ, and the case it bites is the one the feature exists for: a phone whose PSM is assigned by its OS gets a different one when its app restarts, and a daemon that keeps running will dial the old number until BlueZ ages the device out of its object tree and re-adds it. Purging the address from `learned_psm` on `DeviceRemoved` would turn a stale PSM back into an ordinary fallback dial. I have no radio here, so this is a reading of the code and of BlueZ's device lifecycle rather than something I have reproduced; if you have already seen it recover in testing, I would rather hear that than have you write code for it.

**Three smaller things.**

The UUID expansion in `PSM_SERVICE_DATA_UUID` is correct as written, but a wrong shift there would produce a plausible-looking UUID that simply never matches any peer, which surfaces as silent discovery failure rather than a build error. `psm.rs` already sets the precedent with `test_key_is_the_leading_16_bits_of_the_fips_uuid`. One assertion that the constant equals `00009C90-0000-1000-8000-00805F9B34FB` would close it. I am not asking for coverage of the advertising or scanning paths themselves, which need a real adapter.

`psm.rs` argues that the PSM must ride the primary advertisement rather than the scan response, because an active-scan round-trip drops asymmetrically across chipsets and a peer that never answers would become undiscoverable rather than merely slower. BlueZ's advertising API takes a set of AD structures and chooses the placement itself, so nothing in this code enforces that requirement. A comment saying so would help the next reader, and I am curious whether your laptop-to-phone test used passive or active scanning.

`docs/reference/configuration.md`, under "Advertising and scanning", still says the transport advertises the FIPS service UUID, which is now incomplete: the advert also carries the PSM and no longer carries a name. Happy to take that one myself if you would rather not touch the docs tree.

fr34aky:
Thanks for the close read — all five items are addressed: three in the amended commit (now `1550455`, force-pushed), two as answers here.

**Attribution (blocking).** Done — the trailer is out of the commit and the badge out of the PR body.

**The `learned_psm` read.** The staleness is half-handled already, and the half that exists is outside this commit's footprint: `scan_probe_loop` in `mod.rs` forgets a learned PSM on both failing dial arms (connect-error and connect-timeout — "A learned PSM that does not answer is stale — forget it"), and `test_a_failed_dial_forgets_the_learned_psm` pins it. This commit touches only `io.rs`, so that path is integration/platform's own. Concretely: after a phone's app restart reassigns its PSM, a long-running daemon dials the old number exactly once, forgets it, and is back to the ordinary fallback dial — which for an Android peer also fails, since its listener was never on the configured constant. The residual gap is the re-learn, and there your reading of the lifecycle matches mine: the scanner yields only on `DeviceAdded`, so the fresh PSM arrives only once BlueZ expires the temporary device and re-adds it on a later advert; until then the address sits in `PendingProbes` backing off (failures never evict it, per `failures_never_evict_the_address_itself`). So the failure window is bounded by BlueZ's device lifecycle rather than by a dial-forever loop — but I have not reproduced the restart-reassignment case on radio, only ordinary discovery and linking. I can run exactly that scenario (force-stop the phone app, relaunch, watch the daemon recover) and report the observed window here. If the age-out latency turns out ugly in practice, the follow-up shape would be surfacing removals through the scanner seam and purging on `DeviceRemoved` as you suggest — I would rather size that from a measurement than guess.

**The UUID expansion.** Test added alongside the constant: `psm_service_data_uuid_expands_the_key_over_the_base_uuid` asserts equality with `00009C90-0000-1000-8000-00805F9B34FB` parsed, mirroring the psm.rs precedent.

**Placement, and your scanning question.** Comment added at `start_advertising`: psm.rs's primary-PDU requirement is not something this code can enforce — BlueZ takes the AD structures and places them itself — and what keeps the requirement holding is the 27-of-31 arithmetic, which dropping the name is for. On the question: honestly, the laptop-to-phone test proves nothing about placement. Both ends scanned actively — BlueZ discovery sessions scan actively, and the Android side used the platform scanner's default, which is active too — so a PSM riding the scan response would have looked identical in that test. The arithmetic is the argument; confirming placement on the air would need a sniffer capture, which I do not currently have set up.

**Docs.** Took it here rather than leave it split from the change that made it stale: the "Advertising and scanning" paragraph in `docs/reference/configuration.md` now says the advert carries the listener's actual PSM as service data and no longer carries a name (with the visible cost you named — an unnamed device with a UUID), and that scanning learns a peer's advertised PSM with the configured `psm` as the fallback. Happy to reword if it fights the reference voice.

jmcorgan:
This is in. Your commit now sits on `fix/platform-ble`, a branch off `master` that collects the BLE and Android work, with you as its author. It applied cleanly and the branch is green.

I am closing this pull request because the change has moved, not because anything is wrong with it. It targets the platform integration branch, and retargeting it here would show seventeen commits instead of your one, since our branch rebuilt that work on a current trunk.

Thanks for the revisions. The attribution trailer, the UUID assertion, the placement comment and the configuration reference are all done. Follow-ups are welcome against `fix/platform-ble`.

PR #141, opened 2026-08-31
fix(lookup): accept our own lookup response instead of relaying it away

Origami74:
## The defect

An originator could misfile the answer to its own lookup as transit. Discovery reported "requests went unanswered" while the replies were in fact arriving and being reverse-path forwarded to a peer.

A request is flooded to every tree peer whose bloom filter claims the target. At a high fill ratio a false positive sends a copy out into the wider network, which can circulate it back to the originator. On arrival `classify_request` applied its only identity test, `target == my_addr`, which a lookup we originated never satisfies — so the copy was recorded in `recent_requests` as an ordinary transit entry keyed on our own `request_id`. When the target answered, `classify_response` consulted `recent_requests` first, matched that entry, and relayed our own answer to the peer that looped the request. The pending lookup was never satisfied, the ladder retried with a fresh id, and the race repeated.

The `None` arm's reasoning — *"not a request we transited, so it claims to answer one of ours"* — held only while our own ids stayed out of the dedup cache, and nothing enforced that.

It is a race, not a hard failure: when the genuine reply beats the looped copy the cache is still clean and the lookup succeeds. It bites hardest at the junction between a quiet subtree and a dense public network, where a lookup for a descendant goes both correctly downward and outward through a false positive, and the wide network returns it first.

## Symptom

`fipsctl probe` on an affected node, repeatably:

```
bloom       ok        0.5s  claimed by a peer filter, 2 peers asked
discovery   failed   15.0s  a peer filter claimed this address; 4 requests went unanswered
overall     failed, 15499 ms
```

The counter tell is `resp_unsolicited` pinned at zero while `resp_received == resp_forwarded` exactly — control never reached the originator arm. Affected nodes showed `resp_accepted` at 1 of 291 (0.34%).

## The change

Two changes, on both sides of the invariant:

- **`classify_response`** tests the pending lookups before `recent_requests`. A response naming a target with a lookup outstanding, carrying an id that lookup issued, is ours whatever the dedup cache holds. The id is fresh 64-bit randomness per attempt and the target signs over it, so an id we never issued still cannot match — the replay properties the `Unsolicited` arm protects are unchanged.

- **`classify_request`** drops a request whose id a pending lookup for that target issued, rather than recording it. Our own id never enters the transit cache, so a duplicate reply cannot be misrouted after we accept the first, and the returning copy is not forwarded a second time. The guard keys on the id, so another node's lookup for the same target still transits normally.

Either alone fixes the reported failure. Both, because the `classify_response` ordering is the correct precedence and the `classify_request` guard restores the invariant the `None` arm's reasoning rested on.

## Tests

- `classify_response_originator_wins_over_a_transit_dedup_record` — the exact race: id present in both maps, asserts `Originator` and that `response_forwarded` stayed false.
- `classify_request_drops_our_own_request_looped_back_to_us` — asserts `Duplicate`, empty `recent_requests`/`recent_by_peer`, and that a later response classifies `Originator`.
- `classify_request_still_transits_a_foreign_id_for_a_target_we_are_looking_up` — guards against over-matching on target.
- `classify_response_transit_on_fresh_forwarded_request` pinned the old precedence in its comment; corrected, behaviour unchanged.

Full suite: 2138 passed, 0 failed. `cargo fmt --check` clean.

## Validation

A/B tested on the reporting deployment with two packages built off this same base, differing only by this commit. Baseline (`66f0de8a13`) reproduces the four-attempt ladder failure; with the fix (`04131017be`) discovery completes.

On the fixed build `req_initiated` and `req_duplicate` track one another — every lookup's own request coming back and being dropped by the new guard, direct evidence the loop is real.

### DIFF
--- a/docs/design/fips-mesh-operation.md
+++ b/docs/design/fips-mesh-operation.md
@@ -272,6 +272,17 @@ follows the same path as the request. Greedy tree routing toward the
 `origin_coords` is used only as a fallback if the reverse-path entry has
 expired.
 
+**Originator check first**: a node tests its own outstanding lookups before
+consulting `recent_requests`. A request is flooded to every bloom-matching
+tree peer, and a bloom false positive can send a copy out into the wider
+network and back to the originator, which would otherwise file its own
+`request_id` as an ordinary transit entry and relay its own answer away. The
+originator arm therefore wins: a response naming a target with a lookup
+outstanding, carrying a `request_id` that lookup issued, is accepted here
+whatever the dedup cache holds. A returning copy of the request is likewise
+dropped as a duplicate rather than recorded, so the originator's id never
+enters the transit cache.
+
 **Response-forwarded flag**: Each `recent_requests` entry tracks whether a
 response has already been forwarded for that `request_id`. If a second
 response arrives (e.g., from convergent request paths that reached the
--- a/src/proto/lookup/core.rs
+++ b/src/proto/lookup/core.rs
@@ -258,6 +258,26 @@ pub(crate) fn classify_request(
         };
     }
 
+    // A request this node originated, flooded to every bloom-matching tree
+    // peer and circulated back to us by one of them. The only identity test
+    // below is `request.target == *my_addr`, and for a lookup we originated
+    // the target is someone else, so without this the copy is filed as an
+    // ordinary transit entry keyed on our own `request_id` — and the answer,
+    // when it comes, is reverse-path forwarded to the peer that looped the
+    // request instead of being accepted here. Dropping it as the duplicate it
+    // is also stops the copy being forwarded a second time: our flood already
+    // reached the peers that could carry it.
+    if lookup
+        .pending_lookups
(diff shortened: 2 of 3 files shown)


jmcorgan:
Merged. This landed on `maint` as `27870624` rather than on `master`: the lookup module is identical on the two lines, so the defect is on both, and a bug fix is authored on the earliest branch that carries it. It has already merged up to `master` and `next`.

I folded a CHANGELOG entry into your commit, under Unreleased / Fixed, credited to you. That is the only change from the branch as you pushed it, and you are still the author.

A follow-on commit gives the looped-back drop its own rejection counter, `req_own_loopback`, instead of recording it as an ordinary duplicate, since a returning copy of our own request has a nonzero floor in healthy operation and says nothing about the peer that delivered it. It adjusts one assertion in your test to match, and adds two comments recording why `request.origin` cannot be the identity test and how far the id correlator reaches.

Issue #143, opened 2026-09-03
Ethernet: a peer reachable over two interfaces has no way to choose the better path


Origami74:
I think this is very much related to our earlier talks about allowing
multiple transports for one peer.

---

*Everything below this line is LLM-written analysis of logs from a live
two-node setup.*

**FIPS version:** `0.6.0-dev` (`rev 6b9faa0c2f`)
**Rust:** `rustc 1.94.1 (e408947bf 2026-03-25)`
**OS:** macOS 14.8.9, arm64

MAC addresses below are placeholders; the topology and counts are from a real
two-node setup.

### What I expected

Two nodes joined by a direct cable use the cable. If they are also both on a
shared wifi LAN, that is a second, worse path — a fallback, not a coin flip.

### What actually happened

The peer's link migrated between the cable and the wifi path on a fixed
cadence, tearing down and re-establishing its session each lap. Because that
peer was also the tree parent, each migration switched parents, invalidated the
downstream coordinate cache and re-announced to every peer.

Nothing was down. Both interfaces stayed present with carrier, `binds: 1`, and
no interface detach event occurred in the whole run — so `show transports`
looked healthy throughout.

### Reproduction

Two nodes, both with an `ethernet` transport on their wifi interface **and** on
a directly-cabled interface, `auto_connect: true` on both (the shipped
default). No further configuration; the flap is the steady state.

### Evidence

Scoped to a single daemon run (transport IDs are reassigned per run, so windows
spanning restarts are not comparable):

```
17 auto-connect dials to one peer in 15 minutes, alternating:
  10  transport:4  02:00:00:00:ca:b1   (cable, en12)
   6  transport:2  02:00:00:00:wi:fi   (wifi,  en0)
   1  transport:4  02:00:00:00:ca:b2
```

```
23:01:07  Peer promoted to active        <peer>
23:02:10  Peer promoted to active        <peer>
23:02:12  Parent switched after first RTT measurement
23:07:06  Parent switched, invalidated downstream coord cache entries,
          announcing to all peers
```

Three parent switches in fifteen minutes. Across hours, promotions appear as
pairs 32-36 s apart. The displaced link reported `etx = 1.0`, `loss = 0.0`.
`show peers` showed the peer's `link_id` climbing (149 → 186) as links were
rebuilt.

---

**Component:** `node.lifecycle` / peer discovery, `peer.active` / link cost.
Contrast with `proto.stp` / parent selection, which already solved the general
problem.

[...]

Issue #125, opened 2026-07-07
Ethernet transport never recovers receive path after interface destroy/recreate


Origami74:
## Symptom

Field-tested during dual-band 802.11s backhaul work (#123): disable + re-enable a mesh interface (`wifi` reload, LuCI toggle — anything that destroys and recreates the netdev). L2 mesh peering comes back, but the FIPS peer never re-appears. Only a daemon restart recovers.

## Diagnosis

`wifi` disable/enable destroys the netdev and recreates it with a new ifindex. The transport's AF_PACKET socket, bound to the old ifindex, is permanently dead. The three consumers of that socket behave differently:

- **Beacon sender** — has ENXIO detection and reopens its socket by interface name after 3 consecutive failures (`beacon_sender_loop`, src/transport/ethernet/mod.rs — "Stale veth detected (ENXIO), attempting socket reopen"). Recovers, but only itself: the reopened socket is local to the beacon task.
- **Receive loop** — no recovery path. Worse than erroring: on a destroyed netdev the fd never becomes readable again, so `ethernet_receive_loop` blocks forever in `recv_from` with no error and no log line. Inbound beacons and handshake msg1 are invisible from then on.
- **Data send path** — still uses the original dead socket held by the transport (`self.socket`); every send fails with ENXIO.

Net effect after an interface bounce: the node transmits discovery beacons (sender recovered) but can neither receive nor send anything else — peers see us, try to handshake, and get silence. Observed log signature (beacon warn only, no recv errors):

```
WARN fips::transport::ethernet: Failed to send beacon transport_id=transport:5 error=send failed: No such device or address (os error 6)
```

This also undermines the dual-band failover story from #123: a standby band whose interface was bounced after daemon start is silently dead as a failover target.

## Fix sketch

Socket recovery must be transport-wide, not per-task:

1. Hold the `AsyncPacketSocket` behind a swappable handle (e.g. `ArcSwap` or `Mutex<Arc<_>>`) shared by send path, receive loop, and beacon sender.
2. On ENXIO detection (beacon sender is the natural watchdog since it transmits periodically), reopen by interface name, swap the shared handle, and respawn/cancel the blocked receive task (it cannot self-detect — it must be aborted and restarted with the new socket).
3. Re-read MAC/MTU on reopen (the recreated interface may differ).
4. Consider attempting reopen on first ENXIO rather than after 3 × beacon_interval (~90 s with defaults).

[...]

PR #147, opened 2026-09-07
feat(transport): bind and rebind network interfaces dynamically

Origami74:
An interface-bound transport is a long-lived object that is *sometimes bound*.
The interface it names may not exist when the daemon starts, may appear minutes
later, may vanish and return mid-operation, and may never appear at all. Until
now the first observation was final: an interface missing at start was logged
once and skipped for the life of the process, and one that disappeared at
runtime published a health change but was never rebound.

On OpenWrt that is a live bug. procd starts fips before wifi has created
`fips-mesh0` / `fips-ap0`, both transports are skipped and never retried, and
the 802.11s peer link forms anyway — that is mac80211, not the daemon — so the
node looks healthy and reaches nothing.

This gives every interface-bound transport presence state and a binder task.
`start_async` returns `Ok` with the transport absent rather than failing; the
binder binds when the interface appears, tears the socket down when it goes
away, and rebinds when it returns. Start-time absence and runtime detach are
one transition and one code path.

## Design decisions worth reviewing

**Presence is `IFF_UP`, not `IFF_UP|IFF_RUNNING`.** Carrier and bindability are
different questions and only the second belongs in a bind gate. Gating on
carrier would report a wifi-only router `Degraded` forever for an empty
`br-lan`, turn every carrier flap the socket would have survived into a rebind
cycle, and deadlock an 802.11s interface that reports `RUNNING` only once
peered — peering needs beacons, beacons need a bound socket. Carrier is
reported beside presence in `show_transports` instead.

**A name is not a device.** Both backends bind by device: `AF_PACKET` stores
`sll_ifindex`, a BPF descriptor follows the interface it was attached to. A
netdev deleted and recreated under the same name leaves the socket attached to
something gone while the name resolves fine, and nothing notices — a stale
`AF_PACKET` socket never becomes readable, so the receive loop neither errors
nor exits. The bound index is captured at bind and compared on every poll. This
is the mechanism for #125.

**`Degraded` is a level, not a latch.** The supervisor's reason set was
monotonic, which was correct while no child could recover. Absence now lives in
its own reversible set and health is recomputed in both directions.

**The node's egress MTU follows the bound set.** `transport_mtu` filters on
`is_bound()`, not `is_operational()`, and the TUN MSS ceiling moved to a shared
atomic since that minimum now changes at runtime.

**`optional: true`** decides how loudly absence is reported. It describes the
interface's presence, not the transport's importance.

## Behaviour changes beyond the binder

- A transport losing its interface now withdraws the peers that needed it,
  instead of leaving routes selectable until the liveness reaper fires.
  Measured before the fix: 27 seconds routing through a dead link with four
  alternatives available.
- A transient send failure during msg2 no longer tears down the half-built link
  or records it as peer misbehaviour. A local interface flap is not the
  remote's fault.
- Discovery no longer re-dials an active peer on an alternate path while the
  link it holds is live. A peer reachable over both a LAN and a cable was
  previously re-dialled every tick, migrating between them on a fixed cadence —
  17 dials in 15 minutes on real hardware, displacing a link reporting
  `etx 1.0` / `loss 0.0`.

## Testing

Unit coverage for the presence machine, churn damping, the supervisor's absence
set, and the detach classification. A docker suite (`testing/iface-binding`)
drives two daemons against a real veth: boot race, late attach, flap,
destroy-and-recreate, an interface present *before* the daemon starts, the
netlink fast path actually being in use, and log hygiene. A chaos scenario
(`ethernet-churn`) carries iperf3 traffic over Ethernet while node churn removes
interfaces underneath it, with a new `min_traffic` assertion — previously
traffic results were written to disk and never read, so a scenario could have
every session fail and still pass.

A privileged CI step runs the interface-binding tests as root, because every
other unit-test leg runs unprivileged and so the entire bind-success path ran
nowhere.

Each regression test was verified against the defect it guards — break the fix,
confirm the test goes red.

## Known limitations

- `"socket died"` has no end-to-end test; the decision is covered, the input is
  not.
- The alternate-path fix stops the churn but does not choose the *better* path —
  a node that lands on wifi stays there while a cable idles. Tracked in #143.
- Three Low-severity security findings are recorded and not addressed: no
  privilege drop, per-transport rather than process-global netlink watchers, and
  `fips-*-setup remove` no longer revoking the config's claim on an interface
  name.

Depends on the link watcher from #144, and reuses it rather than opening a
second netlink socket.

---

*Note: `native::tests::an_arrival_reaches_the_listener_descriptor_…` fails
intermittently under full parallel load. Pre-existing and unrelated to this
branch — green on master and on re-run here (2230 passed).*


jmcorgan:
Your branch rebases onto current master with no conflicts, and I have pushed a trial regrouping of it that you are free to take or ignore.

First the mechanical part, since it is the useful half. All 23 commits replay onto `master` at `813c50bb` cleanly. Master has moved 21 files since your base at `fc5b4ade` and your branch touches 47, and the two sets intersect in only `CHANGELOG.md` and `testing/ci-local.sh`, neither of which collided. On the rebased branch the workspace builds, `cargo fmt --check` is clean, and `cargo test --lib` gives 2406 passed and 0 failed.

The trial is on `trial/pr147-squash`, at `25d7adc2`. It groups your 23 commits into 7 and changes nothing else: `git diff` between the plainly rebased branch and the squashed one is empty, so the trees are byte-identical, and your authorship is intact on all seven. The grouping is:

1. **Dynamic interface binding and the presence machine.** Your original first commit, plus the six-defect wiring fix, the config validation for impossible interface names, the presence unit tests with the CI lines that enable them, the dummy-carrier assertion fix, and the documentation. Ten of yours in one, on the principle that a fix belongs with the thing it fixes.
2. **The iface-binding harness**, with the pre-existing-interface case, the fast-path and churn assertions, and the chaos rebind scenario.
3. **Central classification of send failures**, carrying the peer withdrawal on interface loss and the half-built-link fix. Those two are what the classification exists for, so reading them apart is harder than reading them together.
4. **The re-dial fix**, alone. It is an independent defect with its own measurement, and it answers a separate report.
5. **The fipstop transports view**, with the 80-column fix and the `show_transports` snapshot test.
6. **The OpenWRT packaging change**, untouched, since it has nothing to do with interface binding.
7. **The presence lift out of the ethernet module**, last and by itself.

Two of those positions are forced rather than chosen, and I found both by hitting conflicts rather than by reasoning:

- **The presence lift cannot be folded into commit 1.** It deletes `watcher.rs`, which the commits folded in front of it still modify, so any attempt to absorb it produces a modify-delete conflict. Keeping it last is also the better read: it is a pure move, and a reviewer can confirm that in one pass without separating a move from a change.
- **The harness has to come before the node work**, because your withdraw commit edits `testing/iface-binding/test.sh`, which the harness commit creates. Ordering them the other way fails the same way.

To look at it: `git fetch origin trial/pr147-squash`. If you like it, the simplest path is `git reset --hard origin/trial/pr147-squash` on your own branch and a force push; if you would rather group it differently, that is your call and I am not attached to these seven. The reason for offering it at all is review clarity rather than tidiness: at 23 commits the fixes sit far away from what they fix, and a reader cannot tell which of them are corrections to earlier ones in the same branch.

Nothing about the pull request itself has changed, and it is still a draft. The trial branch is separate and I will delete it whenever you say, or leave it as a reference.

jmcorgan:
This is on master now and carried up to next, so I am closing it by hand: the commits were rebased, so GitHub cannot match them and will not mark it merged. Thank you for this, it is a large piece of work and it lands well.

Since you had said you were done with it and were away for a few days, I rebased the branch onto current master and worked through it commit by commit, folding the remaining review items in here rather than sending them back to you. Your seven commits are on master as `abe3f0b4` through `a3545015`, with your authorship on all of them.

What went in on top of your work: the four items from the branch review, a commit-ordering fix so each commit passes its own checks rather than only the tip, changelog entries for the three commits that had none, and the fipstop transports reference page, which the branch changes and which the review had missed on its first pass.

Two of those are worth naming. The stale pending presence edge is now fixed in both directions: `publish_presence` returns the pending slot instead of a bool, so an edge that lands supersedes one the channel refused in the same statement, and there is a test for it. The `debug_assert` on the promoted tail is widened rather than commented around, because a transient msg2 send failure deliberately leaves the leg registered, and three comments that the deferral had falsified are corrected with it.

One behaviour change goes with it and is written up in the changelog, since it is not obvious from the diff. A peer held on an adopted NAT-traversal transport that is also reachable by beacon used to drift onto the local path on the next discovery tick. It now stays on the traversed path for as long as that path answers. Moving it is still done by the configured-peer refresh, and a traversed link that goes quiet still releases the peer to every path.

Running the full suite on a second runner turned up two test harness problems, both fixed in one commit on top, `d5e4533c`. Neither was in the daemon. First, your `iface-binding` suite built its host veth names from the run suffix, and an interface name only gets fifteen characters, so on a runner that sets `FIPS_CI_NAME_SUFFIX` to a timestamp ip refused the name. GitHub's job leaves that variable unset, which is why it passed there. The suite now hashes the suffix through `sim.naming`, as the chaos simulation and the NAT topology script already do, and `ci-cleanup.sh` knows the new names. Second, the `ethernet-churn` chaos scenario failed to converge on that runner, on master and next alike. That turned out to be the chaos harness: when a stopped node's network namespace outlived the stop, recreating its veth links failed on a name collision, the harness logged success anyway, and the links stayed down. That is fixed, along with a few smaller harness defects found on the way, and the scenario is recalibrated to full convergence.

Carrying it up to next needed your binder rewrite adapted to next's Ethernet framing, which has a 4-byte header and beacons that carry no pubkey, so the binder's MTU arithmetic on next is not the same as on master.

The merged tip is green on GitHub CI, 39 jobs including the full integration matrix and your new `iface-binding` suite. The trial branches I pushed while working on this, `trial/pr147-squash` and `trial/pr147-folded`, are deleted now that everything is on master.

PR #148, opened 2026-09-07
netmon: fingerprint the path to each peer instead of the host

### LINKED ISSUE #145: `interface_addrs()` likely returns nothing on Android API 30+, silently halving the medium-change fingerprint
Follow-up from #142 / PR #144, filed at @jmcorgan's request. Not reproduced on a
device — I have no Android host — so the failure is reasoned from bionic's
implementation and the code, and the first thing this needs is confirmation.

### What the fingerprint is made of

`node::netmon` samples two independent signals and reports a medium change when
either moves:

- `preferred_source()` — `connect(2)` on a UDP socket, resolving the route the
  kernel would use for an off-link destination. Two syscalls, no packets,
  identical on every platform.
- `interface_addrs()` — every up, non-loopback unicast address, via
  `getifaddrs(3)`.

The second exists to catch a medium arriving or leaving *without* displacing the
default route: a LAN cable plugged in beside live WLAN, or the address a
link-local Ethernet transport peers over. The module docs say so at
`src/node/netmon/mod.rs:58-62`.

### The problem

bionic implements `getifaddrs` over netlink `RTM_GETLINK` / `RTM_GETADDR`.
Android 11 (API 30) restricted exactly that enumeration for apps targeting R and
above. So on the platform this detector matters most for, the address half of
the fingerprint is expected to return an empty set — or a partial one limited to
the caller's own interfaces.

`interface_addrs()` returns `BTreeSet::new()` on a `getifaddrs` failure
(`src/node/netmon/mod.rs:571-573` — it returns the empty set it started with). That is the right thing to do and it
degrades gracefully: an empty set is stable, so it contributes nothing to the
fingerprint rather than producing spurious changes, and `preferred_source()`
still catches every default-route move — which is the Wi-Fi ↔ cellular case, the
one Android actually needs.

So this is a **loss of coverage, not an outage**. What is lost is the
non-displacing case: a medium arriving or leaving while the default route stays
put.

### Why it is worth an issue anyway

The degradation is completely silent. Nothing logs that half the fingerprint is
inert, and nothing distinguishes "this host genuinely has one address" from
"the kernel refused to tell us". An operator or an embedder debugging a missed
handover on Android has no way to tell which of the two signals is live.

The `#[cfg(not(unix))]` arm has a doc comment describing precisely this
tolerated degradation for Windows (`:601-604`), but the unix arm can hit the
same state at runtime with nothing said.

### Suggested shape

1. **Confirm on a device.** Whether `getifaddrs` returns `EPERM`, an empty list,
   or a loopback-only list on API 30+ decides the rest. This is the blocking
   step; everything below assumes it degrades.
2. **Distinguish failure from emptiness.** `interface_addrs()` currently
   collapses both into an empty set. Returning `Option<BTreeSet<IpAddr>>`, or
   logging once at `debug` on the `getifaddrs` error path, makes the state
   observable without changing the fingerprint's behaviour.
3. **Say so in the docs.** The platform table in `src/node/netmon/mod.rs:24-28`
   and the operator-facing section in `docs/reference/configuration.md:197-207`
   both now list Android as event-driven, which PR #144 made true for the
   *wake* source. Neither says the *sample* is narrower there.

I do not think the address half should be replaced with a JNI call to
`NetworkInterface.getNetworkInterfaces()`: it has the same restriction behind
it, and the durable Android answer is a `ConnectivityManager` push, which is
its own issue.

### Environment

Reasoned against `master` at `b1a1043` (PR #144), bionic behaviour from the
Android 11 behaviour-change notes. No device, no reproduction.

fr34aky:
The per-peer probe, as owed. Plus four of the six follow-ups.

Three commits: the probe, a fix to a hole the probe's own shape opens, and the
follow-ups.

## The probe

`NetFingerprint` becomes `NodeAddr -> local source address`. For every peer
whose transport address is a numeric IP endpoint, `connect(2)` a UDP socket at
it and read back the address the kernel picked. `interface_addrs()`, its
`getifaddrs` walk and the `sockaddr` decoding are gone, and with them the class
of noise you described: a container bridge, a VPN, a `veth` pair or a tunnel is
not the route to any peer, so it cannot enter the sample. On-link peers become
visible for the first time, and a more specific route moving under one peer is
representable at all.

Verified against a real kernel rather than argued: in a namespace, an interface
coming up with its own subnet moves nothing, and a more specific route to the
peer out of that same interface moves exactly that peer. `interface_addrs()`
has no call site left, so **this absorbs #145** — I spent nothing on it.

## Where I deviated from your instructions, and why

You asked for the intersection of consecutive samples, so peer churn cannot
fire the fan-out on its own. I built that, and on its own it is wrong — not in
a corner, but in a window that opens after every peer that authenticates.

`last` gains a peer only at the first wake *after* it appears. A medium change
inside that window is the detector's first sight of that peer: the intersection
is empty, nothing is reported, and the sample is adopted as the new baseline.
The event is not delayed, it is consumed. The peer's connected socket stays
pinned to the path the host just left and nothing repairs it, so the peering
black-holes to the liveness timeout — the outage this subsystem exists to
prevent. A medium change is itself what wakes the detector, so the two coincide
readily. The host-wide fingerprint did not have this hole: it sampled from
detector startup independently of any peer. Keying on peers introduces it.

Demonstrated as an A/B in the medium-change lab, with `poll_interval_secs`
raised so the move's own netlink event is the detector's first wake:

| tree | events |
|---|---|
| the probe commit alone | **0** — move swallowed entirely |
| with the fix | **1** — `1/1 peers: 85904a3e -> 172.31.61.10 sockets_rebound=1 heartbeated=1` |

My first attempt at that A/B did *not* reproduce — both sides fired, because
container churn let the detector sample the peer first. Worth saying plainly: a
single green suite run proves nothing here, which is how the original 8/8
passed over a live bug.

So a peer seen for the first time is judged against its own send path instead
of against history — the source `connect(2)` pinned its socket to, against what
the routing table would choose for it now. No baseline needed, and it asks the
question the subsystem is actually about rather than inferring it from a
difference. Your condition's purpose holds: a peer joining onto a path that has
not moved is pinned exactly where its traffic goes and reports nothing.

**This nearly went in badly.** `ConnectedPeerSocket::local_addr()` looks like
the right value and is the *wildcard* bind (`0.0.0.0:2121`) — production passes
`udp.local_addr()` into `from_fd`, which is why that accessor was dead code.
Seeding from it would have compared an unspecified address against a real one
for every peer and fired the fan-out on every join: worse than the bug being
fixed, and exactly what your intersection rule exists to prevent. The socket now
reads its pinned source back with `getsockname` once at construction.

Residual, documented at the call site: a first-seen peer with no `bound`. That
is every platform but Linux and macOS, every peer on a stream or proxied
transport on those two, and the one-tick gap before an installed socket becomes
visible. The first two are harmless — with no connected socket there is nothing
pinned to repair, and the cost is the immediate heartbeat only. The third is a
real hole: the tick publishes the snapshot before it installs sockets, so a
socket installed on tick N is first seen on tick N+1, and a peer whose path
moves inside that window does hold a pinned socket. Bounded rather than
permanent, since the ordinary diff recovers it on the next sample.

A non-wildcard `transports.udp.bind_addr` is no longer part of the residual —
see below.

## Your other condition

The peer table is reached through `entities_snapshot`; the detector holds no
node state and takes no lock. Judgment call to flag: `PeerRow` gains two typed
fields (`probe_target`, `bound_source`) rather than having the detector
re-parse the display string next to it. Re-parsing would work — MACs, `.onion`,
Nym and hostnames all fail a `SocketAddr` parse, so it is a correct filter for
free — but it would make a safety mechanism depend on a display string's
formatting and degrade silently to "detect nothing" if that changed. Same
silent-degradation shape as your netlink-mask item, so I paid two fields for
it. Say if you would rather have the re-parse.

## Follow-ups

**2, 3, 4, 5 and 6 close here — five of six, not four. 1 is not here.**

Correcting my own count: item 5 was the two documentation claims that do not
match the code. Both were still wrong at the base and both are fixed by
`879007e` in this pull request, whose message says so without connecting it to
the item.

- **2** — a failed heartbeat no longer resets the timer. The attempt is
  recorded separately and gates retries, deliberately: without that gate a
  failing peer would retry every tick, and the send behind it is item 1's
  unbounded write. Fixing 2 naively makes 1 ten times worse. The constant's doc
  says it can come down once 1 lands.
- **3** — `node.netmon.*` validated. A zero `poll_interval_secs` was silently
  clamped to 1s, so a typo produced a node polling 20x more often than
  configured, silently; now refused, and only while detection is enabled. Plus
  a cross-field check that the worst-case debounce cannot outlast
  `link_dead_timeout_secs`, since then the reaper wins and the detector cannot
  help. `MAX_DEBOUNCE_ROUNDS` is shared with the detector so the two cannot
  drift, and a test asserts the shipped defaults still validate.
- **4** — the group mask is now checked in CI. A wrong mask cannot be caught at
  the bind: subscribing to the wrong groups succeeds exactly like the right ones
  and differs only in what never arrives. So the watcher reads its subscription
  back with `getsockname` (`sockaddr_nl.nl_groups`, no privileges), and the test
  drives `build_wake_source` so it asserts on the production path rather than a
  second copy of the constant. It is a new test and runs in CI; nothing that was
  previously ignored became unignored, and `a_route_change_alone_reaches_the_watcher`
  is still ignored at head. Confirmed against the regression you
  named — reverted to link-events-only it fails `left: 1, right: 1361`.
- **6** — `heartbeated` counts sends that succeeded. A medium change is exactly
  when sends start failing, so the old count read the same whether every frame
  left or none did.

**1 is assessed and deliberately not here.** All three stream transports already
have `close_connection_async`, so bounding `write_all` and evicting on expiry is
~15 lines each, and dropping a partial write is safe if the connection goes with
it — the half-frame objection only holds while the stream stays open. But it
buys less than it looks: a bounded write still stalls the rx loop for the whole
budget, and to sit under `link_dead_timeout_secs` that budget has to be short
enough to threaten a healthy Tor circuit. The real fix is to stop needing
`&mut self` on that path so the send can leave the loop. That is a transport
change with its own tuning judgment across three hot paths, and it does not
belong in a detector rewrite. Happy to take it next.

## Verification

- Churn rules mutation-checked: union iteration fails four tests; dropping the
  first-sight seed fails only the reproduction and leaves the churn tests green,
  so the two rules are independently pinned.
- The `getsockname` wiring has its own test, failing `Some(0.0.0.0)` against
  `Some(127.0.0.1)` if the publish site reverts. Nothing renders that field, so
  nothing else would catch it.
- All 16 static gates pass — `fmt`, both clippy passes, unit tests (2352 + 39 +
  50 + 3 passed, 0 failed), build, CI parity.
- Integration suites run individually rather than as one pass, because the full
  runner exhausts memory on this host: **medium-change**, **static-mesh**,
  **chaos-churn-mixed-10** and **chaos-tcp-mesh**, all green. Those are the ones
  this touches — the medium-change lab for the detector, and the other three for
  the heartbeat-timer change, which affects all peering rather than just netmon.
  In the lab both route-move phases keep the peering intact and the negative
  control still black-holes for 12.2s with detection off. The remaining suites
  are downstream of the same gates and exercise areas this does not touch; the
  one change with global reach is the config validation, and only the
  medium-change suite sets `node.netmon.*` at all — everything else takes the
  defaults, which have a test asserting they validate. GitHub runs the full
  matrix here.
- Three `CAP_NET_ADMIN` tests pass under `unshare -rn`, including a new one for
  an interface going down and returning — the shape the docker lab deliberately
  does not cover, since it keeps both interfaces up throughout.

---

Closes #145. The probe deletes `interface_addrs()` outright, which is that
issue's whole subject, so it closes by deletion rather than by fix — the
observability question it asked becomes unfalsifiable rather than answered. I
have said so on the issue itself rather than leaving the link to imply it was
investigated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   fingerprint, while a peer on the same LAN — reached by its subnet route, not
   the default route — is covered, as is a more specific route moving under a
   single peer. Peers joining and leaving are ignored on their own, being
-  ordinary node behaviour rather than a statement about the medium. A peer
+  ordinary node behaviour rather than a statement about the medium — except
+  that a peer seen for the first time is checked against its own
+  `connect()`-ed socket, and reported if that socket is pinned to a source the
+  routing table would no longer choose, so a medium change in the window
+  between a peer authenticating and the next sample is not adopted silently
+  while that peer sits stranded on the old path. A peer
   addressed by MAC, by `.onion` or Nym recipient, by a scoped IPv6 literal, or
   by a hostname it has not yet been heard from on, has no route to ask about
   and contributes nothing; a node with no peers detects nothing, having nothing
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,13 +47,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Node lifecycle
 
 - Transport-medium change detection, controlled by the new `node.netmon.*`
-  block (on by default). The node samples a coarse fingerprint of its network
-  attachment — the source addresses the routing table would pick for an off-link
-  destination, plus the set of up, non-loopback interface addresses — and
-  reports a change once the picture settles. A handover is not atomic (the old
-  address goes, briefly nothing has a route, the new one arrives), so a short
-  debounce coalesces the burst into one event and a fingerprint that settles
-  back where it started reports nothing. Linux subscribes to `NETLINK_ROUTE`
+  block (on by default). For each peer whose transport address is a numeric IP
+  endpoint, the node asks the kernel which local address it would use to reach
+  *that peer* — a `connect(2)` on a UDP socket, which resolves the route and
+  sends nothing — and reports a change once some peer held across two
+  consecutive samples is reached from a different local address, or has stopped
+  being reachable at all. Asking the question per peer rather than about the
+  host is what keeps it quiet: a container bridge, a VPN, a `veth` pair or a
+  tunnel appearing is not the route to any peer and cannot move the
+  fingerprint, while a peer on the same LAN — reached by its subnet route, not
+  the default route — is covered, as is a more specific route moving under a
+  single peer. Peers joining and leaving are ignored on their own, being
+  ordinary node behaviour rather than a statement about the medium. A peer
+  addressed by MAC, by `.onion` or Nym recipient, by a scoped IPv6 literal, or
+  by a hostname it has not yet been heard from on, has no route to ask about
+  and contributes nothing; a node with no peers detects nothing, having nothing
+  bound to the old path to repair. The peer table is read through the node's
+  existing lock-free entity snapshot, so the detector stays a detached task
+  holding no node state. A handover is not atomic (the route goes, briefly
+  there is none, the new one arrives), so a short debounce coalesces the burst
+  into one event and a fingerprint that settles back where it started reports
+  nothing. Linux subscribes to `NETLINK_ROUTE`
   multicast (the groups `ip monitor` uses) and macOS and FreeBSD to a
   `PF_ROUTE` socket, both reacting to the kernel event in milliseconds; every
   other platform samples on a timer at `node.netmon.poll_interval_secs`, which
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+#### Node lifecycle
+
+- A heartbeat whose send failed no longer counts as one that was delivered.
+  The peer's "last heartbeat" timestamp was stamped before the send and left
+  alone whatever came back, so a failure suppressed the next attempt for a
+  full `node.heartbeat_interval_secs` even though the peer had heard nothing —
+  on a 10s interval against a 30s `link_dead_timeout_secs`, three failures in
+  a row were the whole budget. The timestamp now moves only on a send that
+  returned cleanly, and a separate record of the *attempt* spaces the retries
+  so a peer that keeps failing is retried in seconds rather than either
+  hammered every tick or left for a full interval. That retry spacing
+  applies to the failure path only: gating a healthy peer on it as well
+  would have floored `node.heartbeat_interval_secs` at two seconds, so a
+  configured value below that would silently not have been honoured.
+
+- A peer that rotates its address no longer keeps sending from a socket
+  aimed where it used to be. The authenticated-frame path updated the
+  peer's address and discarded the flag saying it had changed, so the
+  per-peer `connect()`-ed UDP socket stayed pinned to the old 5-tuple;
+  the sibling path already cleared it.
+
 #### Data plane
 
 - A per-peer `connect()`-ed UDP socket is no longer left pinned to an interface
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -250,6 +250,14 @@ on the same LAN, reached by its subnet route rather than the default route, is
 covered as well as one across the internet, and so is a more specific route
 moving under a single peer.
 
+The converse is the residual. The probe answers for the peer's *current*
+address, and that address is the source of the last authentic packet it sent,
+so a peer that roams between two of its own addresses which leave this host by
+different interfaces is indistinguishable from a local path move. The reaction
+is scoped to the peers named in the change, so such a peer moves nothing but
+its own send path — but it is the peer, not this host, that decided the
+fingerprint changed.
+
 Peers appearing and leaving are ignored on their own — that is ordinary node
 behaviour and says nothing about the medium. A peer seen for the first time is
 the one exception, and it is not judged against history but against its own
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -202,7 +202,7 @@ an interface arriving or leaving — and rebinds the send path immediately.
 | Parameter | Type | Default | Description |
 |-----------|------|---------|-------------|
 | `node.netmon.enabled` | bool | `true` | Whether medium-change detection runs |
-| `node.netmon.poll_interval_secs` | u64 | `5` | How often the host's network attachment is sampled (backstop period where an event-driven backend exists) |
+| `node.netmon.poll_interval_secs` | u64 | `5` | How often the path to each peer is sampled (backstop period where an event-driven backend exists) |
 | `node.netmon.debounce_ms` | u64 | `250` | How long to wait for the picture to settle before acting (`0` disables) |
 
 Established UDP peers use a per-peer `connect()`-ed socket for the send fast
@@ -219,9 +219,34 @@ connected socket is reinstalled on a later tick) and heartbeats every peer on a
 connectionless transport at once so the far side re-pins to the new source
 address. A peer reached over TCP, Tor, Nym or BLE is left to its periodic
 heartbeat, since sending to it here would block the node's receive loop on a
-stream the medium change has very likely just stranded; those transports
-re-dial on send. No peering is torn down: sessions, tree positions and routes
-survive the switch.
+stream the medium change has very likely just stranded. No peering is torn
+down: sessions, tree positions and routes survive the switch.
+
+**What counts as a change.** For each peer whose transport address is a numeric
+IP endpoint, the node asks the kernel which local address it would use to reach
+*that peer* — a `connect(2)` on a UDP socket, which resolves the route and sends
+nothing. A change is reported when a peer present in two consecutive samples is
+now reached from a different local address, or has stopped being reachable at
+all.
+
+Because the question is asked per peer, an interface the node does not peer over
+cannot trigger anything: a container bridge, a VPN, a `veth` pair or a tunnel
+appearing is not the route to any peer, so it does not enter the sample. A peer
+on the same LAN, reached by its subnet route rather than the default route, is
+covered as well as one across the internet, and so is a more specific route
+moving under a single peer.
+
+Peers appearing and leaving are ignored on their own — that is ordinary node
+behaviour and says nothing about the medium. A peer whose address is not a
+probeable IP endpoint contributes nothing: a MAC on Ethernet or BLE, a `.onion`
+or Nym recipient reached through a local proxy, an IPv6 literal with a scope
+suffix, or a peer still carrying the hostname it was configured with (resolving
+one would put a DNS lookup on the sample path; the address becomes numeric as
+soon as an authenticated packet arrives from the peer). A node holding no peers
+detects nothing, which is correct — it has nothing bound to the old path.
+
+The cost is three syscalls per peer per sample, bounded by
+`node.limits.max_peers`, with no packets sent and no name resolution.
 
 Detection uses the best backend the platform has:
 
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -202,8 +202,22 @@ an interface arriving or leaving — and rebinds the send path immediately.
 | Parameter | Type | Default | Description |
 |-----------|------|---------|-------------|
 | `node.netmon.enabled` | bool | `true` | Whether medium-change detection runs |
-| `node.netmon.poll_interval_secs` | u64 | `5` | How often the path to each peer is sampled (backstop period where an event-driven backend exists) |
-| `node.netmon.debounce_ms` | u64 | `250` | How long to wait for the picture to settle before acting (`0` disables) |
+| `node.netmon.poll_interval_secs` | u64 | `5` | How often the path to each peer is sampled (backstop period where an event-driven backend exists). **Must be at least 1 while `enabled`; `0` is refused at startup.** |
+| `node.netmon.debounce_ms` | u64 | `250` | How long to wait for the picture to settle before acting (`0` disables). **Refused at startup when `debounce_ms × 8` reaches `node.link_dead_timeout_secs`** — see below. |
+
+Both refusals stop the node rather than degrade it, so they are worth knowing
+before they are met.
+
+A handover is ridden out for up to **8** settling rounds (`MAX_DEBOUNCE_ROUNDS`)
+of `debounce_ms` each before a change is reported. If that worst case reaches
+`node.link_dead_timeout_secs`, the liveness reaper tears the peering down before
+the detector ever reports, so the machinery runs and cannot help — the node
+refuses to start rather than run in that shape. At the shipped defaults the
+margin is wide (8 × 250 ms = 2 s against 30 s), but the constraint couples two
+keys in different blocks: **shortening `link_dead_timeout_secs` for fast
+failover can make an untouched `debounce_ms` illegal.** The refusal names both
+values and the multiplier.
+
 
 Established UDP peers use a per-peer `connect()`-ed socket for the send fast
 path. `connect(2)` makes the kernel resolve the route once and pin the local
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -237,7 +237,15 @@ covered as well as one across the internet, and so is a more specific route
 moving under a single peer.
 
 Peers appearing and leaving are ignored on their own — that is ordinary node
-behaviour and says nothing about the medium. A peer whose address is not a
+behaviour and says nothing about the medium. A peer seen for the first time is
+the one exception, and it is not judged against history but against its own
+send path: if its `connect()`-ed socket is pinned to a source the routing table
+would no longer choose, it is reported. Without that, a medium change in the
+window between a peer authenticating and the detector's next sample would be
+the detector's first sight of that peer, and would be adopted silently while
+the peer's socket stayed pinned to the path the host had just left. A peer
+joining onto a path that has not moved has its socket pinned exactly where its
+traffic goes, so it still reports nothing. A peer whose address is not a
 probeable IP endpoint contributes nothing: a MAC on Ethernet or BLE, a `.onion`
 or Nym recipient reached through a local proxy, an IPv6 literal with a scope
 suffix, or a peer still carrying the hostname it was configured with (resolving
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -1100,6 +1100,44 @@ impl Config {
             }
         }
 
+        // Medium-change detection. Both checks are about the detector being
+        // able to do its job at all, not about taste in numbers.
+        let netmon = &self.node.netmon;
+        if netmon.enabled {
+            if netmon.poll_interval_secs == 0 {
+                return Err(ConfigError::Validation(
+                    "`node.netmon.poll_interval_secs` must be at least 1; it is the backstop \
+                     period behind the kernel event source, and the only detection signal at \
+                     all on a platform without one"
+                        .to_string(),
+                ));
+            }
+            // A handover is ridden out for up to `MAX_DEBOUNCE_ROUNDS` rounds
+            // of `debounce_ms` before the change is reported. If that can
+            // outlast the liveness timeout, the reaper tears the peering down
+            // first and the detector never gets to rebind anything — the
+            // machinery runs and cannot help.
+            // Saturating: both operands are operator-supplied `u64`s, and an
+            // overflow here would wrap to a small number and silently accept
+            // the very configuration this refuses.
+            let worst_case_debounce_ms = netmon
+                .debounce_ms
+                .saturating_mul(u64::from(crate::node::netmon::MAX_DEBOUNCE_ROUNDS));
+            let dead_timeout_ms = self.node.link_dead_timeout_secs.saturating_mul(1000);
+            if dead_timeout_ms > 0 && worst_case_debounce_ms >= dead_timeout_ms {
+                return Err(ConfigError::Validation(format!(
+                    "`node.netmon.debounce_ms` = {} can hold a report for up to {}ms across \
+                     {} settling rounds, which meets or exceeds \
+                     `node.link_dead_timeout_secs` = {}s: the peering would be reaped \
+                     before the medium change was ever acted on",
+                    netmon.debounce_ms,
+                    worst_case_debounce_ms,
+                    crate::node::netmon::MAX_DEBOUNCE_ROUNDS,
+                    self.node.link_dead_timeout_secs,
+                )));
+            }
+        }
+
         let native = &self.node.native_api;
         // Both floors refuse a node that would start, answer every setup call
         // and then drop every datagram a peer sent. A zero `backlog` makes the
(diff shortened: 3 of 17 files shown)


jmcorgan:
Thank you for this. It lands the probe that pull request 144 recorded as owed, it meets both entry conditions that acceptance imposed, and the design is right. I am going to ask for one substantial round before I merge it, and I want to be straight about why, because it is a change from what you were told on 144.

That acceptance said the probe could land with follow-ups. Reading the finished work, the follow-up list came out longer and wider than a follow-up list should carry: it includes an untested predicate that reaches every peering rather than just netmon, two wiring links whose failure silently restores the bug one of your own commits exists to fix, and a documented residual whose stated justification is wrong. I would rather have all of that in the commit that introduces the change than spread across later ones. That is a judgment about the shape of the change, not about the quality of your work.

**Your central finding is correct, and I checked it rather than taking it on trust.** Two readers here re-derived your window argument separately and reached the same place. `run_detector` seeds the baseline from a snapshot that is empty until the first tick publishes it, so on a host whose first wake is the move itself every peer is first-seen, the intersection is empty, and the post-change sample is adopted as the new baseline. The event is consumed rather than delayed, and nothing inside the intersection rule recovers it. The rule I gave you was written for the host-wide fingerprint, which sampled from detector startup and therefore always had a baseline; keying on peers is what removes that. Your deviation is warranted and the instruction is the thing that was wrong.

Your replacement also preserves the property the instruction existed to provide. Under the default wildcard bind, the pinned source and the current preferred source are the same kernel route lookup for the same destination, so a first-sight report happens if and only if the route changed between `connect(2)` and the sample. A peer joining onto an unmoved path stays silent, and your test pins that.

**Two things about the submission are worth saying plainly.** You reported a defect in the instruction you were given, and you documented a near miss against yourself: seeding from `ConnectedPeerSocket::local_addr()` would have compared the wildcard address against a real one and fired the node-wide reaction on every join. You also said that your first attempt to demonstrate the window did not reproduce, and why. That is the right way to report a timing bug, and it is the reason the rest of this comment reads as detail rather than as doubt.

**On your explicit question, keep the typed `PeerRow` fields.** Your own reason holds, and a simpler one settles it: `bound_source` cannot be re-parsed from anything, because `transport_addr` does not encode it. Re-parsing would save one field and buy a silent degradation path, since an empty target set reads as "nothing moved".

## What I am asking for before merge

Grouped by kind. Take them in whatever order suits you, and push as many commits as you like.

### 1. The one that changes behaviour

Narrow `handle_net_change` to act on `change.summary.moved` rather than the whole peer table. This is the item already deferred at `src/node/handlers/netmon.rs:56-70`, and this pull request is what earns it priority: keying the sample on peers introduces remote influence over the trigger for the first time, since `probe_target` comes from the observed source of every authentic packet with no throttle. On a multi-homed node, one peer alternating between two addresses that resolve to different local sources can drive the node-wide reaction, bounded by the poll interval. The narrowing closes that by construction, because the only peer in the set is then the roamer whose socket `encrypted.rs:373-378` already cleared.

To size it honestly: nothing black-holes today. The sockets reinstall on the next receive-loop tick and sends continue through the wildcard socket meanwhile, so the cost is internal teardown and respawn work, up to 128 drain threads per poll interval, not an outage. It is worth doing here because the lever is new with this change, not because it is urgent.

### 2. The defect the heartbeat change introduced

The retry floor clamps a configured heartbeat interval below two seconds. On a healthy peer `last_heartbeat_attempt` equals `last_heartbeat_sent`, so `retry_due` gates the healthy path too, and the effective interval becomes the larger of the configured value and `HEARTBEAT_RETRY_INTERVAL`. `heartbeat_interval_secs` has no validation floor, and `src/node/tests/tcp.rs:180,230` already set it to 1 against a 3 second dead timeout. Either consult `retry_due` only when `last_heartbeat_attempt > last_heartbeat_sent`, or validate the floor the way `poll_interval_secs` now is. Your choice; both are defensible.

The same defect is present unchanged on the maintenance branch. That one is mine to author separately, because changes flow one way between the branches, so please do not try to cover it here.

### 3. Tests

Each of these is green today with the thing it should be testing removed or broken, which is what puts them in this list rather than in a follow-up.

- **The two-gate heartbeat predicate at `src/node/handlers/mmp.rs:497-509` has no test anywhere.** Deleting `retry_due`, flipping the `&&` to `||`, or raising the constant to minutes all stay green. This is the one change in the pull request that reaches beyond netmon into all peering. Drive `check_link_heartbeats` on a peer whose send fails at a controlled clock: assert no second attempt inside the retry interval and one after it.
- **The two `bound` wiring links are asserted by nothing**, at `src/node/netmon/mod.rs:619` and `:231`. `src/node/tests/netmon.rs:333-336` maps `probe_targets()` straight to `.dest`, the live-sampling helper hard-codes `bound: None`, and the unit tests use `for_test_bound` and bypass both. Substituting `None` at :619 leaves the suite green and silently restores the bug `9027a18c` exists to fix, which is the same failure class you caught for `local_addr()`, one layer down. `a_peers_connected_socket_publishes_the_source_it_was_pinned_to` already loads the snapshot, so an assertion after :394 that `probe_targets(&snapshot)` yields a target for `addr_1` with `.bound == Some(Ipv4Addr::LOCALHOST)` covers the first link, and one assertion that `sample()` against an empty fingerprint reports a move covers the second.
- **The live-probe tests all pass if `preferred_source` returns `None` for every peer.** `sampling_the_live_host_is_self_consistent` is satisfied by two all-`None` samples, `every_target_is_recorded` never checks values, and `a_probe_never_yields_a_loopback_or_unspecified_source` skips through its `if let`. One test that `sample()` on `127.0.0.1:9` yields `Some(Some(127.0.0.1))` fixes this and is true on any host with loopback, including `--network none`.
- **The group-mask test returns green when the netlink bind is refused**, at `src/node/netmon/tests.rs:1070-1078`, which is exactly the environment difference between a real check and a vacuous one. Assert `is_event_driven()` first and fail where netlink is expected, or emit a line on the early-return arm so the run output says which branch was taken.
- **Two smaller ones.** The pinned-source publish test is Linux-gated though the field publishes on macOS too, where `open_connected_fd` is available; widen both attributes. And `reports_are_spaced_out_under_clean_flapping` cannot distinguish the pacing floor from the poll period, since both are one second, so deleting the `last_emit` block stays green; build the wake at 100 ms and the assertion reds.

### 4. The two residuals on the first-sight rule

Both are consequences of your new rule, so I want them settled here rather than discovered later.

- **A non-wildcard `transports.udp.bind_addr` breaks the equality your argument rests on.** `open_connected_fd` binds `udp.local_addr()` verbatim before connecting, at `src/transport/udp/io/connected/fd.rs:80-104`, so a socket bound to a specific address keeps it regardless of route, while `preferred_source` binds the unspecified address and takes the kernel's choice. When the two differ, every first-seen peer reports a move, permanently, with nothing having moved. That is a supported posture: `src/config/mod.rs:2127` carries an example and `docs/how-to/enable-nostr-discovery.md:143` names it. I reached this by reading the two bind sites, not by running a host in that configuration, so tell me if I have it wrong. The fix I would take is carrying the transport bind address on `ProbeTarget` and binding the probe to it, address only and port 0, so both sides answer the same constrained question. Skipping the first-sight rule when the bind address is not unspecified is acceptable if it is documented next to the existing residual.
- **The documented residual is one tick wider than the text says, and its stated reason does not hold there.** `record_stats_history` publishes the snapshot at `src/node/dataplane/rx_loop.rs:486` while `activate_connected_udp_sessions` installs the socket at :510, so the socket exists from step 510 of tick N but the first snapshot carrying its pinned source is tick N+1's. In that window there is a pinned socket to strand, so "with no socket there is nothing pinned to repair" does not cover it. About one second per join against a five second poll, so the hole is small; I am asking because the justification is wrong, not because the size worries me. Amending the sentence at `src/node/netmon/mod.rs:312-317` is enough.

### 5. Documentation and body

- `docs/reference/configuration.md:202-206` documents neither new startup refusal, including the `MAX_DEBOUNCE_ROUNDS` value its own error message multiplies by. An operator shortening `link_dead_timeout_secs` for fast failover gets a hard refusal citing a netmon key they never set, with no reference page predicting it. Add the constraint to both rows.
- The sans-IO contract documents at `src/proto/mmp/core.rs:42-43` and `:172-173` were invalidated by the heartbeat change: the first still says the shell marks the heartbeat sent before the send, the second still omits the retry gate. In this split those comments are the only written statement of the shell's obligation, so a later edit made against them reinstates exactly what you fixed.
- The cost note at `src/node/netmon/mod.rs:216-221` says three syscalls per target. `UdpSocket::bind` is a socket and a bind, and the socket closes on drop, so it is five, which puts a debounced change back at roughly the previous review's estimate rather than below it. In the same place, "bounded by `node.limits.max_peers`" does not hold when that value is 0, which the same document defines as unlimited.
- The "what counts as a change" block claims that asking per peer is what keeps it quiet, without the residual that a roaming peer whose two destinations leave by different interfaces is indistinguishable from a local path move.
- The body's residual sentence is narrower than the code. `bound_source` comes only from `peer.connected_udp()`, so the skip is also every peer on a stream or proxied transport, on Linux and macOS included; the doc comment at `src/control/snapshot.rs:646-656` states this correctly. Suggested wording: "every platform but Linux and macOS, every peer on a stream or proxied transport, and the sub-tick gap before installation."
- Nothing was un-ignored: `a_route_change_alone_reaches_the_watcher` is still ignored at head. What you added is a new mask test that was never ignored. Please correct that line.
- Correct the follow-up tally. It says five landed in the previous pull request. Item 5 was the two documentation claims that do not match the code, both still wrong at the base and both fixed by `879007e7` here, whose message says so without connecting it to item 5. Five of six close here, not four.
- Add `Closes #145`. This deletes `interface_addrs()` outright, which is that issue's whole subject, and the prose "this absorbs #145" does not close it. Because it closes by deletion rather than by fix, so the Android question becomes unfalsifiable rather than answered, please pair it with a one-line comment on #145 saying the observability request died with its call site.

### 6. Small code items

- The private `short()` at `src/node/netmon/mod.rs:407-414` duplicates `NodeAddr::short_hex()` and renders without the suffix every other operator surface uses, so an operator grepping for the peer id that `show_peers` printed will not match the netmon lines. Delete it and call `m.peer.short_hex()`.
- At `src/transport/watcher.rs:243-264`, `subscribed_groups` was inserted below the existing doc line for `is_event_driven`, so that sentence now heads the new function and `is_event_driven` has no documentation. Move the line down.
- `PeerSourceMove::before` is set on every move and wildcarded away by its only non-test consumer at `src/node/netmon/mod.rs:391-397`, so the log says where a peer moved to but not from, which is the address the stale `connect(2)` had pinned and the thing an operator correlates against route history. Render it or drop the field.
- Make both multiplications at `src/config/mod.rs:1120-1122` `saturating_mul`, matching the `saturating_add` at :1321, and fix the 22 literal spaces mid-sentence in the debounce refusal message at :1124-1131. They come from a continuation join, rustfmt does not touch literal contents, and the test asserts substrings only, so nothing catches it. That message is the whole diagnostic for the only refusal an operator reaches by editing `node.netmon.debounce_ms`.
- `src/node/dataplane/encrypted.rs:278` updates the peer address and discards the changed flag without clearing the connected socket, where the path at :361-379 does clear it. That line is pre-existing and you did not touch it, but your first-sight rule now depends on the invariant it breaks. Capture the return the way :377 does, or state the assumption in `moved()`'s documentation. Either is fine.

### 7. Commit shape

Since there is a round anyway, please split the third commit. It bundles four independent fixes across disjoint files, and its message already reads as four commit bodies. I will merge without squashing once it lands, so the history is what a later bisect sees.

Ask me about any of these if you disagree, particularly the non-wildcard bind case, which I derived by reading rather than by running.

fr34aky:
Thank you for this — particularly for re-deriving the window argument
independently rather than taking it from me, and for saying plainly which
instruction was wrong. That is a more useful review than an approval would have
been, and the point about the follow-up list being wider than a follow-up list
should carry is fair. Everything below is pushed.

The branch is now twelve commits. The third is split four ways as you asked, and
the round itself is six more, each independently correct as it lands. Every
commit builds with `--all-targets`, so a bisect will not land on a broken tree.

## 1. Narrowing the reaction — done, `6063a1b`

Your reason for prioritising it is one I had missed: `probe_target` is the
observed source of every authentic packet, updated with no throttle, so keying
the sample on peers put the trigger within reach of a remote party. The
narrowing closes that by construction.

A test holds two peers, moves one, and asserts the other keeps both its socket
and its heartbeat timestamp. The medium-change lab still passes 17/17, which is
the end-to-end check that the peer whose route really did move is still in the
set and still repaired.

## 2. The retry floor — confirmed and fixed in the commit that introduced it, `4153f75`

You are right, and it is worse than it first reads: `mark_heartbeat_sent` stamps
both timestamps with the same instant, so `retry_due` gated the healthy path
outright. `tests/tcp.rs:180,230` do configure 1s against a 3s dead timeout, and
there is no validation floor, so that margin would have halved silently.

I took your first option — consult the retry gate only when the last attempt is
strictly newer than the last success. That fixes the logic rather than
constraining a documented knob, and it leaves the failure-path spacing intact,
which still matters while the unbounded write behind it is unfixed.

Folded into the commit that introduced the defect rather than added after it, so
the defect never appears in the history.

I have not touched `maint`.

## 3. Tests — all five done, and each verified against the mutation you named

- **The two-gate predicate**, `4153f75`. Drives the real sweep, staging elapsed
  time on the peer because `check_link_heartbeats` reads `std::time::Instant`,
  which tokio's paused clock does not move — that cost me a first attempt.
  Deleting `retry_due`, flipping the `&&` to `||`, and raising the constant to
  minutes each fail it; I ran all three.
- **The two `bound` wiring links**, `a421db1`. Both asserted, and both
  mutations you predicted now fail. Your suggested placement worked, with one
  addition: the harness peers sit on a synthetic `loopback:1` address that is
  correctly not probeable, so the peer has to be re-pinned to a numeric
  endpoint first or `probe_targets` filters it out and the assertion cannot see
  the link at all.
- **The vacuous live-probe tests**, `a421db1`. A probe to loopback must now
  answer with loopback.
- **The group-mask early return**, `5f4d563`. Asserted rather than logged, on
  your own precedent — `the_egress_path_mask_opens_a_source` already expects the
  bind to work on Linux.
- **The two smaller ones**, `a421db1`. Both cfg attributes widened to macOS,
  along with the two sibling tests on the same helper. The pacing test now wakes
  at 100ms; at one second against a one-second floor, deleting the pacing block
  still passed.

## 4. The two residuals

**The non-wildcard bind, `dd25481` — you have it right, and it is now measured
rather than read.** `open_connected_fd` binds `local_addr` verbatim before
connecting, so the socket keeps a configured address whatever the route says,
while the probe took the kernel's choice. The probe now binds what the transport
binds, address only and port 0, so both sides answer the same constrained
question; under the default wildcard bind nothing changes.

Rather than leave it derived, I put it on the scale. The namespace test already
builds two addresses on two interfaces, so it now asserts that an unconstrained
probe answers with the carrier and a constrained one with the other address, and
that the two differ — the disagreement that would otherwise report a phantom
move on every first-seen peer. Ignoring the constraint in `preferred_source`
fails it.

**The residual width, `a421db1`.** Confirmed against the tick body:
`record_stats_history` runs at rx_loop.rs:486 and
`activate_connected_udp_sessions` at :510, so the window is a full tick and
there is a pinned socket in it. The text now separates the harmless groups from
that hole instead of covering both with one wrong reason.

## 5. Documentation

Both refusals are on the reference page with the multiplier the error message
uses, and with the coupling called out — shortening `link_dead_timeout_secs` can
make an untouched `debounce_ms` illegal. The sans-IO contract documents move
with the change they describe. The cost note says five syscalls and no longer
claims a bound that `max_peers = 0` does not provide. The "what counts as a
change" block now carries the roaming-peer residual.

The body is corrected on all four points: the residual sentence, the
un-ignored claim, the tally — five of six close here, and I checked that both
documentation claims were still wrong at the base — and `Closes #145`, paired
with a note on the issue saying it closes by deletion and leaves its question
unfalsifiable rather than answered.

## 6. Code items

`short()` deleted for `short_hex()`; the `is_event_driven` doc line moved back;
`before` rendered at both ends; both multiplications saturating and the
twenty-two literal spaces gone, with a test asserting the message carries no run
of spaces, since substring assertions cannot see how it reads.

On `encrypted.rs:278` I took the fix rather than the documented assumption,
`eb73216`. You are right that my rule depends on the invariant, but the stale
socket is a defect on its own terms — it is aimed at an address the peer has
left, and the sibling path already clears it for that reason. Happy to reduce it
to a comment if you would rather that change did not ride along here.

## Verification

`cargo fmt --check`, both clippy passes, `cargo test` (2356 + 39 + 50 + 3
passed, 0 failed). Three `CAP_NET_ADMIN` tests pass under `unshare -rn`. Each of
the twelve commits passes `cargo check --all-targets`.

Integration suites are run one at a time here rather than as a single pass,
because the full runner exhausts memory on this host: the medium-change lab is
17/17 with the narrowed reaction, which is the one this change can break.

jmcorgan:
Merged. This is on `master` and has been carried up to `next`.

Thank you for the round. It was a large ask, you did all seven groups, and the work came back better than it went out: the commit split, the mutation checks on each new test, and the measured bind asymmetry are all things I would rather have than the follow-up list I originally offered you on 144.

**How it landed, and why the pull request shows as closed rather than merged.** Your twelve commits are regrouped into five, with the tree byte-identical to your head and your authorship on every one of them. The grouping is: the probe itself with its tests, the configuration validation, the heartbeat accounting fix, the connected-socket clear on address rotation, and the scoped reaction. The two pre-existing defects you found are deliberately their own commits, so each can be cherry-picked on its own; the heartbeat one has already been carried to the maintenance line separately. Because those commits carry different identifiers from yours, GitHub cannot see the merge, so this pull request is closed by hand. That is bookkeeping, not a rejection. `git log --author` on `master` will show your name on all five.

**What I changed after your push, rather than sending it back to you.** Four things, and the first is the one that matters.

The narrowing I asked for in group 1 introduced a defect, in the mechanism the change exists to protect. `run_detector` advanced its baseline before offering the change to a channel that holds one slot, and the full arm discards the newer change with only a debug line. That was sound while the reaction was node-wide, and the doc on the spawning function said so in as many words: any later change would rebind everything anyway. Once the handler reads only the peers a change names, a discarded change loses that peer set for good, because the baseline has already moved and the ordinary diff never names those peers again. Nothing else repairs them: session activation early-returns while a connected socket is present, and the send path prefers that socket, so the peer transmits from the abandoned source until the liveness timeout reaps the peering. The baseline now advances only on a send the channel accepted. The coalescing test could not see any of this, because all three of its scripted fingerprints named the same peer; it is rescripted over disjoint peer sets and now fails on the old ordering.

Two tests passed with the thing they check removed. The connection-oriented exclusion test captured the sent timestamp and asserted it had not moved, which your own split into attempt and success made vacuous: delete the filter and the stream peer is selected, its send fails at the readiness gate, and the sent timestamp sits still either way. It asserts on the attempt stamp now. Separately, the probe bind address reached the sampler through three sites with no test on any of them, so nulling either end left the suite green; there is a test for it now that needs no privileges. Both were checked by making the mutation and watching the test go red.

The remaining documentation still described the node-wide reaction in six places, including one sentence written by this branch that was false in the branch that wrote it, and an intra-doc link to the function name the narrowing renamed away. Nothing catches that last one, since there is no rustdoc gate.

Finally, the changelog gained a line about the new startup refusal. It is the one operator-visible consequence of this change on an existing deployment: with detection on by default, a node that shortened `node.link_dead_timeout_secs` to one or two seconds for fast failover is refused at startup after the upgrade, citing a `node.netmon.*` key its operator never set. A timeout of zero is exempt.

**On the heartbeat fix.** You and I had written the same fix on two lines. Yours took the first of the two options offered, so the two agreed on behaviour, and reconciling them was a text merge rather than a decision. The merge kept the shape from the maintenance line, which lifts the decision into a pure function so it can be tested without driving a send, and kept both of your integration tests, which exercise whichever implementation survives. One thing that merge did not announce: the test file merged with no conflict marker while both sides had added the same helper in different places, which does not compile. Worth knowing if you meet the same shape.

The remaining items from the review are small and are mine to carry, not yours. Your pull request body is left as you wrote it.

This also closes #145, by deleting its subject rather than by answering it, exactly as your note there says.

PR #150, opened 2026-09-07
transport: take the stream write off the caller's task

fr34aky:
Follow-up 1 from the medium-change review: the unbounded stream write is still
reachable from the tick's own heartbeat sweep. Taking it out of the handler did
not take it out of the path behind the handler — so this takes it out of the
path.

TCP, Tor, Nym and BLE all wrote to their links from whatever task called
`send`. A write to a stream blocks once the peer's receive window and this
node's send buffer are both full, and nothing bounds how long that lasts. A
peer that has stopped reading, or one whose path has just changed medium,
produces exactly that state. The callers are the rx loop's tick handlers, so
while the loop sits in that write it serves nothing else: every other peer's
liveness, the forwarding path, and the control socket all queue behind one
unresponsive peer.

## Not a timeout

I sketched a bounded write with eviction in #148 and it is the wrong answer.
A bounded write still stalls the rx loop for the whole budget — it converts an
unbounded stall into a long one — and to sit under `link_dead_timeout_secs` the
budget has to be short enough to threaten a healthy Tor circuit, which
legitimately stalls for seconds. There is no value that is both.

Each connection now owns a writer task holding the write side, and `send`
enqueues onto a bounded channel. The only code that can await the wire is a
task with nothing else to do, so the property holds by construction and there
is no number to tune.

`try_send`, not `send`: awaiting a full queue would reinstate the same block one
level up. A full queue means the writer has not drained a frame in the time it
took to offer 64, which is a peer that is not receiving, so the send fails and
the caller's existing retry and liveness handling takes over — the same shape as
the connect gate already above it, which refuses rather than waits.

Frames are written whole by a single task, so ordering is preserved and the
half-written-frame hazard cannot arise: a write error takes the connection down
with it, and the peer sees a closed connection rather than a frame it cannot
resynchronise from. That was the objection to a timeout in the first place, and
it only ever applied while the stream stayed open. Teardown mirrors each
transport's existing receive-loop contract exactly — the pool entry is removed
and the direction counter decremented only when the removal returned `Some`, so
a concurrent close or stop of the same address cannot double-count.

Tor and Nym share `socks5::pool`, so the writer loop is written once there and
once for TCP.

## BLE was the worst of the four

`send_async` awaited the L2CAP write **while holding the connection-pool
mutex**. One unresponsive peer therefore froze every other BLE operation as
well as the caller — connects, evictions, and each receive loop's teardown all
queue behind that guard. The other three only blocked the caller.

Same fix. The pooled stream was already an `Arc` and `BleStream` is already
`Send + Sync` with a `Send` future, so the writer task holds a clone and no
generic surgery was needed. Queue depth 16 rather than 64: a BLE link carries a
fraction of the throughput, so the same depth would be seconds of backlog
rather than a burst.

This also settles the Android backend **without changing it**. Its
`BleStream::send` pushes onto the embedder's queue and waits for a slot rather
than dropping, which was an unbounded wait on whatever task called it. That call
now happens only in the writer task, where waiting is the job, and the layer
above it is the connection's own bounded queue, which fills and refuses without
waiting on anything. The comment there now records that its safety depends on
that caller, since putting it back on a caller's task would restore the stall.

## Two deliberate costs

The byte count `send` returns is now what was *queued* rather than what reached
the wire. That is the prediction the UDP fast path already reports when it
dispatches to the encrypt workers; bytes actually written are recorded by the
writer task as they go.

The frame is copied into the queue, because `write_all` borrows and a queue must
own — one memcpy of at most an MTU, against an unbounded stall.

## Verification

Both regression tests drive a peer that accepts and then never reads, push far
more than any buffer holds, and assert every send returns promptly. Both are
mutation-checked: replacing `try_send` with an awaiting `send` fails the TCP one
on the timeout, and reinstating the write under the pool guard fails the BLE
one. The BLE test also asserts the pool stays lockable throughout, which is the
half that was unique to it.

`cargo fmt --check`, both clippy passes and `cargo test` (2335 + 39 + 50 + 3
passed, 0 failed) are clean; the BLE transport's own 212 tests pass. All 16
static gates pass.

Integration suites were run one at a time rather than as a single pass, because
the full runner exhausts memory on this host: `chaos-tcp-mesh`,
`chaos-congestion-stress` and `chaos-churn-mixed-10`, all green. Between them
that is a TCP mesh under load, the queue-pressure path, and reconnection under
churn.

**What the evidence does not cover.** The BLE change is exercised through
`MockBleIo` only. Nothing here was run over a real L2CAP link, so the claim that
a stalled radio behaves like a stalled mock channel is reasoning rather than
measurement. The Android backend is `target_os = "android"` and there is no NDK
on this host, so it has no local compile evidence at all — the `android-check`
leg is the only one, as it was for the netmon Android `cfg` widening. It is
unchanged here except for the comment.

## Note on the retry gate in #148

#148 added a retry gate on failed heartbeats deliberately spaced at 2s rather
than the tick, because the send behind it was this unbounded write and retrying
it every second would have made a stranded stream a stalled node. With this
merged that constant can come down; the comment on it says so. Left alone here
so the two changes stay separable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Data plane
 
+- A peer that stops reading can no longer stall the node. TCP, Tor and Nym
+  wrote to their sockets directly from the caller's task, and `write_all`
+  blocks once the peer's receive window and this node's send buffer are both
+  full — which is what a peer that has gone away, or whose path has just
+  changed medium, produces. The callers are the rx loop's tick handlers, so a
+  single unresponsive peer held every other arm of the event loop behind it:
+  other peers' liveness, forwarding, and control RPCs. Each connection now has
+  its own writer task owning the write half, and sending is an enqueue onto a
+  bounded queue, so the only code that can await the wire is a task with
+  nothing else to do. A peer that stops draining fills its queue and its sends
+  then fail immediately, which is the signal the caller's retry and liveness
+  handling already expects. Frames are written whole, and a write error takes
+  the connection down with it, so a peer never sees a partial frame it cannot
+  resynchronise from.
+
 - A per-peer `connect()`-ed UDP socket is no longer left pinned to an interface
   the host has moved off. Established UDP peers get their own socket for the
   send fast path; `open_connected_fd` binds the wildcard and then calls
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Data plane
 
-- A peer that stops reading can no longer stall the node. TCP, Tor and Nym
-  wrote to their sockets directly from the caller's task, and `write_all`
+- A peer that stops reading can no longer stall the node. TCP, Tor, Nym and
+  BLE wrote to their links directly from the caller's task, and a write
   blocks once the peer's receive window and this node's send buffer are both
   full — which is what a peer that has gone away, or whose path has just
   changed medium, produces. The callers are the rx loop's tick handlers, so a
@@ -24,7 +24,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   then fail immediately, which is the signal the caller's retry and liveness
   handling already expects. Frames are written whole, and a write error takes
   the connection down with it, so a peer never sees a partial frame it cannot
-  resynchronise from.
+  resynchronise from. BLE was the worst of the four: it awaited the L2CAP
+  write while holding the connection-pool mutex, so one unresponsive peer
+  froze every other BLE operation as well — connects, evictions, and each
+  receive loop's teardown.
 
 - A per-peer `connect()`-ed UDP socket is no longer left pinned to an interface
   the host has moved off. Established UDP peers get their own socket for the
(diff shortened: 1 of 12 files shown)


jmcorgan:
Landed on `master` as `813c50bb`. Your two commits are squashed into one with your authorship and author date preserved; the squash rewrote them, so this closes rather than merges.

Four fold-ins on landing, all documentation and no behaviour change: the `tcp_receive_loop` and BLE `receive_loop` doc comments, which ended up attached to the writer loop inserted above them; the Android comment, since `pubkey_exchange` also calls that send and is bounded by its own timeout; and the netmon filter rationale plus its changelog clause, both of which still claimed the unbounded write your change removes. The filter itself is untouched, so widening the fan-out stays separable, as you intended.

The branch was pushed here before landing to get the Android cross-check onto it, since you had no local evidence for that leg. It passed.

The `next` line needed the BLE half authored separately: its XX handshake removed the pubkey exchange and the inbound admission path, so your version does not apply there and resolving it inside a merge was the wrong place for new code. That landed as `b24a9820`, same shape, with your regression test ported.

You were right to withdraw the bounded write. The reasoning in your description is better than the sketch it replaced, and the property now holds by construction.

Two follow-ups, neither blocking: the receive-loop teardown detaches the writer task rather than aborting it, unlike the other four teardown paths and unlike BLE's `Drop`; and a writer orphaned that way can then remove a pool entry by address with no identity check. Both are tracked.

PR #153, opened 2026-09-10
add a pfSense package

### LINKED ISSUE #151: fips-dns setup does not detect when /etc/resolv.conf bypasses systemd-resolved
### Summary

On Debian 12, `fips-dns.service` can be active and correctly configure
systemd-resolved for `.fips` queries, while normal applications still do
not use the FIPS DNS routing configuration.

The cause was that `/etc/resolv.conf` was a regular file containing DNS
server addresses instead of a symlink to systemd-resolved's local stub
resolver. In addition, the file had previously been made immutable, which
prevented it from being replaced.

The FIPS DNS service appeared healthy, but applications such as `ping` were
bypassing systemd-resolved and therefore could not use the `.fips` routing
rule.

### Environment

- OS: Debian GNU/Linux 12 (bookworm)
- Network configuration: ifupdown
- NetworkManager: not installed
- Resolver: systemd-resolved
- Network interface: `eth0`
- FIPS interface: `fips0`
- FIPS service: enabled and running

### Initial state

The FIPS daemon was running:

```text
fips.service       active (running)
fips-dns.service   active (exited)
```

The FIPS resolver configuration was present:

```ini
# /etc/systemd/resolved.conf.d/fips.conf

[Resolve]
DNS=[::1]:5354
Domains=~fips
```

`systemd-resolved` therefore had the expected global configuration:

```text
Global
    DNS Servers [::1]:5354
    DNS Domain ~fips
```

However, `/etc/resolv.conf` was originally a regular file containing DNS
servers directly:

```text
nameserver 46.38.225.230
nameserver 46.38.252.230
nameserver 2a03:4000:0:1::e1e6
```

Because applications read this file directly, they did not use the local
systemd-resolved stub at `127.0.0.53`. As a result, the routing rules known
to systemd-resolved were bypassed.

### Steps to reproduce

1. Install and configure FIPS on a Debian 12 system using `ifupdown`.
2. Ensure `systemd-resolved` is running.
3. Configure the FIPS DNS service.
4. Leave `/etc/resolv.conf` as a regular file containing external DNS
   servers.
5. Query a `.fips` name using a normal application such as `ping` or
   `getent`.

Example:

```bash
systemctl status fips.service fips-dns.service
resolvectl status
ls -l /etc/resolv.conf
getent hosts <name>.fips
```

`systemd-resolved` reports the correct `~fips` route, but normal applications
may not use it because `/etc/resolv.conf` bypasses systemd-resolved.

### Workaround

The issue was fixed manually with:

```bash
sudo chattr -i /etc/resolv.conf 2>/dev/null || true
sudo rm -f /etc/resolv.conf
sudo ln -s /run/systemd/resolve/stub-resolv.conf /etc/resolv.conf
sudo systemctl enable fips-dns.service
```

After applying the workaround:

```bash
resolvectl query google.com
resolvectl query <name>.fips
getent hosts google.com
getent hosts <name>.fips
```

Both regular domains and `.fips` domains resolved successfully.

### Expected behavior

The FIPS DNS setup should either:

1. Ensure that `/etc/resolv.conf` points to the systemd-resolved stub when
   systemd-resolved is the selected backend; or
2. Detect that `/etc/resolv.conf` bypasses systemd-resolved and display a
   clear warning with the required remediation steps.

It would also be useful for the setup to detect an immutable
`/etc/resolv.conf` and report that the file cannot be replaced until its
immutable attribute is removed.

### Questions

- Is managing `/etc/resolv.conf` intentionally outside the scope of
  `fips-dns-setup`?
- Should the setup script verify that applications can reach the configured
  resolver backend?
- Should the documentation explicitly describe the required
  `/etc/resolv.conf` symlink for Debian systems using systemd-resolved?
- Should `fips-dns.service` or its setup script warn when
  `/etc/resolv.conf` is a regular file or has the immutable attribute set?

### Suggested improvement

After selecting the systemd-resolved backend, the setup script could check:

```bash
readlink -f /etc/resolv.conf
```

and warn if the result is not:

```text
/run/systemd/resolve/stub-resolv.conf
```

It could also check the immutable attribute with:

```bash
lsattr /etc/resolv.conf
```

and provide an actionable message if the file cannot be replaced.

This may be primarily an installation/documentation issue rather than a
problem with the FIPS DNS responder itself, but the current behavior can
leave the service active while `.fips` resolution is unavailable to normal
applications.

fr34aky:
## pfSense packaging

Adds `packaging/pfsense/` — a pfSense `.pkg` shipping `fips`, `fipsctl`,
`fipstop`, a boot script, and `.fips` DNS integration. `fips-gateway` is
excluded (nftables backend, Linux-only).

pfSense is FreeBSD underneath, but the FreeBSD package does not work
there, and fails silently in three ways. That is why this is a separate
build rather than a flag on `packaging/freebsd/`:

| | FreeBSD package | pfSense |
|---|---|---|
| Boot | `rc.d/fips`, an `rc.conf`-gated rc.subr service | `rc.start_packages` globs `/usr/local/etc/rc.d/*.sh` and runs each as `<script> start`, at boot and again on a WAN IP change. A suffixless script never runs; the `.sh` script must make a re-run a quiet no-op. |
| DNS | drop-in at `/var/unbound/conf.d/` | `unbound.conf` is generated from `config.xml` and includes no `conf.d` directory, so the drop-in is never read. |
| Responder bind | `::1` | `do-ip6: no` unless "Allow IPv6" is set, so `[::1]` is unreachable. |

So: `fips.sh`, DNS via the DNS Resolver custom options in `config.xml`,
and the responder bound on `127.0.0.1`.

`start` is idempotent: that same glob is re-run when WAN gets a new
address (`rc.newwanip` -> `service reload packages`), so a re-run must
neither fork a second daemon nor disturb the one already running. (A
plain link flap on a static-address interface does not reach that path;
a WAN address change does — confirmed on hardware.) A running daemon is
identified by process name, not by the pidfile alone — a recycled pid
would otherwise make `start` a
permanent no-op.

The DNS integration is deliberately **not** run from post-install:
installing a package should not rewrite a firewall's `config.xml`
unannounced. It is a documented manual step, revertable from Config
History. The block is written between markers, leaving the operator's
own custom options untouched.

### Statically linked by default

Unlike every other platform's package. pfSense runs a FreeBSD base that
cannot be obtained — Netgate builds Plus from its own CURRENT snapshot,
and only the most recent CURRENT builds are published — so a build
host's `libc` is normally newer than the appliance's. That is the
direction that breaks: the binary references a versioned symbol the
appliance does not export, installs cleanly, and then will not start.

Static linking removes the negotiation rather than managing it. The
package declares no shared libraries at all, where a dynamic one needs
`libc.so.7`, `libm.so.5`, `libthr.so.3` and `libgcc_s.so.1` to agree
with it. It also survives a firmware upgrade's change of base, where a
dynamic package is pinned to the image it was built against.

Viable here because nothing in the tree uses `dlopen`/`libloading`, and
FreeBSD compiles `files`/`dns` resolution into `libc`, so a static
binary still resolves hostnames. `crt-static` is a request a target may
silently ignore, so the build inspects the produced binaries and fails
if any came out dynamic. `--dynamic` opts out.

**Except on aarch64, where it is refused.** A statically linked aarch64
FreeBSD binary faults at `addr=0x0` where `posix_spawn` should be, so the
daemon dies the first time it shells out — which is `sysctl`, called from
`is_ipv6_disabled()` at the top of `TunDevice::create`. That makes it
present as a TUN bug, and with `tun.enabled: false` the daemon never
spawns anything and looks healthy. The same trace on static amd64 reaches
`rfork(RFSPAWN)` and spawns normally, so this is specific to the
architecture, not to static linking. `build-pkg.sh` refuses `--static`
there rather than silently downgrading, and amd64 keeps the default its
unobtainable-base problem needs.

### Naming and ABI

Packages are named for the pfSense product, not the FreeBSD ABI
(`fips-<version>-pfsense-<product>-<arch>.pkg`), because the product is
what someone choosing a download knows. Architecture alone would not do:
CE 2.8 is `FreeBSD:15:amd64` while CE 2.9 and Plus 26.x on Intel are
`FreeBSD:16:amd64` — all three "amd64", but the FreeBSD 15 and 16
packages are not installable on each other's base, so named by arch
alone they would collide on one filename and the second build would
overwrite the first. (CE 2.9 and Plus 26.x on Intel do share one
`FreeBSD:16:amd64` artifact, named for both products.)

A product name promises something the ABI does not, so the
product-to-base mapping is declared in one place and the build fails
when the two disagree, and refuses to guess for an unknown base.

`--target` supports a build host already of the right architecture and
cross-checks the triple against `--abi`, because `--abi` alone only
relabels the manifest: a package labelled `FreeBSD:16:aarch64` full of
x86-64 binaries installs and then cannot exec.

### Toolchain provenance

Every package records the compiler that produced it, its linkage, the
pfSense product, and whether the `rust-toolchain.toml` pin was honoured,
readable from the artifact with `pkg info -A`.

This exists because aarch64 cannot honour the pin: rustup publishes no
toolchain for `aarch64-unknown-freebsd` in either direction, so an ARM
build must use the ports Rust, whose cargo ignores `rust-toolchain.toml`
outright. `rust-toolchain.toml` is unchanged — it still governs every
rustup-capable platform. The build refuses below the edition-2024 floor
(1.85) and prints a notice naming both versions when the compiler is not
the pinned one.

**No aarch64 package is published as a release artifact**, for that
reason; it is build-it-yourself, documented in the README. Rejected
alternatives are recorded there too, so the decision is not silently
revisited.

### Why this is a fork of `packaging/freebsd/` rather than a caller

A reviewer will reasonably ask why this is a separate builder. It is
not a copy: the mechanics the two share — argument parsing, the
`tr -- '+-' '..'` version mapping, the stage layout, the manifest
fields, the install-if-absent `@sample` scripts and `pkg create` — are
extracted into `packaging/common/pkg-lib.sh`, which both builders
source. The FreeBSD package is byte-identical before and after that
extraction (its manifest and file list; the acceptance test for the
refactor, since it is a released artifact). What is left in each builder
is only what the two genuinely disagree about.

The divergence is not stylistic. Each piece is forced by a pfSense
behaviour the FreeBSD package has no reason to know about:

- pfSense runs `/usr/local/etc/rc.d/*.sh` at boot and re-runs them when
  WAN gets a new address, so the boot script needs the `.sh` name and an
  idempotent start. `rc.subr` would in fact work (it reads
  `rc.conf.d/<name>` and a script may default its rcvar to YES); plain
  sh is a choice, to keep what happens on a re-run visible in the file
  rather than behind `rc.subr`.
- `unbound.conf` is generated from `config.xml` with no `conf.d`, so the
  DNS drop-in becomes a marked block edited through PHP, and the
  `fips_dns` rc service disappears entirely.
- unbound is generated with `do-ip6: no` when "Allow IPv6" is off (it is
  on by default), so the responder binds IPv4 loopback for robustness
  and the shipped `dns:` block is owned here.
- The base cannot be obtained, so linkage, ABI/product naming and
  toolchain provenance all become build-time decisions the FreeBSD
  package never has to make.

Where the duplication is worth attacking, if the maintainer wants it, is
that shared 20% of boilerplate — not the divergent logic, which is
divergent for reasons the comments record.

### CI

The pfSense package is built and checked in its own job (`pfsense`), in
its own FreeBSD VM, with no `continue-on-error`. The `release` job does
not depend on it, so a pfSense-only failure reds that job and nothing
else; and the release job downloads only the FreeBSD artifact (by
`pattern`) and fails if a `*-pfsense-*` file ever reaches `dist/`, so the
pfSense package cannot become a release asset by accident. It is a
workflow artifact until someone has installed it on a real pfSense box.
The `build` job is unchanged from `master`.

It compiles its own binaries rather than reusing the FreeBSD build:
`crt-static` is different codegen, so the dynamic binaries cannot be
repackaged as static — the artifact assertion would reject them.

### Release coverage, and the one open question

No pfSense package is a release asset. `pkg` enforces the whole
ABI, not just the architecture, so the three targets are not
interchangeable:

| Target | pkg ABI | linkage | pin | Published by CI |
|---|---|---|---|---|
| CE 2.8.1, Intel | `FreeBSD:15:amd64` | static | honoured | built + checked, held back* |
| CE 2.9.0 and Plus 26.x, Intel | `FreeBSD:16:amd64` | static | honoured | no — no FreeBSD 16 host |
| Plus 26.x, ARM | `FreeBSD:16:aarch64` | dynamic | **not** honoured | no, by design |

All three have been built and pass the checker. The Intel Plus package
was produced on a FreeBSD 16 amd64 host and carries `linkage: static`,
`pin_honoured: yes` — the same provenance as every other published
artifact — so its absence from releases is an infrastructure gap, not a
property of the target.

\* The CE 2.8.1 package is built and checked in CI but **not attached to
a release**: it has not been installed on a real CE 2.8.1 box, and that
base can no longer be obtained (the public mirror stops at 2.7.2; the
Netgate installer only lays down the current release). It stays a 30-day
workflow artifact for anyone who wants to test it. This is the
maintainer's own item-1 fallback.

The two absences are not the same kind of absence, and only one of them
is a decision for the maintainer.

**ARM is excluded deliberately and permanently.** rustup publishes no
toolchain for `aarch64-unknown-freebsd` in either direction, so such a
build cannot honour `rust-toolchain.toml` and would carry
`pin_honoured: no`. Publishing it beside artifacts that do honour the pin
would imply a provenance it does not have. It is documented as
build-it-yourself, and a comment in the workflow says that if an aarch64
leg is ever added it must upload for testing only and stay out of the
release job.

**Intel Plus is absent only for want of a build host, and that is worth
a decision.** On amd64 rustup serves `x86_64-unknown-freebsd`, so the
pinned compiler installs and the package would carry
`pin_honoured: yes` — the same provenance as every other released
artifact. Static linking also works there (it faults at `posix_spawn`
only on aarch64), so it would ship `linkage: static` with no `libc.so`
to drift against Netgate's base. By both of the tests that exclude ARM,
Intel Plus passes.

What stops it is infrastructure rather than principle:
`vmactions/freebsd-vm`, which this workflow uses, offers nothing newer
than 15.1, and FreeBSD 16 is not released — pfSense Plus is built from
16.0-CURRENT snapshots. So an Intel Plus leg needs a builder running a
CURRENT snapshot, which is a different provisioning model from anything
in this repo today, and it pins CI to a moving base that will not match
Netgate's snapshot anyway. Static linking is what makes that mismatch
survivable.

This PR does not force that decision: it builds and checks the CE 2.8.1 amd64 package in CI (held back from releases until tested),
documents ARM as build-it-yourself, and simply does not mention Intel
Plus as a release artifact. Adding it is follow-up work.

### Tests

`testing/check-pfsense-pkg.sh` validates a built package on any FreeBSD
host: contents, modes, the `.sh` boot script, pid handling including
recycled pids, the shipped config, the annotations, and that `linkage`
and the ABI describe the actual binaries rather than the intent.

It is 58 checks, including a positive boot-script lifecycle against a
stub daemon (start/status/re-entrant start/restart/stop, with the
no-trailing-newline pidfile the read must handle), `php -l` on the
config.xml helper, and a unit test of its `fips_strip_block`. Re-adding
the original `read ... || return 1` trips four of those checks, so the
defect that shipped once is now caught.

It still cannot drive the **live** config.xml edit (that needs a running
pfSense with its `config.inc`), unbound actually answering `.fips`, or pf
passing mesh traffic. There is no pfSense CI image for those; the README
marks them manual rather than letting a pass imply more than it checks.

**Gates run.** `cargo fmt --check`, `cargo build` and `cargo test`
(2142 passing) on FreeBSD; both `cargo clippy` passes and
`testing/ci-local.sh` on Linux, the platform the CI gates actually run
on. `check-ci-parity`, `check-action-pins`, `check-image-scoping`,
`check-comment-refs`, `check-log-strings` and `check-trailing-log` all
pass. `shellcheck` reports nothing across the seven shell scripts this
adds or touches, at style level.

Two results worth reading rather than skipping:

- `cargo test` fails one test, `transport::udp::tests::
  test_resolve_socket_addr_invalid`, on any host whose DNS search domain
  has a wildcard A record: `nonexistent.invalid` then resolves and the
  assertion that it must not inverts. Pre-existing, unrelated to this
  change, and a trailing dot in the test literal would make it
  search-independent — left alone, as it is a file this change does not
  otherwise touch.
- Both clippy passes fail when run *on FreeBSD*, on `unused_imports` and
  dead code in `src/node/tests/netmon.rs` and `src/nostr/tests.rs`.
  Those symbols are used only inside `#[cfg(any(target_os = "linux",
  target_os = "macos"))]` tests while their imports are ungated, so they
  are unused on FreeBSD and used on Linux, where the gates run. This
  change contains no `.rs` files at all.

### Validated on hardware

Installed and run on pfSense Plus 26.03.1 aarch64: package install, the
boot script's start/stop/restart/status lifecycle, `fips-dns-setup`
writing the DNS Resolver block into `config.xml`, `.fips` resolving
through unbound, the TUN interface coming up, and the mesh carrying
traffic (304 packets delivered, no loss, no drops) — plus reboot and a
WAN-event re-run of `start`. That is one run on one appliance rather than
a gate; nothing re-checks it when this code changes. It exercised the
helper's core path — write the block, restart unbound, unbound answers
`.fips` — and the additions this PR makes to it were then run on the same box: a
re-apply passes the `unbound-checkconf` pre-check and writes; a
deliberately broken snippet is refused with `config.xml` left untouched
and DNS still up (the fail-closed path); and teardown removes the block,
leaving `custom_options` byte-for-byte as it was before FIPS. That run
also caught a real bug in the pre-check — the `test_unbound_config`
return code was tested inverted, so it would have rejected every valid
re-apply — now fixed; a FreeBSD CI host cannot catch it, as it has no
`test_unbound_config`.

That run found several defects, all in this packaging and none in the
daemon; they are folded into this commit. The two worth a reviewer's
attention, because both were the same mistake:

- `running_pid()` guarded its pidfile read with `|| return 1`.
  `daemon(8)` writes the pid with no trailing newline, so `read` reaches
  EOF and returns non-zero *having read the value* — the good pid was
  discarded and the daemon reported dead. `status` was therefore always
  wrong, `stop` never stopped anything, `restart` never came back, and
  every start orphaned another supervisor that blocked the next one.
  `stop_supervisor` had the identical bug on an empty pidfile.
- A first attempt at the orphan problem swept for supervisors by process
  title before every start. pfSense re-enters `start` on every interface
  event, so it killed the supervisor just started, and `daemon(8)`
  forwards the signal to its child — a clean shutdown seconds after
  start, with nothing in the log to explain it. Recovery now targets
  only the pid `daemon(8)` names, after confirming that pid is a
  `daemon(8)`; nothing matches on title.

### Still unexercised

- `fips-dns-setup`'s daemon-not-listening refusal path, and its DNS
  Forwarder (dnsmasq) branch — the box runs the DNS Resolver.
- The byte-exact teardown *preserving other operator text*: on the box
  the only custom option was the FIPS block, so its clean removal is
  shown but not the "leaves the rest alone" case. That case is unit
  tested (`fips_strip_block` over several fixtures), not run on hardware.
- `pkg delete`. And installing any amd64 package: the CE 2.8.1 one is
  built and checked in CI, the CE 2.9 / Plus 26 one was built and checked
  in a FreeBSD 16 VM (CI has no FreeBSD 16 host) — neither has been
  installed anywhere.

### Related

Related to #151, which reports the Linux form of the same failure: a DNS
setup that configures the resolver correctly while nothing on the box
actually asks it, so `.fips` never resolves and the service still looks
healthy. This PR does not fix #151 — that is the systemd-resolved path —
but the pfSense setup script refuses to touch `config.xml` unless
something is listening on the responder's address, validates the merged
resolver config with `unbound-checkconf` before writing it, and warns
when `/etc/resolv.conf` points somewhere other than the resolver it just
configured. Both checks exist because
this exact failure happened during the hardware run: the forward-zone
was correct and the queries went to a public server instead.

### DIFF
--- a/.github/workflows/package-freebsd.yml
+++ b/.github/workflows/package-freebsd.yml
@@ -227,6 +227,99 @@ jobs:
           echo "Build Summary for freebsd/x86_64:"
           echo "  Package: ${{ steps.freebsd-assets.outputs.pkg }}"
 
+  pfsense:
+    name: Build and check the pfSense package (x86_64)
+    # A job of its own, and deliberately NOT a dependency of `release`. A
+    # pfSense-only failure — a new key in the common dns: block, a
+    # dependency that stops linking statically, a checker regression — reds
+    # this check and nothing else; it can never hide behind the FreeBSD
+    # job's result, and it cannot block the FreeBSD release asset. The
+    # pfSense package is therefore never a release asset. It is published
+    # here as a workflow artifact (30-day retention) for anyone to test,
+    # until someone has installed it on a real pfSense box; see
+    # packaging/pfsense/README.md.
+    #
+    # This VM is FreeBSD 15.1, so the package it produces is FreeBSD:15:amd64
+    # (pfSense CE 2.8.1). CE 2.9 and Plus 26.x are FreeBSD 16 and need a
+    # FreeBSD 16 host this workflow does not have. No aarch64 package is built
+    # here or published anywhere: rustup ships no toolchain for
+    # aarch64-unknown-freebsd, so such a build cannot honour the
+    # rust-toolchain.toml pin every published artifact is built with. ARM is
+    # build-it-yourself, per the README.
+    runs-on: ubuntu-latest
+    needs: determine-versioning
+    # Its own full release build in an emulated FreeBSD VM, like the build
+    # job; the same generous bound applies.
+    timeout-minutes: 45
+
+    steps:
+      - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
+        with:
+          fetch-depth: 0
+
+      - name: Set SOURCE_DATE_EPOCH from git
+        run: echo "SOURCE_DATE_EPOCH=$(git log -1 --format=%ct)" >> "$GITHUB_ENV"
+
+      - name: Build and check the pfSense package in a FreeBSD VM
+        uses: vmactions/freebsd-vm@f0552d3b69211736abd97f02ff3d4674c56b73b1 # v1
+        env:
+          FREEBSD_PACKAGE_VERSION: ${{ needs.determine-versioning.outputs.freebsd_package_version }}
+        with:
+          release: "15.1"
+          usesh: true
+          sync: rsync
+          copyback: true
+          mem: 6144
+          envs: "SOURCE_DATE_EPOCH CARGO_TERM_COLOR FREEBSD_PACKAGE_VERSION"
+          prepare: |
+            # curl for rustup; bash and php for testing/check-pfsense-pkg.sh
+            # (the checker is bash, and it runs php -l plus a fips_strip_block
+            # unit test on the config.xml helper).
+            pkg install -y curl bash php85
+          run: |
+            set -e
+            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs               | sh -s -- -y --default-toolchain none --profile minimal
+            . "$HOME/.cargo/env"
+
+            # Builds the release binaries (static by default) and packages
+            # them; on a FreeBSD 15.1 VM this yields the FreeBSD:15:amd64
+            # package for pfSense CE 2.8.1.
+            packaging/pfsense/build-pkg.sh --version "$FREEBSD_PACKAGE_VERSION"
+
+            PKG=$(ls deploy/fips-*-pfsense-*.pkg)
+            testing/check-pfsense-pkg.sh "$PKG"
+            php -l packaging/pfsense/fips-unbound-custom.php
+
+            ( cd deploy && sha256 -q "$(basename "$PKG")" \
+                | { read -r h; printf '%s  %s\n' "$h" "$(basename "$PKG")"; } \
+                > "$(basename "$PKG").sha256" )
+
+            rm -rf target
+
+      - name: Resolve pfSense asset path
+        id: pfsense-asset
+        shell: bash
+        run: |
+          : ${GITHUB_OUTPUT:=/tmp/github_output}
+          set -euo pipefail
+          VER="${{ needs.determine-versioning.outputs.freebsd_pkg_file_version }}"
+          PKG=$(ls deploy/fips-${VER}-pfsense-*.pkg)
+          if [[ ! -f "$PKG" ]]; then
+            echo "No pfSense package was produced" >&2
+            ls -la deploy >&2 || true
+            exit 1
+          fi
+          echo "pkg=$PKG" >> "$GITHUB_OUTPUT"
+
+      - name: Upload pfSense artifact
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: fips_${{ needs.determine-versioning.outputs.freebsd_package_version }}_x86_64_pfsense
+          path: |
+            ${{ steps.pfsense-asset.outputs.pkg }}
+            ${{ steps.pfsense-asset.outputs.pkg }}.sha256
+          retention-days: 30
+
   release:
     name: Publish FreeBSD assets to GitHub Release
     runs-on: ubuntu-latest
@@ -239,6 +332,11 @@ jobs:
       - name: Download FreeBSD artifacts
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
         with:
+          # Only the FreeBSD artifact. Without a pattern this action downloads
+          # every artifact in the run — including the pfSense package from the
+          # `pfsense` job, which is a workflow artifact by design and must never
+          # reach a release. `needs` only orders jobs; it does not scope this.
+          pattern: fips_*_x86_64_freebsd
           path: dist
           merge-multiple: true
 
@@ -247,6 +345,15 @@ jobs:
         run: |
           set -euo pipefail
           cd dist
+
+          # The pfSense package is never a release asset (see the `pfsense`
+          # job). The download step is scoped to the FreeBSD artifact; this
+          # keeps the rule if that artifact is ever renamed or a new one added.
+          if compgen -G '*-pfsense-*' >/dev/null; then
+            echo "FAIL: a pfSense package reached the release stage; it must stay a workflow artifact:" >&2
+            ls -la ./*-pfsense-* >&2 || true
+            exit 1
+          fi
 
           pkgs=$(find . -maxdepth 1 -type f -name '*.pkg' | LC_ALL=C sort)
           if [[ -z "$pkgs" ]]; then
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -137,6 +137,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   a typo'd interface name were previously the same flat
   `StartFailed(String)`; nothing downstream could branch on absence.
 
+#### Packaging
+
+- A pfSense package (`packaging/pfsense/`, `gmake pfsense`). pfSense is
+  FreeBSD underneath, but the FreeBSD package fails there in three
+  silent ways: pfSense runs only `/usr/local/etc/rc.d/*.sh` at boot and
+  re-runs them on WAN IP changes, so the suffixless rc script never
+  starts; `unbound.conf` is generated from `config.xml` with no `conf.d`,
+  so the DNS drop-in is never read; and — on a firewall where the
+  default-on "Allow IPv6" has been turned off — the responder's `::1`
+  default is unreachable when unbound is then generated with
+  `do-ip6: no`, so it binds `127.0.0.1` for robustness. The package
+  ships `fips.sh`, wires the `fips.` zone into the DNS Resolver through
+  `config.xml`, and binds the responder on `127.0.0.1`. It links
+  statically by default, since pfSense runs a FreeBSD base that cannot
+  be obtained to link against; aarch64 is refused, where static binaries
+  fault at `posix_spawn`. Mechanics shared with the FreeBSD builder live
+  in `packaging/common/pkg-lib.sh`, which both source; the FreeBSD
+  package is byte-identical before and after. The pfSense package is
+  built and checked in its own CI job and published as a workflow
+  artifact, not attached to a release, until it has been installed on a
+  real pfSense box; CI produces the CE 2.8.1 (`FreeBSD:15:amd64`)
+  package, while CE 2.9 and Plus 26.x on Intel need a FreeBSD 16 build
+  host the CI does not have, and ARM stays build-it-yourself because
+  rustup ships no toolchain for it. See `packaging/pfsense/README.md`.
+
 ### Changed
 
 - `Degraded` is now a level rather than a latch. The supervisor's reason set
--- a/README.md
+++ b/README.md
@@ -132,7 +132,8 @@ default `/etc/fips/fips.yaml` you can edit before starting. The package
 enables `fips` and `fips-dns` but starts neither, which is why the
 second command is there.
 
-For macOS, Windows, FreeBSD, OpenWrt, the systemd tarball or a Nix
+For macOS, Windows, FreeBSD (including a pfSense build under
+`packaging/pfsense/`), OpenWrt, the systemd tarball or a Nix
 flake, see [docs/getting-started.md](docs/getting-started.md)
 for the full multi-platform installation guide.
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -88,6 +88,12 @@ included: its NAT backend is nftables, which is Linux-only. The
 Ethernet and BLE transports are unavailable on FreeBSD; UDP, TCP,
 Tor, and Nym are.
 
+On **pfSense** (CE or Plus) use `packaging/pfsense/` rather than this
+package: pfSense diverges from stock FreeBSD in how it boots packages,
+generates the DNS resolver config, and applies upgrades, and the pfSense
+package handles each. See
+[packaging/pfsense/README.md](../packaging/pfsense/README.md).
+
 **One architecture.** The published artifact is
 `fips-<version>-freebsd-amd64.pkg`. There is no aarch64 FreeBSD
 build, so on any other architecture use the from-source path below.
--- a/packaging/Makefile
+++ b/packaging/Makefile
@@ -12,6 +12,7 @@
 #   make aur        Build fips-git AUR package and validate with namcap
 #   make pkg        Build a macOS .pkg installer
 #   make freebsd    Build a FreeBSD .pkg package (on FreeBSD; use gmake)
+#   make pfsense    Build a pfSense .pkg package (on FreeBSD; use gmake)
 #   make zip        Build a Windows .zip package
 #   make all        Build deb and tarball (default)
 #   make clean      Remove deploy/ directory
@@ -20,7 +21,7 @@ SHELL := /bin/bash
 PACKAGING_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 PROJECT_ROOT  := $(abspath $(PACKAGING_DIR)/..)
 
-.PHONY: all deb deb-host tarball ipk apk aur pkg freebsd zip clean
+.PHONY: all deb deb-host tarball ipk apk aur pkg freebsd pfsense zip clean
 
 all: deb tarball
 
@@ -56,6 +57,12 @@ pkg:
 
 freebsd:
 	@sh $(PACKAGING_DIR)/freebsd/build-pkg.sh
+
+# pfSense is FreeBSD underneath but boots, resolves and is upgraded
+# differently enough that the FreeBSD package does not work there; see
+# packaging/pfsense/README.md for the three divergences.
+pfsense:
+	@sh $(PACKAGING_DIR)/pfsense/build-pkg.sh
 
 zip:
 	@powershell -File $(PACKAGING_DIR)/windows/build-zip.ps1
--- a/packaging/README.md
+++ b/packaging/README.md
@@ -14,6 +14,7 @@ make apk        # OpenWrt .apk (apk-tools, mandatory on OpenWrt 25+)
 make aur        # Arch Linux AUR package (fips-git, local build + namcap)
 make pkg        # macOS .pkg installer
 make freebsd    # FreeBSD .pkg package (on FreeBSD; use gmake)
+make pfsense    # pfSense .pkg package (on FreeBSD; use gmake)
 make zip        # Windows .zip package
 make all        # deb + tarball (default)
 ```
@@ -70,9 +71,11 @@ runtime dependency and is not needed to build.
 ```text
 packaging/
   aur/            Arch Linux AUR packaging (PKGBUILD, supporting files)
-  common/         Shared assets (default config, hosts file)
+  common/         Shared assets (default config, hosts file) and pkg-lib.sh,
+                  the helpers the FreeBSD and pfSense builders share
   debian/         Debian/Ubuntu .deb packaging via cargo-deb
   freebsd/        FreeBSD .pkg packaging via pkg-create(8)
+  pfsense/        pfSense .pkg packaging (FreeBSD-based, but not the same)
   macos/          macOS .pkg installer via pkgbuild
   nixos/          NixOS flake module (services.fips.*)
   systemd/        Generic Linux systemd tarball packaging
@@ -211,6 +214,58 @@ service fips_dns start
 
 See [freebsd/README.md](freebsd/README.md) for host resolver setup and
 field-tested caveats.
+
+### pfSense (`.pkg`)
+
+pfSense is FreeBSD underneath, but the FreeBSD package does not work
+there, and fails silently in three ways: pfSense boots packages by
+globbing `/usr/local/etc/rc.d/*.sh` (a suffixless rc script is never
+run), it generates `unbound.conf` from `config.xml` and reads no
+`conf.d` directory (the DNS drop-in is never read), and it writes
+`do-ip6: no` unless "Allow IPv6" is enabled (so a responder on `[::1]`
+is unreachable). This package ships `fips.sh`, integrates DNS through
+the DNS Resolver custom options in `config.xml`, and binds the
+responder on `127.0.0.1`.
+
+Unlike the other packages, this one **links statically by default**
+(`--dynamic` opts out). pfSense runs a FreeBSD base you cannot
+obtain — Netgate builds Plus from its own 16.0-CURRENT snapshot — so
+a dynamically linked binary can reference a libc symbol the appliance
+does not export, install cleanly, and then refuse to start. A static
+package declares no shared libraries at all.
+
+**On aarch64 this is refused, not applied.** A statically linked
+aarch64 FreeBSD binary faults where `posix_spawn` should be, so the
+daemon dies the first time it shells out. ARM builds must pass
+`--dynamic`, and then `ldd` on the appliance is the check that the
+base drift is not real.
+
+The build host's architecture and FreeBSD major must still match the
+target's: pfSense CE 2.8.1 is FreeBSD 15 amd64; CE 2.9.0 and Plus 26.x are
+FreeBSD 16 (amd64, plus aarch64 for Plus on ARM appliances), and `pkg` refuses a
+mismatched ABI. No aarch64 package is published: rustup ships no
+toolchain for aarch64 FreeBSD, so such a build cannot honour the
+`rust-toolchain.toml` pin. It is build-it-yourself.
+
+```sh
+# Build (on FreeBSD; this Makefile needs GNU make — pkg install gmake)
+gmake pfsense
+# or directly, no gmake needed:
+./packaging/pfsense/build-pkg.sh
+
+# Validate the package before shipping it
+./testing/check-pfsense-pkg.sh deploy/fips-<version>-pfsense-ce2.8-amd64.pkg
+
+# Install (on the firewall, as root)
+pkg add ./fips-<version>-pfsense-ce2.8-amd64.pkg
+/usr/local/etc/rc.d/fips.sh start
+/usr/local/libexec/fips/fips-dns-setup   # edits config.xml; run deliberately
+```
+
+Not a Netgate-supported package, and a pfSense firmware upgrade removes
+it. See [pfsense/README.md](pfsense/README.md) for the "Allow IPv6"
+prerequisite the mesh depends on, firewall-rule notes, and removal
+behaviour.
 
 ### Windows (`.zip`)
 
--- a/packaging/common/pkg-lib.sh
+++ b/packaging/common/pkg-lib.sh
@@ -1,0 +1,173 @@
+# Shared helpers for the FreeBSD-family package builders.
+#
+# packaging/freebsd/build-pkg.sh and packaging/pfsense/build-pkg.sh
+# produce different packages for different systems, but the mechanics of
+# getting there — deriving a pkg-legal version, locating the binaries,
+# laying out the stage, and the manifest scripts that give the config
+# files @sample semantics — are the same work, and were duplicated
+# verbatim. They live here so a fix lands in both.
+#
+# What deliberately does NOT live here is anything the two packages
+# disagree about: the boot script, the DNS integration, linkage, the ABI
+# and product naming. Those differences are the reason there are two
+# builders at all, and folding them into a shared file with flags would
+# hide them.
+#
+# POSIX sh, sourced with `.` — no bashisms, no `local`.
+
+# Print the version to stamp on the package, given the project root and
+# an optional override (CI passes a derived version on branch builds).
+#
+# '-' is the pkg name/version separator and neither '-' nor '+' is legal
+# inside a pkg version, so both map to '.': 0.6.0-dev -> 0.6.0.dev.
+pkg_resolve_version() {
+    pkg_rv_root="$1"
+    pkg_rv_version="${2:-}"
+    [ -n "$pkg_rv_version" ] \
+        || pkg_rv_version="$(sed -n 's/^version = "\(.*\)"/\1/p' "${pkg_rv_root}/Cargo.toml" | head -1)"
+    if [ -z "$pkg_rv_version" ]; then
+        echo "error: could not read version from Cargo.toml" >&2
+        return 1
+    fi
+    printf '%s\n' "$pkg_rv_version" | tr -- '+-' '..'
+}
+
+# The build host's pkg ABI, or a sane default where pkg cannot say.
+pkg_host_abi() {
+    pkg config abi 2>/dev/null || echo "FreeBSD:15:amd64"
+}
+
+# Fail unless every named binary is present and executable in $1.
+pkg_require_binaries() {
+    pkg_rb_dir="$1"
+    shift
+    for pkg_rb_bin in "$@"; do
+        if [ ! -x "${pkg_rb_dir}/${pkg_rb_bin}" ]; then
+            echo "error: ${pkg_rb_dir}/${pkg_rb_bin} missing (run without --no-build)" >&2
+            return 1
+        fi
+    done
+    return 0
+}
+
+# The directory skeleton both packages install into.
+pkg_stage_tree() {
+    install -d "$1/usr/local/bin" \
+               "$1/usr/local/etc/fips" \
+               "$1/usr/local/etc/rc.d" \
+               "$1/usr/local/libexec/fips"
+}
+
+# Copy the named binaries from $1 into the stage at $2.
+pkg_stage_binaries() {
+    pkg_sb_dir="$1"
+    pkg_sb_stage="$2"
+    shift 2
+    for pkg_sb_bin in "$@"; do
+        install -m 0755 "${pkg_sb_dir}/${pkg_sb_bin}" "${pkg_sb_stage}/usr/local/bin/" || return 1
+    done
+    return 0
+}
+
+# Emit the post-install lines that copy a sample into place if the real
+# file is absent. Arguments are "<name>:<mode>" pairs, e.g. fips.yaml:0600.
+#
+# This is the @sample plist keyword spelled out by hand: that keyword
+# lives in the ports tree (/usr/ports/Keywords/sample.ucl), which neither
+# a plain pkg-create host nor pfSense has.
+#
+# FreeBSD has no "root" group; wheel is gid 0.
+pkg_sample_seed_script() {
+    for pkg_ss_entry in "$@"; do
+        pkg_ss_name="${pkg_ss_entry%%:*}"
+        pkg_ss_mode="${pkg_ss_entry##*:}"
+        cat <<EOS
+[ -f /usr/local/etc/fips/${pkg_ss_name} ] || install -m ${pkg_ss_mode} -o root -g wheel \\
+    /usr/local/etc/fips/${pkg_ss_name}.sample /usr/local/etc/fips/${pkg_ss_name}
+EOS
+    done
+}
+
+# Emit the pre-deinstall lines that remove a seeded config only when it
+# is still byte-identical to the sample, so an edited config — and the
+# identity it may carry — is never deleted. Arguments are bare names.
+pkg_sample_purge_script() {
+    printf '    for f in %s; do\n' "$*"
+    cat <<'EOS'
+        s="/usr/local/etc/fips/${f}.sample"
+        t="/usr/local/etc/fips/${f}"
+        if [ -f "$t" ] && cmp -s "$t" "$s"; then rm -f "$t"; fi
+    done
+EOS
+}
+
+# Run pkg-create(8) over a staged tree and move the result to $4.
+#
+# pkg create always names the file <name>-<version>.pkg; every caller
+# wants something more specific, so the rename is part of the helper
+# rather than repeated after it.
+pkg_create_package() {
+    pkg_cp_stage="$1"
+    pkg_cp_deploy="$2"
+    pkg_cp_version="$3"
+    pkg_cp_out="$4"
+
+    mkdir -p "$pkg_cp_deploy"
+    echo "==> pkg create"
+    pkg create -M "${pkg_cp_stage}/+MANIFEST" -p "${pkg_cp_stage}/pkg-plist" \
+        -r "$pkg_cp_stage" -o "$pkg_cp_deploy" || return $?
+    mv "${pkg_cp_deploy}/fips-${pkg_cp_version}.pkg" "$pkg_cp_out" || return $?
+
+    echo "==> built:"
+    ls -l "$pkg_cp_out"
+}
+
+# Emit the manifest fields both packages agree on, in the order
+# pkg-create(8) expects: $1 version, $2 ABI, $3 comment, $4 description.
+#
+# Shared because they are shared *policy*, not merely duplicated text —
+# origin, maintainer, licence and prefix describe one project, and two
+# copies is two things to forget to update. Anything a package decides
+# for itself (annotations, the scripts block) is appended by the caller
+# after this.
+pkg_manifest_header() {
+    cat <<EOS
+name: "fips"
+version: "$1"
+origin: "net/fips"
+comment: "$3"
+desc: <<EOD
+$4
+EOD
+maintainer: "johnathan@corganlabs.com"
+www: "https://fips.network"
+abi: "$2"
+prefix: "/usr/local"
+licenselogic: "single"
+licenses: ["MIT"]
+categories: ["net"]
+EOS
+}
+
+# The post-install lines creating the control-socket access group. Both
+# packages create /var/run/fips as root:fips 0750 from their boot
+# script, so members of this group can use fipsctl and fipstop without
+# root; the group has to exist before either runs.
+pkg_group_script() {
+    cat <<'EOS'
+pw groupshow fips >/dev/null 2>&1 || pw groupadd fips
+EOS
+}
+
+# Build the release binaries. $1 project root, $2 optional Rust target
+# triple — passing one keeps RUSTFLAGS off build scripts and proc-macros,
+# and puts the output under target/<triple>/release.
+pkg_cargo_build() {
+    if [ -n "${2:-}" ]; then
+        echo "==> cargo build --release --target $2"
+        (cd "$1" && cargo build --release --target "$2")
+    else
+        echo "==> cargo build --release"
+        (cd "$1" && cargo build --release)
+    fi
+}
--- a/packaging/freebsd/README.md
+++ b/packaging/freebsd/README.md
@@ -4,6 +4,11 @@ Builds a native FreeBSD `.pkg` shipping `fips`, `fipsctl`, `fipstop`,
 rc.d services, and `.fips` DNS integration. `fips-gateway` is excluded
 (its NAT backend is nftables, Linux-only).
 
+**On pfSense?** Use [`packaging/pfsense/`](../pfsense/README.md)
+instead. pfSense is FreeBSD underneath, but it boots packages, wires up
+DNS and handles upgrades differently enough that this package does not
+work there; the pfSense one addresses each difference.
+
 Platform notes: the Ethernet and BLE transports are not available on
 FreeBSD (UDP, TCP, Tor, and Nym are). The UDP datapath deliberately
 uses the portable single-packet receive loop — FreeBSD's `recvmmsg(2)`
--- a/packaging/freebsd/build-pkg.sh
+++ b/packaging/freebsd/build-pkg.sh
@@ -14,6 +14,10 @@ set -eu
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 PROJECT_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
 
+# Mechanics shared with the pfSense builder: version derivation, the
+# stage layout, the @sample manifest scripts, and pkg create.
+. "${PROJECT_ROOT}/packaging/common/pkg-lib.sh"
+
 NO_BUILD=0
 VERSION=""
 while [ $# -gt 0 ]; do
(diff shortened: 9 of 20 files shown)


jmcorgan:
Thanks for this. It is a substantial and careful package, and the review found nothing wrong with the approach: the `pkg-lib.sh` extraction leaves the FreeBSD package byte-identical (checked by building both trees against a stubbed `pkg`, and your CI smoke-install agrees), the pfSense premises match CE source where they could be checked, and the config.xml edit is idempotent. Because a pfSense package is a new support surface for the project, I would like everything below addressed in this pull request rather than as follow-ups. The branch also needs a rebase onto current master, which conflicts in `CHANGELOG.md` only.

**Release and CI**

1. The next `v*` tag will publish the CE 2.8.1 amd64 package automatically, and the [README](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/README.md#L457-L463) says neither Intel package has been installed. Please install that exact artifact on a CE 2.8.1 amd64 box and report the result, or leave it out of the release upload until someone has.
2. The pfSense build and checker run inside the FreeBSD step, after the smoke-install and before any SHA-256 sidecar is written ([package-freebsd.yml](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/.github/workflows/package-freebsd.yml#L215-L232)). A pfSense-only failure, such as a new key in the common `dns:` block or a dependency that cannot link statically, would therefore block the FreeBSD artifact and its release asset. Please decouple them: write the FreeBSD sidecar first and move the pfSense build and check into their own step or job.
3. Please name a contact for pfSense reports in `packaging/pfsense/README.md`. The manifest maintainer comes from the shared header, so reports will reach the project, and the ABI-to-product table has to track Netgate's releases. Would you be willing to be that contact?

**The config.xml path**

4. `fips-unbound-custom.php` writes the options and restarts unbound without the config test pfSense's own GUI runs ([L146-L166](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/fips-unbound-custom.php#L146-L166)). If the merged options fail to parse, every client behind the firewall loses DNS. Please call `test_unbound_config()` on the would-be config before `write_config`, guarded with `function_exists` like the other accessors, and exit 1 with its output on failure, as the FreeBSD script does with `local-unbound-checkconf`.
5. Pre-deinstall runs `fips-dns-teardown 2>/dev/null || true` ([build-pkg.sh](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/build-pkg.sh#L554-L556)), and the helper reports its refusals on stderr, so `pkg delete` can leave `fips.` forwarded to a dead port and print nothing. Please keep `|| true`, drop the redirect, and on failure print a line pointing at Services > DNS Resolver > Custom options.
6. The readiness check greps sockstat for `${BIND_ADDR}:${DNS_PORT}` ([fips-dns-setup](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/fips-dns-setup#L110-L113)), but sockstat prints IPv6 as `[::1]:5354`, and `::1` is the fallback when `bind_addr` is absent ([L77-L79](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/fips-dns-setup#L77-L79)). A `::1` bind waits 30 seconds and reports nothing listening while the responder is up. Please accept the bracketed form and refuse a wildcard `bind_addr` with a clear message, since `forward-addr` needs a specific address.
7. When the DNS Resolver is disabled the helper exits 0 after writing an inactive block ([L152-L160](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/fips-unbound-custom.php#L152-L160)), so setup still prints "Done. Verify with". Please return a distinct exit code there and have setup skip that line, and have the dnsmasq hint print the real address and port.
8. `rtrim` on the operator's text ([L88-L90](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/fips-unbound-custom.php#L88-L90), [L128-L130](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/fips-unbound-custom.php#L128-L130)) drops trailing newlines on an add followed by a remove, while the [README](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/README.md#L355-L358) promises the text is left byte-for-byte alone. Please preserve the surrounding text exactly.
9. The snippet ends inside `forward-zone:`, and custom options are spliced in last, so a server option an operator later adds below `# END FIPS` lands inside that clause. Please reopen `server:` at the end of the snippet.

**Boot script and upgrades**

10. The `fips.sh` rationale says rc.subr's rcvar has to live in `/etc/rc.conf` ([L17-L23](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/fips.sh#L17-L23)), but `load_rc_config` also reads `rc.conf.d/<name>` and a script can default its rcvar to YES. What pfSense forces is the `.sh` name and a quiet exit 0 when start is re-run. Please reword the comment and the README to state those constraints.
11. On FreeBSD's `pidfile_read`, a locked but empty pidfile reads as pid -1, so daemon(8) may print `process already running, pid: -1`; the `sed` in the orphaned-supervisor recovery accepts only digits ([L151-L162](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/fips.sh#L151-L162)), so every later start would fail with the generic error. Please try it (`lockf -k -t 0 /var/run/fips/fips.pid sleep 600 &`, then `fips.sh start`) and either handle -1 by reading the `-P` file and confirming `comm=daemon`, or narrow the comment at [L208-L219](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/fips.sh#L208-L219) to the case that is handled.
12. The upgrade handling relies on pkg running the old pre-deinstall and the new post-install with `PKG_UPGRADE=true` ([L517-L523](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/build-pkg.sh#L517-L523)). From pkg's source, `pkg add ./newer.pkg` does neither, so the daemon keeps running the old binary, and the README documents `pkg upgrade`, which needs a repository these packages are not in. Please document the command people will actually run, followed by `fips.sh restart`, or test whether `pkg install ./file.pkg` takes the upgrade path and document that.
13. The banner, `pkg-descr` and the README's [upgrade section](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/README.md#L406-L411) say a firmware upgrade removes the package, while the [static-linking section](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/README.md#L241-L245) says a static package survives. CE's `pfSense-upgrade` appears to remove only `pfSense-pkg-*` packages. Please run one CE firmware upgrade and make all three agree with what happens.
14. The hardware run started the boot script by hand. Please cover a reboot and an interface bounce, which are the reasons this script exists.
15. `/var/log/fips.log` has no newsyslog entry, and `/var` can be a RAM disk on pfSense. Please add a `newsyslog.conf.d` entry.

**The checker**

16. None of the boot-script cases can catch a pid check that never succeeds, which is the defect the hardware run found, and no case runs start, stop or restart although the [header](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/testing/check-pfsense-pkg.sh#L11-L18) lists idempotent start. Five deliberate breaks stayed green: `running_pid` never succeeding, the original `read -r _pid < "$pidfile" || return 1` restored, the checker's pidfile `sed` no longer matching, the teardown call removed from pre-deinstall, and fips.yaml seeded at 0644. Please add a positive case (a copied `/bin/sleep` started as `fips`, its pid written with no trailing newline, asserting "running as pid"), assert the `sed` took effect, and run start twice, stop and restart against that stub in the VM.
17. The teardown check greps the whole `+MANIFEST` ([L242-L249](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/testing/check-pfsense-pkg.sh#L242-L249)), whose file list always contains that path. Please match the invocation inside the pre-deinstall script instead, and add a check for the `install -m 0600` that seeds fips.yaml.
18. The dns reader test ([L381-L395](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/testing/check-pfsense-pkg.sh#L381-L395)) puts `dns:` before `transports:`, so a first-match grep also passes; please add a fixture with `transports:` first. The skipped non-pfSense check counts as a pass ([L369-L372](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/testing/check-pfsense-pkg.sh#L369-L372)); please print it as SKIP and leave it out of the count.
19. Nothing runs the PHP helper. Please add `php -l` on it in the FreeBSD VM and a unit test of `fips_strip_block`, which has no side effects.

**The shared builder**

20. A failed `pkg create` now exits 1 instead of pkg's own status ([pkg-lib.sh](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/common/pkg-lib.sh#L117-L119)); please use `|| return $?`. The pfSense builder repeats the ABI fallback inline ([L71-L73](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/build-pkg.sh#L71-L73)); please call `pkg_host_abi`.

**Security notes for operators**

21. Please add a sentence each to the README: the sample transports bind 0.0.0.0 on every interface, WAN included, so keeping them off the internet relies on pfSense's default of no WAN pass rules; and `do-not-query-localhost: no` applies to the whole resolver, not only `fips.`.

**Stale text, description and changelog**

22. The rename to ABI-to-products left old text behind: the superseded product table comment above its replacement ([L76-L95](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/build-pkg.sh#L76-L95)); the output pattern `-pfsense-freebsd<major>-<arch>.pkg` in [build-pkg.sh](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/build-pkg.sh#L11-L13) and the [README](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/README.md#L217-L219), where the code builds `-pfsense-<products>-<arch>.pkg`; CE 2.9 missing from the ABI table at [L239-L243](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/build-pkg.sh#L239-L243); the "Neither Intel package" paragraph naming CE 2.8 and Plus 26.x amd64 and promising six defects where it describes three kinds; and `--static` accepted "so older invocations keep working" in a new script.
23. The description says the two builders share no code, while the diff extracts `pkg-lib.sh` and changes the builder of a released artifact. It also mentions three fix commits (there is one), says both Intel packages are built and checked in CI (only CE 2.8.1 amd64 is), and says CE 2.9 and Plus 26.x cannot install each other's package while the README says they share one. The commit message says setup refuses to touch config.xml "unless the responder answers", but the code checks only that something is listening. Please bring both up to date.
24. On the third silent failure: pfSense does make `do-ip6` follow Allow IPv6, but Allow IPv6 is on in the factory config, and your README and [fips.yaml.dns](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/packaging/pfsense/fips.yaml.dns#L10-L19) say the mesh cannot pass traffic with it off. So the 127.0.0.1 bind is a robustness choice rather than a fix for a failure users would hit. Please say so in the description and the changelog.
25. The `CHANGELOG.md` entry is one bullet of about 105 lines, repeating README rationale. Please cut it to 15 to 20 lines: what the package is, the three differences from the FreeBSD package, static linking by default, the single asset the release publishes and which pfSense it is for, and a pointer to the README. It also says aarch64 is not published and "only amd64 is" ([L107-L109](https://github.com/jmcorgan/fips/blob/87ed84a29f8970f020dac9e9cdd833c4db4aad01/CHANGELOG.md#L107-L109)); only the FreeBSD 15 amd64 package for CE 2.8.1 is. When you rebase, move `#### Packaging` under master's existing `### Added`, after `#### Library surface and internals`.
26. A pfSense user will likely start from the FreeBSD instructions. Please add a line to `packaging/freebsd/README.md`, to the FreeBSD section of `docs/getting-started.md`, and to the packaging line in the top-level README, pointing at `packaging/pfsense/`.

Happy to answer questions on any of these before you start.

fr34aky:
Thank you for the thorough review — every item is addressed in the force-pushed revision (now one commit, rebased onto current master, CHANGELOG the only conflict). Item-by-item below; where I could verify a claim against source or a test I say so, and I have been careful not to report the two I could not actually run as done.

**Release and CI**

1. I cannot obtain CE 2.8.1 to install the exact artifact: the public mirror stops at 2.7.2, and the Netgate installer only lays down the current release (2.9.0). Taking your fallback — the CE 2.8.1 package is built and checked in CI but **held back from the release upload** (a new release step moves `*-pfsense-*.pkg` aside before publishing); it stays a 30-day workflow artifact. It can go into releases once someone validates it on CE 2.8.1 hardware.
2. Done. The pfSense build and check run in their own `continue-on-error` step after the FreeBSD sidecar is written; the release asset resolution treats the pfSense package as optional, so a pfSense-only failure (a new `dns:` key, a dependency that won't link statically) no longer blocks the FreeBSD artifact.
3. Done — happy to be the contact; named in `packaging/pfsense/README.md` under "Maintenance and reports".

**The config.xml path**

4. Done, and run on the CE-equivalent box (Plus 26.03.1 aarch64): `test_unbound_config()` (guarded with `function_exists`) runs on the merged options before `write_config`; a valid re-apply passes and writes, a deliberately broken snippet is refused with config.xml untouched and DNS still up. The hardware run caught a real bug here — I had the `test_unbound_config` return code inverted (it returns unbound-checkconf's exit code, 0 = valid, and pfSense core treats non-zero as failure), so it would have rejected every valid config. Fixed to match core; a FreeBSD CI host cannot catch this, as it has no `test_unbound_config`.
5. Done. `fips-dns-teardown` keeps `|| true` but the redirect is gone, and on failure pre-deinstall prints a line pointing at Services > DNS Resolver > Custom options.
6. Done. Readiness now matches the bracketed `[::1]:5354` form sockstat prints, and a wildcard `bind_addr` is refused with a message, since `forward-addr` needs a specific address.
7. Done. The helper returns exit 3 when the Resolver is disabled; setup skips the "Done. Verify with" line on that code, and the dnsmasq hint prints the real address and port.
8. Done. Block insertion/removal is now byte-exact (a dedicated `fips_add_block`/`fips_strip_block` pair that cut exactly what was inserted); a unit test confirms add-then-remove round-trips several fixtures byte-for-byte.
9. Done. The snippet reopens `server:` at the end, so a server option added below `# END FIPS` no longer lands inside `forward-zone:`.

**Boot script and upgrades**

10. Done. The comment and README now say the real constraints — the `.sh` name and a quiet idempotent start — and note `rc.subr` would work (reads `rc.conf.d`, can default the rcvar); plain sh is a choice to keep the re-run behaviour visible.
11. Done, and reproduced both ways (`lockf -k -t 0 .../fips.pid sleep 600` then `fips.sh start`): daemon(8) prints `pid: -1`, so recovery no longer parses that — it asks `fstat` who holds the pidfile, confirms `comm=daemon`, and reaps it; a non-daemon holder is refused untouched.
12. Corrected. Measured on a throwaway package: `pkg add` (±`-f`) runs neither upgrade script (`PKG_UPGRADE` unset); `pkg install ./file.pkg` runs the upgrade path with `PKG_UPGRADE=true`; `pkg upgrade` needs a repo. README and banner now document `pkg install ./<file>.pkg` followed by `fips.sh restart`.
13. From `pfSense-upgrade` source, a firmware upgrade removes only `pfSense-pkg-*`, so `fips` (a plain pkg) survives; banner, pkg-descr and README now agree on that, with the caveat that a FreeBSD-15 static binary runs on a FreeBSD-16 kernel only via compat, so after a major upgrade reinstall the package for the new base. I could not run a live CE upgrade (no obtainable 2.8.1 to upgrade from); this is marked as source-derived, not hardware-confirmed.
14. Done on hardware (Plus 26.03.1 aarch64): reboot — one automatic start, one supervisor, clean shutdown; WAN-VLAN flap — daemon survives (a static-IP link event does not reach rc.newwanip); `pfSctl -c 'service reload packages'` (the rc.newwanip path) — re-entrant start is a no-op, `fips: already running` in bootup_messages.
15. Done. Ships `fips.newsyslog`; `fips.sh` copies it into `/var/etc/newsyslog.conf.d/` on start, since pfSense regenerates `/etc/newsyslog.conf` and rebuilds `/var/etc` at boot on RAM-disk systems.

**The checker**

16. Done and proven. Added a positive lifecycle case (a copied `/bin/sleep` started as `fips`, pid written with no trailing newline) asserting "running as pid", then start-twice/stop/restart; the harness rewrites are asserted to have taken effect. Re-injecting the original `read ... || return 1` now trips four of these checks where the old checker stayed green.
17. Done. The teardown check greps the pre-deinstall script body, not the whole manifest; added a check for the `install -m 0600` fips.yaml seed.
18. Done. A `transports:`-first fixture confirms the reader is not rescued by ordering; the non-pfSense check is now SKIP and not counted.
19. Done. The FreeBSD-VM step runs `php -l` on the helper and a unit test of `fips_strip_block` (via `FIPS_UNBOUND_HELPER_TEST`, which defines the functions without pfSense's includes).

**The shared builder**

20. Done. `pkg_create_package` uses `|| return $?`; the pfSense builder calls `pkg_host_abi` instead of the inline fallback.

**Security notes**

21. Done. README now states the sample transports bind `0.0.0.0` (WAN included, relying on pfSense's default no-WAN-pass), and that `do-not-query-localhost: no` is resolver-wide.

**Stale text, description and changelog**

22. Done — superseded product-table comment removed, output pattern is `-pfsense-<products>-<arch>.pkg`, CE 2.9 added to the ABI table, the "Neither Intel / six defects" paragraph reconciled, `--static` comment no longer says "older invocations".
23. Done. The description now says the builders share `pkg-lib.sh` (with the byte-identical acceptance test), one commit not three, CI builds/holds the CE 2.8.1 package, and CE 2.9 / Plus 26 share one artifact. The commit message no longer claims setup checks "unless the responder answers" — it says it waits for a listener and validates the merged config.
24. Done. README, `fips.yaml.dns` and the changelog now say "Allow IPv6" is on by default, so the 127.0.0.1 bind is robustness against the `do-ip6: no` case rather than a fix for a failure a default install hits.
25. Done. The changelog entry is ~18 lines under master's existing `### Added` (after "Library surface and internals"), and no longer says "only amd64 is".
26. Done. Pointers to `packaging/pfsense/` added to `packaging/freebsd/README.md`, the FreeBSD section of `docs/getting-started.md`, and the top-level README.

Two things I could not run, stated plainly rather than claimed: a real CE 2.8.1 install (item 1) and a live firmware upgrade (item 13's empirical half). Separately, and as its own thread rather than folded in here: now that every fresh install is 2.9.0 / 26.x on FreeBSD 16, it may be worth targeting the FreeBSD 16 package in CI — which needs a FreeBSD 16 amd64 builder on a moving 16.0-CURRENT snapshot. The amd64 FreeBSD 16 package already builds static and pin-honoured, and FreeBSD 16 is the base the aarch64 package has been running on. Happy to raise that separately if you're interested.

jmcorgan:
Thanks for the thorough pass. I went through the new head (a8a71b90) item by item against the 26 points. Most are cleanly done, and the hardware work on items 4, 11, 12 and 14 is appreciated, particularly catching the inverted `test_unbound_config` return code; I confirmed against pfSense's `unbound.inc` and `services_unbound.php` that non-zero is the failure. Three things need another round before this can land, plus some text that the item 1 change left inconsistent.

**1. The new lifecycle checks fail in CI, and the job still reports green (items 2 and 16)**

In the FreeBSD Package run on this head ([run 34586486901](https://github.com/jmcorgan/fips/actions/runs/34586486901)), `check-pfsense-pkg.sh` ends with `pfSense package checks FAILED (5 of 58)`. Start prints `fips: started` but no pidfile is ever written, and then status, re-entrant start, restart and stop all fail. I did not dig into why the stub harness behaves differently on the CI VM than wherever you ran it. Because the step has `continue-on-error: true` ([package-freebsd.yml L214-L223](https://github.com/jmcorgan/fips/blob/a8a71b901a7f73174916a7ce767d28aa04b490f0/.github/workflows/package-freebsd.yml#L214-L223)), the step and the job both report success, and the only trace is a warning annotation. So a pfSense failure now can never show red, which is a bigger problem than the coupling item 2 was about.

Please move the pfSense build and checker into their own job that the release job does not `needs`, so a failure reds its own check without blocking the FreeBSD asset, and then get the lifecycle cases passing there.

A smaller point in the same block: `pidfile has no trailing newline` passed in that run even though the pidfile did not exist, because `tail` on a missing file prints nothing ([check-pfsense-pkg.sh L341-L351](https://github.com/jmcorgan/fips/blob/a8a71b901a7f73174916a7ce767d28aa04b490f0/testing/check-pfsense-pkg.sh#L341-L351)). Please require `-s "$WORK/test.pid"` before that test.

**2. Byte-exact removal joins two operator lines (item 8)**

`fips_strip_block` now always removes the newline before `BEGIN` and the one after `END` ([fips-unbound-custom.php L90-L100](https://github.com/jmcorgan/fips/blob/a8a71b901a7f73174916a7ce767d28aa04b490f0/packaging/pfsense/fips-unbound-custom.php#L90-L100)). That round-trips your fixtures, but not the case item 9 anticipates, where an operator adds options below `# END FIPS`. Running the PR's own `fips_add_block` and `fips_strip_block` under PHP 8.3: start from operator text `# my local tweaks` (no trailing newline), add the block, append `private-domain: "lan"` below it, then strip. The result is `# my local tweaksprivate-domain: "lan"`. The option has become part of the comment, so it passes `unbound-checkconf` and is silently dropped on the next setup re-run or on `pkg delete`. The previous `rtrim` version gave `# my local tweaks\nprivate-domain: "lan"` for the same input.

Suggested fix: only consume the newline before `BEGIN` when nothing follows the block, for example:

```php
$stop = $end + strlen(FIPS_END);
if (substr($text, $stop, 1) === "\n") {
	$stop += 1;
}
$tail = substr($text, $stop);
$start = ($tail === '' && $begin > 0 && $text[$begin - 1] === "\n") ? $begin - 1 : $begin;
return substr($text, 0, $start) . $tail;
```

Under PHP 8.3 that passes all five of your fixtures and the case above. Please also add a text-after-the-block fixture to the unit test.

**3. The newsyslog entry rotates a log daemon(8) never reopens (item 15)**

`fips.newsyslog` has no pid file or signal, on the reasoning that daemon(8) "writes with O_APPEND, so rotation by rename is safe" ([fips.newsyslog](https://github.com/jmcorgan/fips/blob/a8a71b901a7f73174916a7ce767d28aa04b490f0/packaging/pfsense/fips.newsyslog#L1-L11)). `O_APPEND` does not change which file a descriptor points at. newsyslog's `savelog` hard-links the log to `fips.log.0` and `createlog` renames a fresh file into place, so daemon(8) keeps writing to the old inode. Once the `J` flag compresses and unlinks `fips.log.0`, the daemon is writing into a deleted file: `/var/log/fips.log` stays empty and the space is never freed, which on a RAM-disk `/var` is the exact problem the item was meant to prevent.

FreeBSD's daemon(8) has `-H` ("Close and re-open output file on SIGHUP") for this. Please add `-H` to the daemon(8) invocation in [fips.sh L128-L133](https://github.com/jmcorgan/fips/blob/a8a71b901a7f73174916a7ce767d28aa04b490f0/packaging/pfsense/fips.sh#L128-L133), give the newsyslog entry `/var/run/fips/daemon.pid` as its pid file (the default signal is SIGHUP), and correct the comment. The include location itself is right; pfSense's `syslog.inc` includes `/var/etc/newsyslog.conf.d/*`.

**Text still to reconcile**

- Items 1 and 25: the release job now holds the pfSense package back, but the CHANGELOG entry ("A tagged release publishes one pfSense asset", [CHANGELOG.md L140-L160](https://github.com/jmcorgan/fips/blob/a8a71b901a7f73174916a7ce767d28aa04b490f0/CHANGELOG.md#L140-L160)), the commit message, the README table ("yes, from CI", [README.md L113-L117](https://github.com/jmcorgan/fips/blob/a8a71b901a7f73174916a7ce767d28aa04b490f0/packaging/pfsense/README.md#L113-L117)), and the workflow comments on the pfSense step and the asset-resolution step ("published with the rest") still say it is published.
- Item 24: the README, `fips.yaml.dns` and the banner now say Allow IPv6 is on by default, but the CHANGELOG entry and the commit message still list the `::1` bind as one of the three silent failures without that caveat.
- Item 23: the PR description's Naming section still says CE 2.9 and Plus 26.x on Intel are "neither installable on the other"; its boot table still gives `/etc/rc.conf` being replaced as the reason, which item 10 withdrew; and its idempotency paragraph still says the glob re-runs on link changes, which your item 14 test showed it does not. The commit message also states firmware-upgrade survival as fact, where the README correctly marks it as source-derived.

Items 3, 4, 5, 6, 7, 9, 10, 12, 14, 17 to 22 and 26 look done to me, and 1, 11 and 13 are fine as resolved given that CE 2.8.1 cannot be obtained. The FreeBSD 16 builder question is worth its own thread when you are ready.

fr34aky:
Thanks — all three blockers and the text are addressed in the new head. I reproduced the CI failure before fixing it: this dev box is a Linux/FreeBSD hybrid where the checker only fails as root, so I booted a real FreeBSD VM, ran the checker as root, and got your exact `FAILED (5 of 58)`. Each fix below is verified there, as root, not just locally.

**1. CI green-over-failure, and the lifecycle checks (items 2 and 16)**

Root cause of the failing lifecycle: the checker wrapped `start` in `timeout`, and `timeout` runs the script in its own process group whose teardown reaps the daemon(8) child before it detaches — so `start` printed "started" but no pidfile. `timeout --foreground` (no separate process group) fixes it; the daemon survives and all lifecycle cases pass. Confirmed by re-injecting the original `read ... || return 1`, which now trips four checks. The no-trailing-newline test now requires `-s "$WORK/test.pid"` first, so it can't pass on a missing file.

On the structural point, which was the real problem: the pfSense build and checker are now their own job (`pfsense`), which `release` does not `needs` and which has no `continue-on-error`. A pfSense failure reds that job on its own and cannot hide behind the FreeBSD job or block the FreeBSD release asset. As a consequence the pfSense package is never a release asset — it is a workflow artifact from that job — which also settles item 1 without the separate hold-back step.

**2. Byte-exact removal joining two lines (item 8)**

Fixed with your suggested logic: consume the trailing newline always, the leading one only when nothing follows the block. Your `# my local tweaks` + append-below case now yields `# my local tweaks\nprivate-domain: "lan"`, and the five fixtures still round-trip. Added a text-after-the-block fixture to the unit test, which the CI job runs under php.

**3. newsyslog rotating a log daemon(8) never reopens (item 15)**

You are right that O_APPEND does not change the inode. The daemon(8) invocation now has `-H`, and the newsyslog entry names `/var/run/fips/daemon.pid` (the supervisor's `-P` file), so newsyslog sends SIGHUP and daemon(8) reopens the rotated-away log. Comment corrected.

**Text reconciled**

- Items 1/25: the CHANGELOG, README table, and commit message now say the pfSense package is a CI workflow artifact, not a release asset; the workflow no longer has the "published with the rest" wording.
- Item 24: the CHANGELOG and commit message now state that "Allow IPv6" is on by default and the 127.0.0.1 bind is robustness for the case where it has been turned off, not a failure a default install hits.
- Item 23: the description's Naming section no longer says the two amd64 packages are "neither installable on the other" (it notes CE 2.9 and Plus 26.x Intel share one artifact); the boot table no longer gives `/etc/rc.conf` replacement as the reason; the idempotency paragraph now says the glob re-runs on a WAN IP change, not on link changes; and the commit message marks firmware-upgrade survival as source-derived, not fact.

**Item 13, now partly on hardware.** A live **Plus 26.03.1 -> 26.07** upgrade (aarch64) just ran: the `fips` package survived (`pkg info` still lists it), the daemon restarted at boot on the new base, the dynamic binary still linked (`ldd` clean across the within-16.0-CURRENT move), `.fips` still resolved, and the DNS Resolver block was still in `config.xml`. That confirms the package-survival and auto-restart halves of item 13 on real hardware — for a *minor*, FreeBSD 16 -> 16 change. The *cross-major* compat case (FreeBSD 15 -> 16) is still only source-reasoned; the README says so.

Still genuinely unrun: a real CE 2.8.1 install and a cross-major firmware upgrade — CE 2.8.1 cannot be obtained (the mirror stops at 2.7.2, and the Netgate installer only lays down the current release). The FreeBSD 16 builder is a separate thread whenever you want it.

fr34aky:
The FreeBSD 16 builder question is now its own thread, as suggested: #157.

jmcorgan:
Thanks, this round is close. I checked each of the three fixes rather than taking them on trust, and all three hold:

- **CI (items 2 and 16).** The new `pfsense` job has no `continue-on-error`, and its log on this head ends `pfSense package checks PASSED (58 checks, 0 skipped)`, with every lifecycle case passing. The `-s` guard on the pidfile check is in.
- **Block removal (item 8).** Under PHP 8.3 I ran the previous and the new `fips_strip_block` against four kinds of operator text: empty, one line without a trailing newline, one line with one, and two lines. Each went through add then strip, a line appended below the block, and a setup re-run. The previous head joins lines in two of those cases; this head passes all twelve. Your new fixture fails against the old code, so it guards the right thing.
- **newsyslog (item 15).** `-H` is on the daemon(8) call, and the entry names `/var/run/fips/daemon.pid`, which is the `-P` file.

One problem blocks, and I want to be clear that it is new in this push rather than something the earlier reviews missed. At the previous head (a8a71b90), the release job had a step that moved any pfSense package out of `dist/` before publishing, so this path did not exist then. Moving the pfSense build into its own job removed that step, and the problem comes from that change.

**The pfSense package would be published on a tag.** The release job downloads artifacts with no `name` or `pattern` ([package-freebsd.yml L359-L365](https://github.com/jmcorgan/fips/blob/537ea72ca14c46b0b98bbc3995c72c7148d9d00b/.github/workflows/package-freebsd.yml#L359-L365)), and `actions/download-artifact` downloads every artifact in the run when `name` is not given. The pfSense package carries its `.sha256` sidecar, so it passes the pre-publish check, and `gh release upload dist/*.pkg` ([L431-L437](https://github.com/jmcorgan/fips/blob/537ea72ca14c46b0b98bbc3995c72c7148d9d00b/.github/workflows/package-freebsd.yml#L431-L437)) attaches it. The `release` job does not `needs: pfsense`, but that only means it does not wait for it. In this run the pfSense job finished at 10:09 and the FreeBSD build at 10:15, so on a tag the pfSense artifact would normally be there when `release` starts. I have not pushed a tag to show it, but nothing in the workflow prevents it. Suggested fix: download only the FreeBSD artifact, with `pattern: fips_*_x86_64_freebsd` on that step. Also make the validate step fail if anything matching `*-pfsense-*` is in `dist/`, so the rule survives a later rename.

Smaller things, none blocking on its own:

- **The `build` job still carries the previous layout.** It raises `timeout-minutes` to 45 for "a second full release build in this job" ([L67-L73](https://github.com/jmcorgan/fips/blob/537ea72ca14c46b0b98bbc3995c72c7148d9d00b/.github/workflows/package-freebsd.yml#L67-L73)). It installs `bash php85` for a checker that no longer runs there ([L92-L99](https://github.com/jmcorgan/fips/blob/537ea72ca14c46b0b98bbc3995c72c7148d9d00b/.github/workflows/package-freebsd.yml#L92-L99)). It also keeps the comment about which pfSense package is "built or published here" ([L175-L196](https://github.com/jmcorgan/fips/blob/537ea72ca14c46b0b98bbc3995c72c7148d9d00b/.github/workflows/package-freebsd.yml#L175-L196)). None of that applies to that job now. Please put the timeout and the `pkg install` line back to what `master` has, and move the comment to the `pfsense` job or drop it.
- **The `--foreground` comment names the wrong mechanism** ([check-pfsense-pkg.sh L334-L340](https://github.com/jmcorgan/fips/blob/537ea72ca14c46b0b98bbc3995c72c7148d9d00b/testing/check-pfsense-pkg.sh#L334-L340)). Without `--foreground`, FreeBSD's `timeout` makes itself a reaper for everything the script spawns. After the script exits, it keeps waiting until all of those descendants have exited too, and at the time limit it kills them all (`bin/timeout/timeout.c`). daemon(8) never exits, so the start call sits for the full 30 seconds and then the daemon is killed. No process group is involved. The fix is right; the comment should describe this, since it is what the next person to touch the checker will read.
- **Text.** The README's list of still-unexercised paths includes `fips-dns-teardown` ([README.md L506-L511](https://github.com/jmcorgan/fips/blob/537ea72ca14c46b0b98bbc3995c72c7148d9d00b/packaging/pfsense/README.md#L506-L511)), while the PR description says teardown was run on the box and restored `custom_options` byte for byte; one of them needs to change. In the description, the CI section still says the package is built "in the existing FreeBSD VM job" with the timeout raised, and "A tagged run publishes **one** pfSense package" should go. The commit message has one 97-character line.

With the release download fixed and the `build` job put back, I expect this can land.

fr34aky:
Thanks — and you are right that the release path was new in my last push, not something earlier reviews missed. Moving the pfSense build into its own job removed the hold-back step, and I reasoned that `release` "never sees" the artifact because it does not `needs: pfsense`. That was wrong: `needs` only orders jobs, and `actions/download-artifact` with no `pattern` downloads every artifact in the run. Fixed on the new head, and every item below is verified rather than assumed.

**The blocker — pfSense package reaching the release**

Two layers, as you suggested. The download step is now scoped with `pattern: fips_*_x86_64_freebsd`, so the pfSense artifact is never fetched into `dist/`. And the validate step fails outright if anything matching `*-pfsense-*` is in `dist/` before publishing, so the rule survives an artifact rename or a new artifact later. One honest limit: the release job is tag-gated, so no branch push exercises this — I have verified it by reading (`pattern` and `merge-multiple` are inputs of the pinned `download-artifact` commit, and the guard is plain bash), not by a run; a tag is the only thing that would. The description's CI section now says so, and "A tagged run publishes one pfSense package" is gone — no pfSense package is a release asset.

**`build` job put back**

The `build` job is now byte-identical to master — not just the timeout and the `pkg install` line, but the whole job. On checking I found my earlier rounds had also left a stale `prepare` comment (explaining a `bash` install that no longer happens), a dangling "see the ARM note further down", a sidecar loop for a second package that no longer exists, and an `if`-to-`||` compaction that was pure churn; all of it is gone with the job restored from master. The one part of the removed comment still worth keeping — this VM yields the CE 2.8.1 package, CE 2.9/Plus 26.x need a FreeBSD 16 host, and no aarch64 package is built or published because rustup ships no toolchain for it so the pin cannot be honoured — now lives in the `pfsense` job's header, where it applies. The workflow's whole diff against master is exactly two things: the added `pfsense` job, and the release job's download `pattern` plus the `*-pfsense-*` guard.

**`--foreground` comment**

Rewritten to the real mechanism, which I confirmed in `timeout.c` before writing it: without `--foreground`, timeout acquires a reaper (`PROC_REAP_ACQUIRE`), waits after the script exits until every descendant has exited, and at the limit kills them all (`PROC_REAP_KILL`) — daemon(8) never exits, so start sat the full 30 s and then the daemon was killed. No process group involved. The fix is unchanged; only the explanation is.

**Text**

- README: `fips-dns-teardown` is out of the still-unexercised list, with a note that it was run on the box and restored `custom_options` byte for byte (it was; the description and README now agree).
- Description: CI section rewritten for the separate job and the release scoping; "publishes one" removed.
- Commit message: the 97-character line is rewrapped; the longest line is now 72.

Gates: YAML, action pins, CI parity, image scoping, log strings, trailing-log, shellcheck all pass; the package rebuilds and the checker is 58/58. On this head the FreeBSD Package workflow is green (run 34698774102): the `build` job with master's layout, the separate `pfsense` job, and `release` skipped as a branch push should — which is also why the release-path fix above is verified by reading rather than by that run.

jmcorgan:
Landed on `master` as `429d7773` with your authorship and author date, and merged up to `next` as `4f5d63f4`. It was cherry-picked onto the current `master`, which had moved two commits since your base, and two small fold-ins went in with it, so the commit is not your head and I'm closing this by hand.

The fold-ins:

- In the pfSense README, the teardown note no longer says to see "Upgrades and removal", since that section does not mention the teardown run.
- `packaging/common/pkg-lib.sh` gains a `#!/bin/sh` line. The other sourced shell libraries in the tree carry one, and without it `shellcheck -x -S warning` reports SC2148 on that file.

Apart from those two lines and the base change, the tree is identical to your last head, so its green run covers the code. The release-job scoping for the pfSense artifact only runs on a tag, so the first real exercise of it will be the next tagged release.

Thank you for sticking with this through three rounds, and for the hardware runs, which found real defects no CI host could. The FreeBSD 16 builder stays open in #157.

PR #156, opened 2026-09-12
fix(testing): guard the empty env_args splat in the mesh-lab loop

shaibearary:
(i actually dont know too much about these but it seems fix some broken test on my macbook  and didnt break other stuff... others are all written by claude)

Running `run-loop.sh rekey` on macOS aborts the setup subshell with "env_args[@]: unbound variable", so generate-configs.sh, the rekey inject-config step and `docker compose up -d` never run. The script carries on and runs the suite against a stack that was never started, so the failure surfaces as a bogus rekey failure rather than a setup error -- and setup.log is empty, because the expansion fails before the redirection is applied.

run_rekey_family() declares `local env_args=()` and fills it only for the rekey-accept-off and rekey-outbound-only variants, so the plain `rekey` variant reaches the `env "${env_args[@]}"` call sites with an empty array. Under `set -u` (line 49) bash 3.2 treats an empty array expansion as unbound. Bash 4.4 stopped doing so, which is why Linux and CI never see this and only macOS's frozen bash 3.2 is affected.

Check the element count before splatting, and drop the `env` prefix when there is nothing to set -- `env cmd` with no assignments is equivalent to `cmd`, so both branches behave identically everywhere. The nat-lan call site is left unguarded on purpose: its env_args is declared with an element and can never be empty.

### DIFF
--- a/testing/mesh-lab/run-loop.sh
+++ b/testing/mesh-lab/run-loop.sh
@@ -203,10 +203,17 @@ run_rekey_family() {
 
     (
         cd "$REPO_ROOT" || exit 1
-        env "${env_args[@]}" bash testing/static/scripts/generate-configs.sh "$variant" \
-            >>"$REP_DIR/setup.log" 2>&1
-        env "${env_args[@]}" bash testing/static/scripts/rekey-test.sh inject-config \
-            >>"$REP_DIR/setup.log" 2>&1
+        if [ "${#env_args[@]}" -gt 0 ]; then
+            env "${env_args[@]}" bash testing/static/scripts/generate-configs.sh "$variant" \
+                >>"$REP_DIR/setup.log" 2>&1
+            env "${env_args[@]}" bash testing/static/scripts/rekey-test.sh inject-config \
+                >>"$REP_DIR/setup.log" 2>&1
+        else
+            bash testing/static/scripts/generate-configs.sh "$variant" \
+                >>"$REP_DIR/setup.log" 2>&1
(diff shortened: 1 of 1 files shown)


jmcorgan:
Hello Sherry,

Thanks for your first contribution to FIPS, and for tracking this down. The diagnosis was right: under `set -u`, bash 3.2 treats an empty `"${env_args[@]}"` as unbound, so the plain `rekey` variant skipped its setup and then reported a rekey failure against a stack that never started.

This landed on `maint` as `31ebec62` with your authorship, and is merged up to `master` and `next`. `run-loop.sh` is identical on all three lines and the bug is in released material, so `maint` was the right base, as your first pull request had it. The commit changed on the way in, so GitHub will not show this as merged, and I'm closing it by hand.

Two changes on landing:

- The if/else blocks became `env ${env_args[@]+"${env_args[@]}"} ...` at the same three call sites. That form expands to nothing for an empty array on every bash version, so each command is written once instead of twice.
- The `Co-Authored-By: Claude` trailer is removed. The project does not carry coding-assistant attribution in commit messages (item 3 in PR-REVIEW.md).

Checked by running `run_rekey_family()` itself under bash 3.2 and 5.2, with docker and the suite scripts stubbed: the old code fails the plain variant on 3.2, and the new code passes both variants on both versions.

Using a coding assistant is fine here. CONTRIBUTING.md does ask that you review what it produces and be ready to stand behind it, so when a change is outside what you can check yourself, saying so in the description, as you did, is the right thing to do.

Issue #71, opened 2026-04-16
BLE: xHCI host controller death after bluetooth service restart on Linux


Amperstrand:
## Bug

Restarting `systemctl restart bluetooth` on the Linux node (Lenovo Legion T5, Realtek BT adapter via USB) causes the xHCI host controller to die completely. The kernel reports:

```
xHCI host controller not responding, assume dead
xHC died, removing slot 1
xHCI host controller not responding during save, 5ms remaining
```

All USB devices on that xHCI controller are lost (Bluetooth adapter, keyboard, etc.). Recovery requires a PCI unbind/rebind cycle:

```bash
echo 0000:02:00.0 | sudo tee /sys/bus/pci/drivers/xhci_hcd/unbind
echo 0000:02:00.0 | sudo tee /sys/bus/pci/drivers/xhci_hcd/bind
```

After rebind, `hciconfig hci0` shows the adapter as UP RUNNING, but `bluetoothctl show hci0` says "Controller hci0 not available" until `hciconfig hci0 down && hciconfig hci0 up` is run.

## Impact

This makes BLE reliability testing fragile — every `systemctl restart fips` or `systemctl restart bluetooth` risks losing the entire USB bus. The FIPS service itself is fine (it handles adapter disappearance gracefully), but the hardware requires manual intervention.

## Reproduction

1. `sudo systemctl restart bluetooth`
2. Observe dmesg for xHCI death messages
3. All USB devices disappear

## Environment

- Linux node: Ubuntu on Lenovo Legion T5-26AMR5
- Bluetooth adapter: Realtek Semiconductor Corp. Bluetooth Radio (USB, Bus 001 Device 005)
- xHCI controller: PCI device 0000:02:00.0
- BlueZ: stock Ubuntu
- Kernel: 6.x (Ubuntu default)

## Possible causes

- Realtek BT firmware reinitialization triggers USB reset storm
- xHCI controller firmware bug on hot-plug cycle
- Race between bluetoothd startup and USB descriptor re-enumeration
- Power management issue (USB autosuspend + BT init)

## Workaround

The FIPS Linux node is managed as a systemd service. When BLE restart is needed:
1. `sudo hciconfig hci0 down && sudo hciconfig hci0 up` instead of restarting bluetooth service
2. Then `sudo systemctl restart fips`
3. If xHCI death occurs: PCI unbind/rebind + hciconfig cycle

This does NOT affect normal operation — only the restart cycle.

Amperstrand:
[...]

PR #159, opened 2026-09-13
ci: package the pfSense FreeBSD 16 build from the FreeBSD 15.1 job

### LINKED ISSUE #157: Add a FreeBSD 16 amd64 CI builder for the pfSense CE 2.9 / Plus 26.x package
## Summary

CI can only build the pfSense package for FreeBSD 15 (`FreeBSD:15:amd64`, pfSense CE 2.8.1), because the FreeBSD VM in `package-freebsd.yml` is 15.1. But every fresh pfSense install is now on FreeBSD 16: CE 2.9.0 (August 2026) and Plus 26.x are both built from 16.0-CURRENT. The package those users need — `fips-<version>-pfsense-ce2.9-plus26-amd64.pkg`, `FreeBSD:16:amd64` — builds and checks cleanly, but only outside CI. This issue is to decide whether and how to add a FreeBSD 16 amd64 builder. Raised separately from #153 at the maintainer's suggestion.

## What is already established

- The `FreeBSD:16:amd64` package builds statically with the pinned toolchain (`linkage: static`, `pin_honoured: yes`) on a 16.0-CURRENT amd64 host and passes `testing/check-pfsense-pkg.sh` — same provenance as every published artifact.
- FreeBSD 16 is the base the aarch64 package has run on hardware throughout #153 (Plus 26.03.1), including surviving a live 26.03.1 → 26.07 firmware upgrade.
- Static linking removes the base-library drift that would otherwise make "built on one 16.0-CURRENT snapshot, run on Netgate's" a problem; what remains is the kernel syscall ABI, stable within a major.

So the gap is infrastructure, not the package.

## Why it is not a one-line change

- `vmactions/freebsd-vm` tops out at 15.1. Its builder has 16.0 configurations, but they are shelved until 16.0 stabilises.
- FreeBSD 16.0-RELEASE is scheduled for **December 2027**. Until then 16 exists only as 16.0-CURRENT snapshots, and download.freebsd.org keeps only the two most recent — a snapshot URL expires within weeks.
- So a 16 builder today means booting a CURRENT snapshot image, and pinning it means mirroring the image somewhere durable.

## Options

**A. Wait for 16.0-RELEASE and a vmactions image.** No work now. Cost: ~15 months in which CI publishes only the package that no new install can use; CE 2.9 / Plus 26.x stays build-it-yourself.

**B. qemu on `ubuntu-latest`, booting a pinned 16.0-CURRENT VM image.** A job of its own, in the shape of the `pfsense` job from #153: boot the image, `ssh` in, run `packaging/pfsense/build-pkg.sh` and the checker, upload the package as a workflow artifact; not a dependency of `release`.
  - Needs: one chosen snapshot image mirrored durably with its SHA256 (a `ci-images` release in this repo, or a separate repo), so the build is reproducible and does not 404 in a month. Refreshed deliberately, not on every run.
  - Speed: `/dev/kvm` is present on `ubuntu-latest` in practice — widely used actions depend on it — but nested virtualisation is not officially supported on standard runners; larger runners are the supported path. Without KVM, qemu TCG still works — the whole package has been built that way — at roughly 3–4x the wall-clock. Either way it is its own job and blocks nothing.
  - Publishing: artifact-only until someone installs it on a real CE 2.9 or Plus 26.x Intel box, the same rule #153 applies to the CE 2.8.1 package.

**C. A self-hosted FreeBSD 16 runner.** Most control, and the maintainer's infrastructure to run and keep patched. Probably more than this warrants.

## Recommendation

B, if a CI-built FreeBSD 16 package before 2027 is wanted at all. It reuses the job pattern #153 introduces, keeps the moving snapshot pinned rather than tracked, and stays out of the release path until hardware-validated. Happy to implement it as a follow-up PR once #153 lands, if you want it; equally happy to leave it at A if the cost is not worth it to the project.

fr34aky:
## The FreeBSD 16 pfSense package, from the existing FreeBSD 15.1 job

Closes #157. This replaces the earlier content of this PR (a builder on a pinned 16.0-CURRENT snapshot) with the relabel discussed and agreed on #157; the previous head, d507f29d, stays in the PR history for reference.

Every fresh pfSense install is on FreeBSD 16 (CE 2.9.0, Plus 26.x), and CE 2.8.1 can no longer be installed, but CI built only the `FreeBSD:15:amd64` package. The `pfsense` job now packages its one 15.1 build twice: after the ordinary build it runs `build-pkg.sh --no-build --abi FreeBSD:16:amd64`, checks both packages with `check-pfsense-pkg.sh`, install-smokes the FreeBSD 15 one, and uploads both with their checksums in the existing `fips_<ver>_x86_64_pfsense` artifact. No VM harness, no snapshot pin, no mirror. The `build` job, `release` job and the `*-pfsense-*` guard are untouched.

### Why a relabel and not a FreeBSD 16 builder

FreeBSD promises that binaries from an older release run on a newer kernel, and nothing the other way. A 16.0-CURRENT snapshot is months newer than any Netgate base (26.03.1 is a `main` commit from May, 26.07 from August), so a package built there runs on the appliance in the unpromised direction. The 15.1 static binaries run in the promised one. The evidence, from the relabel experiment asked for on #157:

| Host (Netgate installer, KVM, ZFS) | `uname -v` | `kern.osreldate` |
|---|---|---|
| Plus 26.03.1 | `16.0-CURRENT #12 plus-RELENG_26_03_1-n256546-1d1bfd578383`, 2026-05-20 | 1600011 |
| Plus 26.07 | `16.0-CURRENT #71 plus-RELENG_26_07-n256584-8183aef9d019`, 2026-08-07 | 1600018 |

On both, with the CE 2.8 artifact from #161's CI run relabelled exactly as this job now does it (binaries byte-identical, manifest `FreeBSD:16:amd64`): `testing/pfsense-install-smoke.sh` 45/45, TUN up with its mesh address, the responder answering directly, `fips-dns-setup` and `.fips` resolving through unbound once the resolver was restarted (that package predates #160), `pfSctl -c 'service reload packages'` leaving the daemon alone, a reboot bringing up one daemon with the resolver block regenerated, `fips-dns-teardown`, `pkg delete`. Then the two boxes were peered with each other over UDP, with a pass-in rule on the tun interface: link authenticated, `ping6` 200×56 and 100×1100 bytes each way with no loss, 10 MiB by TCP each way byte-exact; packets above the effective MTU of 1203 get ICMPv6 Packet Too Big per the no-fragmentation policy in `docs/design/fips-mtu.md`. Full detail is on #157 and in the README's test record. Not covered: physical hardware and CE 2.9.0 (no installer at hand).

### What CI does and does not do with the 16 package

It cannot `pkg add` it: pkg refuses a package whose ABI major differs from the host's. So the install smoke runs on the FreeBSD 15 package, which carries the same bytes, and the checker verifies the 16 label against the binaries from the outside. The 16-kernel install evidence is runs like the one above, which need the Netgate installer and cannot live in CI anyway. The relabel holds as long as the code builds on 15.1 without needing something only 16 provides; if that changes, a FreeBSD 16 build host is needed again, and the previous head of this PR is the record of how to get one.

### Docs

- Both READMEs and the build script's comment now state the compatibility direction instead of "the syscall ABI is stable within a major", as asked on the earlier revision.
- The pfSense README's target table, artifact table and "FreeBSD 16 is not released" section describe the relabel; it says CE 2.8.1 is upgrade-only and why; the test record has the two runs above and the earlier 26.07 run; the note that the tun interface is destroyed on exit is corrected to what was observed (it stays listed, down and unheld).
- CHANGELOG entry updated.
- The top-level README gains a paragraph under the platform table saying what pfSense is in packaging terms, which releases CI builds for, and that the package is a workflow artifact rather than a release asset; no column, since a column there means a released package.

### Stacking and overlap

- Stacked on #161 (its commit 21d52798 is the first in this PR) because both edit the same block of the `pfsense` job; the second commit is this change. Once #161 lands this rebases to one commit.
- #160 rewrites the README test record paragraph that this PR also rewrites; whichever lands second rebases, and this PR's version already includes the 26.07 run #160 describes.

### Checks

No Rust changes, so `cargo fmt`, `clippy` and `cargo test` are unaffected. The three CI parity scripts pass locally. The workflow's shell was extracted and syntax-checked; the relabel command, the checker on its output and the smoke test of that output were run in a FreeBSD 16 VM and on the two pfSense boxes. The `pfsense` job on this PR is the first CI run of the two-package job itself.

Not in this PR: keeping the VM harness script as a developer tool (offered separately if wanted), renaming the artifact to say freebsd15/16, and #160's README follow-ups.


jmcorgan:
The substantive reply is on #157, since the builder choice belongs there and it decides this pull request. What follows is the part specific to the diff.

The harness itself is careful. The ESP and serial-console route boots the plain image with no cloud-init and no third-party action, the image is hash-checked before anything boots, and the artifact is kept out of `release` twice over, by name pattern and by the `-pfsense-` filename guard. My reservations are about what it is pinned to rather than how it boots.

Three changes I would want in the diff regardless of how #157 resolves. Drop the reasoning that a static binary depends only on the kernel syscall ABI, stable within a major, from the script header, the README and the commit message; FreeBSD's compatibility promise runs the other way, and the README paragraph you left in place already says to build from a snapshot close to the appliance's base. In the refresh procedure, prefer the oldest announced snapshot still retained over the newest dated directory, for the same reason. And in the README, record what has actually been measured, which is Plus 26.07 in a VM without mesh traffic.

On retention: with no `ci-images` release the job falls back to download.freebsd.org, and snapshots are kept about two weeks, so once the current pin rotates the job goes red on every push and every pull request until the image is mirrored. Creating that mirror is my action rather than yours, so the landing order in your body is a real precondition and not a formality. `main` 20260916 is announced now, which at least makes the re-pin possible.

Separately, the install smoke test in your second commit is useful for the existing FreeBSD 15.1 job whatever happens with the builder. Would you split it into its own pull request so it can land on its own? If you do, a log rotation check would be a good addition, since the daemon is what that class of defect bites: force `newsyslog -F` on the shipped entry, then confirm the daemon's open log inode still matches `/var/log/fips.log`.

jmcorgan:
Landed on master as eb169ba4, and merged forward into next. Rebasing onto #160 conflicted in the pfSense README's test-record paragraph; I kept your text from this PR, which already carried #160's edits, so the resulting README is identical to your branch. On the master tip both packages built from the one 15.1 build and passed the checker (59 of 59 each). Closing by hand, since the rebase changed the SHA; #157 is closed alongside.

PR #161, opened 2026-09-17
testing: install-smoke the pfSense package in the CI VM

fr34aky:
## Install smoke test for the pfSense package, split out of #159

Requested on #159 and #157: the second commit of #159 on its own, so it can land on the existing FreeBSD 15.1 `pfsense` job whatever happens with the FreeBSD 16 builder. It depends on neither #159 nor #160 and closes no issue; #157 stays open for the builder decision.

`check-pfsense-pkg.sh` reads the package from the outside and drives the boot script against a stub. Nothing installed the package or ran the real daemon under `daemon(8)` on a kernel of the target's major, so a post-install that leaves `fips.yaml` world-readable, a daemon that dies at TUN setup, a supervisor that keeps writing into a rotated log inode, or a pre-deinstall that lets `pkg delete` remove the binary under a running process would all have passed.

`testing/pfsense-install-smoke.sh` now does, in the 15.1 VM after the checker: `pkg add`; the seeded configs, `fips.yaml` at 0600, the `fips` group; `fips.sh start`, the pidfile, `status`, `fipsctl show status` answering on the control socket, `drill` getting an answer from the responder on `127.0.0.1:5354`; a forced `newsyslog -F` on the shipped entry, then the `daemon(8)` supervisor holding the new `/var/log/fips.log` inode open, the rotated inode released, a `.0.bz2` generation kept and the daemon still answering; a second `start` (what `rc.start_packages` does on every WAN address change) exiting 0 with "already running" and still one process; `restart` with a new pid; `stop` with no process and no pidfile left; then `pkg delete` against a running daemon, checking pre-deinstall stopped it and the binaries, boot script and unmodified configs are gone. It cleans up after itself on failure.

It is plain sh, because pfSense ships no bash, and is written to run on pfSense too as the first thing to run against a new package there. Its header says what a plain-FreeBSD pass does not reach: the `config.xml` edit, unbound forwarding `.fips`, `/var` as a RAM disk, pf.

### The rotation check

The shipped `fips.newsyslog` names the supervisor's pidfile, and `fips.sh` starts `daemon(8)` with `-H`, so newsyslog's SIGHUP after the rename must make the supervisor reopen its `-o` file. The check drives `newsyslog -F -f` from the shipped entry, since `fips.sh` copies it into `/var/etc/newsyslog.conf.d` on pfSense only, and compares the inode in the supervisor's `fstat` output against `/var/log/fips.log` before and after. It is the supervisor that holds the file, not the `fips` process, whose stdout is a pipe to it.

### Differences from the commit on #159

- Only the `pfsense` (15.1) job runs it; the FreeBSD 16 job does not exist here.
- The `shellcheck --shell=sh` lint of the script, which #159 had in its FreeBSD 16 job, is a step of the `pfsense` job, before the VM boots. It installs shellcheck with apt if the runner lacks it, as `package-openwrt.yml` does for `install-nak.sh`; that is the only new build-time requirement, and it is on the runner, not in the package.
- The log rotation check, added at the maintainer's suggestion on #159.
- The script header, CHANGELOG and README mention only the 15.1 VM.

### Overlap with open work

- #159 carries the earlier version of this commit as its second; once this lands, #159 drops it in its next revision.
- #160 adds a README sentence saying the CE 2.8.1 package is "built and checked only", which this makes untrue; whichever lands second reconciles it, as noted there.

### Checks

No Rust changes, so `cargo fmt`, `clippy` and `cargo test` are unaffected; the integration suites do not cover packaging. Target is `master` because the pfSense package is under Unreleased in the CHANGELOG.

The first push failed the CI parity job in `testing/check-trailing-log.py`, which reported the `log()` functions of the acl-allowlist and firewall suites as having a status-testing caller. The caller it found was this script's `if out=$(newsyslog -F ...); then`: the checker's `if ... <fn> ; then` pattern, `\bif\s+\S*\s*<fn>\b`, has no word boundary before the name, so `newsyslog` matches `log`. The newsyslog call is in a `force_rotation` helper here, which is the natural shape anyway. The checker itself is not touched in this PR; a `\b` before the name in that pattern would close the false positive, as a separate change.

Run locally on a FreeBSD 16.0-CURRENT amd64 VM (the 20260907 snapshot, under qemu/KVM) against a package built there from this branch: `pfSense install smoke test PASSED (44 checks, 0 skipped)`, `shellcheck --shell=sh` 0.11.0 clean. The rotation check was also run negatively: with `-H` removed from the `daemon` line in the installed `fips.sh`, `newsyslog -F` moved the log from inode 31655 to 32636 while the supervisor kept 31655 open, which is the failure the check reports. The 37-check version without rotation passed in #159's CI in the 15.1 VM (run https://github.com/jmcorgan/fips/actions/runs/34792413899, commit d507f29d). The `pfsense` job on this PR is the 15.1 run of the version with rotation.


jmcorgan:
Landed on master as 0717d27f, on top of #160, with the patch unchanged, and merged forward into next. On the master tip the smoke test passed 44 of 44 in the FreeBSD 15.1 VM (the /var/etc newsyslog check only runs on pfSense itself). Closing by hand, since the rebase changed the SHA.

PR #162, opened 2026-09-18
ci: publish the pfSense packages as release assets

fr34aky:
## Promote the pfSense packages from workflow artifact to release asset

The pfSense packages have been kept out of releases by design until one was installed on a real pfSense box. This PR proposes that the bar is met and lifts the guard. It is the package maintainer's proposal; whether the project's release carries these packages is your call, which is why it is separate from #159.

### Evidence

All on Netgate-installed KVM VMs, so Netgate's kernel, with the package #159's CI job produces (the FreeBSD 15.1 static build, relabelled for FreeBSD 16):

| Host | `uname -v` | `kern.osreldate` |
|---|---|---|
| Plus 26.03.1 | `16.0-CURRENT #12 plus-RELENG_26_03_1-n256546-1d1bfd578383`, 2026-05-20 | 1600011 |
| Plus 26.07 | `16.0-CURRENT #71 plus-RELENG_26_07-n256584-8183aef9d019`, 2026-08-07 | 1600018 |

On each: `testing/pfsense-install-smoke.sh` 45/45 (log rotation included); TUN up with its mesh address; the responder answering directly; `fips-dns-setup` and `.fips` resolving through unbound; `pfSctl -c 'service reload packages'` leaving the daemon alone; a reboot bringing up one daemon with the resolver block regenerated; `fips-dns-teardown`; `pkg delete` against the running daemon. Then the two boxes peered with each other over UDP with a pass-in rule on the tun interface: link authenticated, `ping6` 200×56 and 100×1100 bytes each way with no loss, 10 MiB by TCP each way byte-exact. The earlier run on Plus 26.07 with a snapshot-built package, and the aarch64 hardware run, are in the README's test record as well.

Not covered: physical appliances, and CE 2.9.0 (no installer at hand). If you want either before promotion, say so and this waits.

### The change

- `release` needs `[build, pfsense]`, so a pfSense build failure holds a release the way a FreeBSD one does; that is what a release asset means, and it is the one behavioural cost of this PR.
- A second download step fetches the `fips_*_x86_64_pfsense` artifact into `dist`, and the pre-publish check requires all three packages (`*-freebsd-*.pkg`, `*-pfsense-ce2.8-amd64.pkg`, `*-pfsense-ce2.9-plus26-amd64.pkg`) instead of failing on a pfSense file. The existing sidecar verification loop then covers all three; each job wrote its sidecars inside its own VM.
- `checksums-freebsd.txt` and the upload glob already cover every `.pkg` in `dist`, so they are untouched.
- The pfSense README's artifact table and release paragraph, the top-level README paragraph and the CHANGELOG entry say release asset instead of workflow artifact. The `pfsense` job's header comment is rewritten; the job itself is unchanged.

Stacked on #159, which is stacked on #161; the first two commits are theirs, the third is this. It rebases to one commit as they land.

### Checks

No Rust changes. The workflow parses and the edited release step passes `bash -n`; the parity scripts pass. The release job only runs at a tag, so its first real run is the next release; the pattern it downloads matches the artifact name the `pfsense` job has produced on every run of #159.


jmcorgan:
Landed on master as 66c96386, with the patch unchanged, and merged forward into next. The pfSense packages will be attached to releases from the next tag onward, which will also be the first real run of the new release-job checks. Closing by hand, since the rebase changed the SHA.

PR #164, opened 2026-09-18
feat(netmon): let an embedder wake the medium-change detector

### LINKED ISSUE #146: Medium-change detection has no embedder push seam, so Android has no answer under Doze
Follow-up from #142 / PR #144, filed at @jmcorgan's request.

### Where this stands after #144

Android now takes the netlink wake source rather than an unbound `PF_ROUTE`
socket that never fired. When SELinux policy allows the group bind, a medium
change reaches the detector as a kernel event; when it does not,
`build_wake_source` falls back to the poll timer at
`node.netmon.poll_interval_secs` (default 5s).

Both of those are process-level mechanisms, and that is the gap. Under Doze —
and under standby, and after the app is backgrounded — a periodic timer is not a
reliable wake-up, and a netlink socket the app is not scheduled to read is not
either. The signal Android actually offers for this is a push:
`ConnectivityManager.registerNetworkCallback`, whose `onAvailable` /
`onLost` / `onCapabilitiesChanged` fire on exactly the transitions this detector
exists to catch — Wi-Fi ↔ cellular being the first of them.

### The concrete gap

There is no way for an embedder to inject a change. The whole channel is
private:

- `NetChange` and `NetChangeTx` are `pub(crate)`
  (`src/node/netmon/mod.rs:122-124`, `:242`).
- `spawn_detector` creates the channel and keeps the sender inside the detector
  task (`:402`).
- `Supervisor` holds only the receiver (`src/node/lifecycle/supervisor.rs:743`).

So an Android embedder holding a live `NetworkCallback` has nowhere to put the
event. The module docs already describe this as the intended shape — "an
embedder push on iOS", and for BLE "pushed by the BLE transport rather than
sampled here" (`src/node/netmon/mod.rs:39-43`, `:71-73`) — but the seam those
sentences describe does not exist in the API.

That makes this issue mostly about **exporting the seam**, with Android as its
first consumer. The same seam is what iOS needs (`NWPathMonitor`) and what the
BLE adapter-state signal needs, and none of them can be written until it is
there.

### Suggested shape

1. **A public push entry point.** Something an embedder can call on the node —
   "the host's network attachment changed, re-sample now". It need not carry a
   payload: the detector already owns the fingerprint comparison, the debounce
   and the settled-back suppression, so a bare nudge is enough and keeps the
   decision in one place. That is the same property that let the kernel backends
   land behind one seam.
2. **Keep the sampler underneath it.** A push should not replace the fingerprint
   comparison, or a spurious callback becomes a spurious socket rebind.
   `MIN_CHANGE_INTERVAL` and the settled-back suppression should apply to a
   pushed nudge exactly as they do to a kernel event.
3. **Android wiring, in whatever owns the JNI surface.** `registerNetworkCallback`
   with a `NetworkRequest` for `INTERNET`, calling the entry point above.
   Whether that lives in this repo or in an embedder is a question for you — the
   crate has no JNI layer today.

### Why it is separate from the netlink fix

#144's `cfg` widening is correct on its own terms: where policy allows the bind,
netlink is the lower-latency signal and the right default. This issue is about
the case where the process is not running enough to read it. The two compose —
push for the backgrounded case, netlink for the foregrounded one — rather than
one replacing the other.

### Environment

Against `master` at `b1a1043` (PR #144). Not reproduced on a device; the Doze
behaviour is from the Android power-management documentation, and the missing
seam is from the code above.

fr34aky:
Closes #146.

## What this does

Exports the embedder push seam the netmon module docs already describe but the API did not have. `Node::netmon_trigger()` hands out a `NetmonTrigger`; `NetmonTrigger::poke()` wakes the medium-change detector now rather than at its next poll.

The channel was private end to end (`NetChange` / `NetChangeTx` are `pub(crate)`, `spawn_detector` keeps the sender inside the detector task, the supervisor holds only the receiver), so an embedder holding a live platform callback had nowhere to put the event.

## Shape, against the three points in the issue

1. **A public push entry point.** `poke()` is a synchronous `Notify::notify_one`, callable from any thread, before or after `start()`. It is a detached handle rather than a method on the running node, because `run_rx_loop` borrows the node exclusively for its lifetime. It carries no payload.
2. **The sampler stays underneath it.** A poke only decides *when to look*. The fingerprint comparison, the debounce, the settled-back suppression and `MIN_CHANGE_INTERVAL` all run in `run_detector` after `wait()` returns, so they apply to a pushed wake-up exactly as they do to a kernel event; a spurious callback costs a sample, not a socket rebind. The push is selected *beside* the platform source and the timer backstop, never instead of them. A poke that lands while the detector is mid-sample is held as a permit rather than lost, and a burst coalesces into one wake-up.
3. **Android wiring is left to the embedder.** The crate has no JNI layer, so `registerNetworkCallback` (and `NWPathMonitor` on iOS) stays on the app side and calls `poke()`.

The trigger is allocated with the node rather than lazily, so fetching it after `start()` still reaches the running detector, and the same handle is handed to the detector again on a `stop()` / `start()` cycle. A rebuilt `Node` has a new trigger; the rustdoc says so.

## One behaviour beyond a bare nudge: the second look

A platform callback is not ordered against the routing table it reports on. On Android the `ConnectivityManager` callback can arrive before the default-route swap is visible to the `connect(2)` probe. A kernel source is woken again by every route message in the burst, but a poke is a single shot: it would be spent on a sample of the old picture, and the change left to the poll timer, which is the latency the push exists to remove.

So a **pushed** wake-up whose sample shows nothing moved samples once more after `node.netmon.debounce_ms`. Kernel and timer wake-ups are unchanged, so a quiet poll period still costs one sample (there is a test for that). This rests on reasoning about callback ordering rather than a device measurement; if one second look turns out not to be enough on real hardware the knob is a loop bound, and I am happy to change it.

Related rustdoc guidance: wire `poke()` to the callbacks that mean the attachment moved (`onAvailable`, `onLost`, `onLinkPropertiesChanged` of the default network), not `onCapabilitiesChanged`, which fires every few seconds on cellular and would turn the push into a faster poll. A poke while no detector is running (before `start()`, or between `stop()` and `start()`) is held and costs one extra sample at the next start.

## Why Android needs it

Beyond the Doze argument in the issue: on a Pixel 9 Pro the netlink group bind is refused to the app by SELinux policy and the node logs `Kernel medium-change events unavailable; falling back to polling`, so in practice the detector is on the 5 s poll timer there, on the one platform where the embedder knows the moment the default network moved.

One thing this does not change: on Android the connected-UDP fast path is not compiled in (`drop_connected_sockets_after_net_change` is a no-op outside Linux/macOS), so the reaction a poke buys there is the heartbeat that lets the far side re-pin sooner. Likewise the existing startup `warn!` about falling back to polling still fires on Android even when an embedder pokes; the detector cannot know at startup whether anyone will, so I left it alone.

## Tests

Detector level, on a paused clock with an hour-long poll so only the poke can be what woke it:

- `an_embedder_poke_wakes_the_detector_before_the_timer_would`
- `a_poke_before_the_detector_waits_is_not_lost` (asserts the sample count, so a burst really is one wake-up)
- `a_poke_that_runs_ahead_of_the_route_change_still_catches_it`
- `a_timer_wake_that_finds_nothing_does_not_take_a_second_look`

Node level: `the_nodes_trigger_reaches_the_running_detector_across_a_restart` drives `Node::netmon_trigger()` through `start()`, `stop()` and another `start()`, observing that the running detector consumes the held poke. I checked that it fails when `start()` is handed any other trigger.

## What was run locally

Pinned 1.94.1 toolchain, on the branch as pushed:

- `cargo fmt --check`, `cargo build`, both clippy passes with `-D warnings`, `cargo test` (2494 passed, 0 failed).
- The Android cross-check, `cargo ndk -t arm64-v8a clippy --lib -- -D warnings`, in a container with NDK r27c.
- The full `./testing/ci-local.sh -j 6` matrix: 39 stages, including `medium-change` and all six chaos scenarios.

Three caveats on that matrix run, all about my host (Arch) rather than the change:

- `nft-syntax` calls `sudo nft -c`; I ran the same `nft -c -f packaging/common/fips.nft` as root inside an unprivileged user+net namespace (`unshare -rn`) instead. The file is untouched by this PR.
- The chaos simulator's `jinja2` came from a virtualenv.
- In the full run `deb-install` failed at the host-side glibc-floor check because Arch has no `dpkg`. I re-ran `--only deb-install` with `dpkg` / `dpkg-deb` provided from a `debian:bookworm-slim` container and it passed (4 binaries at glibc 2.34 against the 2.35 floor, install test green). So 38/39 in the full run plus that suite on its own, not a single 39/39 run.

## Docs

Module docs in `src/node/netmon/mod.rs` describe the seam as present rather than "still to come"; rustdoc on `Node::netmon_trigger` and `NetmonTrigger`; `CHANGELOG.md` entry under Unreleased → Library surface and internals. No config, wire-format or dependency changes; the API change is additive, hence `master`.

The two commits are the feature and a follow-up from a review pass before requesting review; they squash to one change.


jmcorgan:
Thanks, this reads well. I checked the wiring against tokio and it holds: the handle survives stop and start, a poke is never lost, and the second look only applies to pushes. What I'm asking for is tests and wording, not behaviour.

1. `a_poke_that_runs_ahead_of_the_route_change_still_catches_it` still passes if the second look's `sleep` is removed, because the `scripted` sampler returns `cell` on its third call no matter when that call happens. Could you have the sampler switch from `wlan` to `cell` at a point on the paused clock (say 100 ms in), so the test only passes if the second look actually waits? [tests.rs#L651-L675](https://github.com/jmcorgan/fips/blob/d9d575335e79a408d0495c328895653ee7d129f9/src/node/netmon/tests.rs#L651-L675)

2. Two small tests: a push attached to an event source (`WakeSource::events(...).with_push(trigger)`), showing a ping and a poke each still wake the detector; and a poke when nothing moved, showing it reports nothing.

3. A few statements need adjusting to match the second look:
   - A spurious poke costs two samples at the default debounce, not one (CHANGELOG, PR body). The same goes for a poke held while the detector is stopped (rustdoc).
   - "A burst coalesces into one wake-up" holds only when no detector is waiting. With the detector parked in `wait()`, a burst gives two wake-ups: one to the waiter, one stored.
   - The `Node::netmon_trigger` doc says a poke is "held for the next one". Worth adding that with `node.netmon.enabled: false` there is no next one and the poke does nothing.

No need to squash or edit the commit messages; I'll write the message when I squash-merge.

fr34aky:
Thanks — all three addressed in d5bd691dd327, tests and wording only.

1. The second-look test now uses a time-based sampler (`switching_at`) that flips `wlan` → `cell` 100 ms after the poke on the paused clock, and asserts the sample count. I checked it fails with the second look's `sleep` removed, and that it is the only test that does.
2. Added `a_push_beside_an_event_source_leaves_both_live` (`WakeSource::events(...).with_push(trigger)`; a ping wakes it, then a poke wakes it) and `a_poke_when_nothing_moved_reports_nothing` (quiet, and asserts the baseline-plus-two sample count).
3. Reworded the `NetmonTrigger` rustdoc, the `Node::netmon_trigger` doc, the CHANGELOG entry and the PR body: a spurious poke, and one held across a stop, costs two samples at the default debounce; a burst coalesces into one held wake-up only when nothing is waiting, and gives two when the detector was parked in `wait()`; with `node.netmon.enabled: false` there is no next start and a held poke does nothing.

fmt, both clippy passes and `cargo test` (2496 passed) are green on the pushed commit. Left the commits unsquashed as you asked.

jmcorgan:
Landed on master as f9da388b, squashed into one commit with your authorship and author date, and merged into next. Closing by hand since the squash has a new SHA. Thanks for the quick turnaround on the review.

PR #165, opened 2026-09-20
Package FIPS for RPM-based distributions

glima:
Fedora, RHEL and openSUSE run the same glibc build the `.deb` carries, and had no way to install it. This adds an RPM built the same way, attached to the Release object beside the `.deb` and the tarball.

`make -C packaging rpm` produces it from `packaging/rpm/fips.spec` and a `build-rpm.sh` that mirrors `build-deb.sh`'s options and dev-version derivation. The release workflow's Linux job grows one step, in both matrix legs.

## The package is `fips-mesh`

Fedora's namespace already carries a `fips`: an OpenGL FITS image viewer, [matwey/fips3](https://github.com/matwey/fips3), currently 3.4.0, which owns `/usr/bin/fips`. A package named `fips` at 0.6.0 is an *older* release of that one to every RPM tool.

I hit the consequence before I had finished writing the packaging. A test machine had the mesh daemon installed, enabled, running with a persistent identity; an ordinary update transaction found a "newer fips", took it, and the host came back from a reboot with no `fips.service`, no `fipsctl`, and an image viewer at `/usr/bin/fips`. Nothing warned, because from dnf's side nothing unusual had happened.

The two still cannot coexist, both shipping `/usr/bin/fips`, so the spec declares `Conflicts: fips` — rpm would refuse on the file conflict anyway, and this makes the refusal name the problem. The Debian side has no such collision and keeps its name.

## What it packages, and how

Same files in the same places as the `.deb`, same `fips` system group, `/etc/fips/fips.yaml` seeded from the shipped example only when absent, `hosts` and `fips.nft` as config files that survive an upgrade, `fips.service` and `fips-dns.service` enabled, `fips-firewall` and `fips-gateway` opt-in. The units are the Debian package's own files, unmodified: both packages install to `/usr/bin` and target the same systemd, so a second copy of four unit files would only be a second thing to keep in step.

`make rpm` compiles nothing on the host. It builds the binaries in the pinned image by way of `build-deb-container.sh` — which has already run the glibc floor and Depends checks on them — and packages those, so all three Linux artifacts carry the same objects. The release workflow calls the same script with `--no-build` over the binaries it has already recovered from the `.deb`: one build, three artifacts. `make rpm-host` remains for iteration, unchecked, as `deb-host` is.

`rpmbuild` itself runs in `FIPS_RPM_BUILD_IMAGE` (`almalinux:9`, declared next to `FIPS_BUILD_IMAGE` in `build-floor.env`). It compiles nothing and supplies two things a build host may lack: `rpmbuild`, and `systemd-rpm-macros`, without which the spec's `%systemd_post` would not expand and the package would ship scriptlets that quietly do nothing. Pinning to the oldest rpm in free support also keeps what it writes readable by every newer one. Docker is the only host prerequisite.

## Three deliberate differences from the .deb

**The floor is checked on the package rather than against it.** `cargo-deb` writes a dependency floor that can disagree with the binaries, which `check-deb-depends.sh` compares; rpm derives it from the ELF files and cannot disagree, which moves the risk one step back — to binaries built above the floor in the first place. The new `testing/check-rpm-floor.sh` reads `libc.so.6(GLIBC_x.y)` out of the finished package, the table dnf enforces at install time, and fails a build above the floor.

**There is no purge.** dpkg's `postrm purge` deletes `/etc/fips` and the group as a separate, deliberate operator action. rpm has no such distinction, so the same code would run on an ordinary erase — and during a distribution upgrade that erases and reinstalls — taking the node's identity keys with it. Erase leaves `/etc/fips` alone, and the spec says why at the point where the code would have gone. The incident above is also the argument for seeding rather than shipping `fips.yaml`: the key files and the config are not owned by the package, so the node's identity came back intact once the right package was installed again.

**Version and Release.** A dev build is `0.6.0-0.dev.git<date>.<sha>` rather than `0.6.0~dev+git<date>.<sha>-1`: rpm has no `~` and does not need one, since a Release beginning with `0.` sorts below the `1` a tagged release carries. No `%{dist}` tag either — there is one build, and a dist tag would name whichever image ran rpmbuild in an artifact that installs on all of them.

## Verification

On a Fedora 44 host, glibc 2.42:

- `make rpm` produced a package recording `GLIBC_2.34`, which passed the 2.35 floor check and installed and ran on `almalinux:9` and `fedora:44` — group created, `fips.yaml` seeded at mode 600, daemon runs.
- Packaging host-built binaries instead reproduces the defect the floor exists to catch: `GLIBC_2.39`, and `check-rpm-floor.sh` fails it.
- An upgrade over a hand-edited `fips.yaml` keeps the edit and leaves no `.rpmnew` or `.rpmsave`; erase keeps `/etc/fips`.
- With the FITS viewer installed, dnf names the conflict and declines; on a clean system `fips-mesh` installs and `dnf list --upgrades` finds no path from it to anything else.
- `cargo fmt --check`, `cargo build`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo clippy --all-targets -- -D warnings` and `cargo test` (2523 passed) all pass on this commit.
- `shellcheck --shell=bash` is clean on both new scripts.

## Known gaps

No install-test suite covers the RPM. `deb-install` exercises the `.deb` across five distributions on every push; the RPM is built on every push and installed by nobody but me. If you want a suite, the natural shape is the `deb-install` one with `almalinux:9` and `fedora:latest` legs, and I am happy to follow up with it.

The `aarch64` leg is built by the same matrix but I have only exercised `x86_64` by hand.


jmcorgan:
Thank you for this, and welcome. An installable package for the RPM distributions is a useful addition to the tree, and the name change to `fips-mesh` is fine.

However, packaging FIPS is a bit more complex than it might seem. Aside from moving binaries into place, the hardest part is managing DNS. For .deb based distributions, `fips-dns-setup` picks between five backends at run time, a result of systemd differences over time and how DNS is configured. So there are a few items that need revision in what has been implemented.

1. Drop the two `systemctl start` lines from `%post` and keep the two `enable` lines. You gated them on `[ $1 -eq 1 ]`, a fresh install; the Debian `postinst` gates every start on `if [ -n "$2" ]`, which is empty on a fresh install. The conditions are inverted, so `dnf install` starts the node where `apt install` does not. Since `fips-dns.service` is `Type=oneshot` running `fips-dns-setup`, that reconfigures the host resolver inside the install transaction, against a `fips.yaml` seeded from the example seconds earlier. With the starts gone it comes up at the next boot, or when the operator says so.

2. Add `/etc/systemd/resolved.conf.d/fips.conf` to the `%postun` erase cleanup. You clean three of the files `fips-dns-setup` writes outside the package, but not the one the recommended backend actually writes. On Debian a later `purge` is a second chance; rpm has no purge, so the erase branch is the only cleanup that will ever run.

3. Your CI is red on both architectures and it is not your build. `cleanup()` at `build-rpm-container.sh:60` is `[ -n "$STAGE" ] && rm -rf "$STAGE"`, and `STAGE` is only assigned in the `if [ "$NO_BUILD" -eq 0 ]` branch at `:69`. Under `--no-build` it is empty, so the test fails, and that failing test is the function's last command, making `cleanup` return 1. With `set -euo pipefail` the EXIT trap at `:130` dies on it, so the script exits 1 after succeeding and leaks its name directory. In the workflow it is piped into `tee`, so the step aborts on the pipeline status before its own checks run, which is why the log just stops. Add `return 0` at the end of `cleanup`, or write it as an `if` so the body rather than the test is last, and check `build-rpm.sh` and `check-rpm-floor.sh` for the same shape.

4. Take up your own offer of an RPM install suite, and make sure it covers `--no-build`. Item 3 is on the one path only the release workflow takes: `make rpm` and `make rpm-host` set `STAGE`, so nothing you can run by hand reaches it. The Debian side installs the real artifact across five distributions on every push and exercises install, upgrade and removal, which is the only reason I can claim a node behaves the same way on Debian and Ubuntu. It does not have to block this pull request, but I would like it soon.

5. Add a row to `packaging/build-floor.env` for the RPM distributions. That file holds the support policy, not a build setting: the rule is that FIPS installs on every version of a supported operating system its distributor still supports for free, and the 2.35 floor is Ubuntu 22.04's. Adding a family without a row leaves a real question open. EL9 has glibc 2.34 and installs; EL8 has 2.28 and cannot, and AlmaLinux 8 and Rocky 8 are in free support for a while yet. Either they are in the set and the floor is wrong for this family, or the policy is EL9 and newer, which is fine but needs saying next to the others. openSUSE Leap and the Fedora window want the same. You know this ecosystem better than I do, so I would rather have your answer than write mine.

6. Tell me whether SELinux needs anything declared. It is enforcing by default on Fedora and RHEL, and the daemon creates a TUN device, takes `CAP_NET_ADMIN`, and binds a control socket in `/run/fips` that is reached by group membership. Nothing in the diff mentions it. "I checked and it needs nothing" is a fine answer; I would rather have it from you than from a reporter.

7. Add a sentence to the packaging README about firewalld. `fips-firewall.service` loads an nftables table straight from `/etc/fips/fips.nft` on distributions where firewalld owns nftables. The unit is opt-in already, so saying what happens if you enable it is probably enough.

8. Leave `NAME_DIR=$(mktemp -d)` and the `STAGE` mount alone for now. A bind-mount source is resolved by the Docker daemon in the host's mount namespace, so where the build runs under a private `/tmp`, which `PrivateTmp=` in a systemd unit gives you, the daemon creates its own directory in the real `/tmp`, the container writes there, and the script reads an empty one. It passes everywhere without a private `/tmp`, which is why we shipped the same bug in our own Debian container script. Our fix puts the directory inside the output directory, which is already a shared bind mount. It is landing now, so please copy that shape once I tell you it is pushed rather than us writing two different fixes.

9. Make `--features` work or remove it. `rpm_args` always begins `--no-build --bin-dir /bins` and then appends `--features`, and `build-rpm.sh` refuses that pair and exits 1, after the full container build at `:79` has already run. The script's own usage advertises the option. If it should work, those binaries were already built with the features, so what you want is the Release marking without a second build: a marker-only option on `build-rpm.sh`, or fold it into the `--version` the container script already passes and drop `--features` from `rpm_args`.

10. Restart `fips-gateway` and reapply the firewall ruleset on upgrade. `%systemd_postun_with_restart` names only `fips.service` and `fips-dns.service`, so a host that opted the gateway in keeps running the old binary. Adding `fips-gateway.service` to that list is safe, because `try-restart` leaves an inactive unit alone. For the firewall, add an upgrade branch in `%post` or a `%posttrans` mirroring `postinst:142`, which does a `try-reload-or-restart` so a changed `/etc/fips/fips.nft` is applied in place; it must be a reload, because that unit's `ExecStop` deletes the table.

11. Pin or cache the rpm build image. `dnf install -y rpm-build systemd-rpm-macros` runs in a stock `almalinux:9` on every invocation, unretried and uncached, where the Debian builder image is retried across transient registry failures and cached between runners. `almalinux:9` is a floating tag, so the `systemd-rpm-macros` that expands `%systemd_post` into your shipped scriptlets is whatever the mirror served that minute: a floating input to a release artifact, declared in the one file whose purpose is preventing that. The workflow fires on three branches, every pull request and every tag.

12. Add a CHANGELOG entry. The precedent is the pfSense bullet under `#### Packaging`, which names the artifact, the build target and why it differs. Most of the text is already in your pull request description.

13. Confirm the Release sorting rather than changing it. Your description says a dev Release begins `0.` so it sorts below a tagged release's `1`, and `build-rpm.sh` derives that form. But on the release path `--version` is always supplied, so the Release becomes whatever follows the first hyphen of the Debian version string: this run produced `fips-mesh-0.6.0-dev+165.merge.1056.cfe8c1d8`. I think it still sorts the way you want, since rpm ranks a numeric segment above an alphabetic one, but that is not the mechanism your description gives, so please check it with `rpmdev-vercmp`.

Two things I went looking for and found you had already got right, in case it saves you second-guessing them: the `%postun` reasoning about identity keys surviving an erase, including the distribution-upgrade case, and keeping `fips.yaml` out of the packaged config files and seeding it instead.

glima:
Thank you — that review found real defects, and two of them were in paths I had convinced myself I had tested. All of it is in the force-pushed single commit `85a3405b`, except where noted below.

**1. Inverted start gating.** Fixed: the two `systemctl start` lines are gone, the enables stay. You are right about the inversion — `[ $1 -eq 1 ]` is rpm's fresh install and `[ -n "$2" ]` is dpkg's upgrade, so `dnf install` was doing what `apt install` deliberately does not, with `fips-dns-setup` rewriting the resolver inside the transaction against a `fips.yaml` seeded seconds earlier. Upgrades still restart what was running, through `%systemd_postun_with_restart`.

**2. `resolved.conf.d/fips.conf`.** Added to the erase branch. Worth passing back: the Debian `postrm purge` misses the same file — it removes the `dns-delegate`, `dnsmasq.d` and `NetworkManager/dnsmasq.d` ones and stops there. `fips-dns-teardown` removes all four on `ExecStop` and `prerm`/`%systemd_preun` stop the unit first, so on both sides the leak only bites a host where the unit was not running. Happy to send that one-liner as a separate PR rather than smuggle it in here.

**3. The `cleanup` trap.** Fixed by making the body the handler's last command. Your diagnosis was exact, including the leaked name directory — with `set -e`, a failing `cleanup` aborts the second handler before its `rm`. I reproduced both halves against binaries recovered from a container-built `.deb`: `build-rpm-container.sh --no-build` exits 1 before the fix and 0 after, producing the same package either way. The other two scripts are clear — `build-rpm.sh` traps a command rather than a function, `check-rpm-floor.sh` traps nothing.

The part I want to acknowledge rather than gloss: `make rpm` and `make rpm-host` both set `STAGE`, so nothing I could run by hand reached it, and my "verified on a Fedora 44 host" list was therefore verifying the wrong path. That is also why my first attempt to measure the fix came out green — I piped the script into `tail` and read `tail`'s exit code.

**4. RPM install suite.** Agreed, and it will cover `--no-build`. Not in this PR. The shape I have in mind is Packit + Testing Farm, which gives real Fedora and CentOS Stream VMs per pull request and would also answer item 6 properly; if you would rather it stayed inside `testing/ci-local.sh` with containers, say so and I will follow the `deb-install` pattern instead. Containers cannot answer the SELinux half, which is the only reason I lean the other way.

**5. The floor row.** Added, and the floor is now 2.34. The RPM family's lowest member is RHEL 9 and its rebuilds at glibc 2.34, supported to 2032-05; openSUSE Leap 15.6 is 2.38; Fedora's current two are 2.42 and up. RHEL 8 and its rebuilds are recorded as a deliberate exclusion — glibc 2.28, in free support until 2029-05, and reaching them means an EL8-era build image and a second floor, which nobody has asked for. The comment says where that decision would go if someone does.

The reason for tightening rather than leaving 2.35: the binaries built in `FIPS_BUILD_IMAGE` happen to reference nothing above 2.34 today, which is the only reason a package built there installs on EL9 at all. Without the declaration that is luck, and the first commit to pull in a 2.35 symbol would pass `check-glibc-floor.sh` and silently drop every EL9 host. One step tighter costs the Debian side nothing, since its own lowest is 2.35. Both checks pass with no headroom left.

**6. SELinux.** Honest answer: I have the static half and not the runtime half, and I would rather say so than send you "it needs nothing".

Fedora 44's shipped targeted policy allows `unconfined_service_t` both `tun_socket` (create, read, write, ioctl, …) and `capability { … net_admin … }`, which is the domain a `bin_t` binary started by systemd with no policy module of its own lands in. The control socket is created by `RuntimeDirectory=` under `/run`, and access to it is a DAC group check that SELinux does not mediate for an unconfined domain. So the expectation is that nothing needs declaring.

What I cannot do is confirm it on an enforcing host: the machine I have has SELinux disabled, and a container cannot stand in, since enforcement is kernel-wide. Getting from disabled to enforcing needs a reboot and a full relabel. I will confirm on an enforcing Fedora and a RHEL-family VM — `install`, `systemctl start fips fips-dns`, `fipsctl` as a non-root group member, then `ausearch -m AVC -ts boot` — and report back, either here or on the install-suite PR, whichever you prefer. If a confirmed answer should block this one, say so and I will hold it.

**7. `NAME_DIR` and the `STAGE` mount.** Left alone. My change to that file is the four-line `cleanup` body and a comment; the directories and the mounts are untouched, waiting for your shape. Ping me when it lands and I will copy it rather than invent a second fix.

**8. `--features`.** Fixed, the second way you suggested. The features do reach cargo — through the Debian container build that produces the binaries — so what was missing was only the Release marking. That is folded into the `--version` the container script already passes, and `--features` is no longer appended to `rpm_args`. It now produces, from one build: `fips-mesh-0.6.0-0.dev.git20260920.<sha>.features.profiling.x86_64.rpm`.

**9. Gateway restart and firewall reload.** Both added. `fips-gateway.service` joins `%systemd_postun_with_restart`, since try-restart leaves an inactive unit alone and so lists it without opting anyone in. `fips-firewall.service` deliberately does not join that list — its `ExecStop` deletes the table — and instead gets an upgrade branch in `%post` doing `try-reload-or-restart`, mirroring `postinst:142` including the failure message. `%post` of the new package runs ahead of `%postun` of the old one, so the ruleset is reapplied before the daemon restarts, which is the ordering the Debian side has.

**10. Pinning the build image.** Pinned by digest, and the package install inside it is retried three times. You are right that this is precisely what `build-floor.env` exists to prevent: the `systemd-rpm-macros` in that image is what expands `%systemd_post` into the scriptlets the artifact ships. The comment carries the command to move the pin and what to check when moving it. I have not built a cached builder image mirroring the Debian one — `pin or cache` read as either, and pinning removes the floating input while retry covers the transient failures. Say the word if you want the cached-image shape instead.

**11. CHANGELOG.** Added under `## [Unreleased]` → `### Added` → `#### Packaging`, following the pfSense bullet: the artifact, the build target, the naming collision and the floor change.

**One more, which came out of reading the built package rather than the source.** rpm expands macros inside `#` comments. Two comments of mine referencing `%systemd_postun_with_restart` and `%post` were expanding into the shipped scriptlets, so every upgrade would have run `systemd-update-helper mark-restart-system-units below,` and one more like it. Escaped, and `rpm -qp --scripts` now shows exactly the statements it should. It is the argument for checking the artifact rather than the spec, which is the same lesson as item 3.

Verified after all of the above, on Fedora 44 with glibc 2.43: the pinned-digest build produces a package recording `GLIBC_2.34`; install, reinstall and erase are clean on `fedora:44` and `almalinux:9`, with the group created, `fips.yaml` seeded at 0600, `/etc/fips` surviving erase and a stray `resolved.conf.d` drop-in not; with the FITS viewer installed, dnf names the conflict and declines. `cargo fmt --check`, both clippy passes and `cargo test` pass on the commit, and `shellcheck` is clean on the three scripts.

jmcorgan:
Thanks for the update. I rebuilt it at 85a3405b and installed, upgraded and erased it in almalinux:9 containers with systemd running. The start gating, cleanup trap, gateway restart, firewall reload and config and key survival all hold. What's left:

1. Our temporary-directory fix is now on master (https://github.com/jmcorgan/fips/commit/bf668384fa307f782cb0a6d30349b632bfe4b747). Please rebase and copy its shape into build-rpm-container.sh: create the name directory with `mktemp -d "$DEST_ABS/.name.XXXXXX"` and sweep stale `.name.*` directories first. STAGE needs to move under DEST_ABS too, so compute DEST_ABS before the build block. Shape to copy: https://github.com/jmcorgan/fips/blob/d6dfb3e1edb105c5f704573557481df764099e10/packaging/debian/build-deb-container.sh#L160-L243

2. openSUSE Leap: `Requires(post): shadow-utils` can't be satisfied there (Leap's package is `shadow`); use `Requires(post): (shadow-utils or shadow)`. Upgrades also don't restart the daemon on Leap, because `%systemd_postun_with_restart` only marks units and Leap's systemd has no trigger to act on the mark; plain `systemctl try-restart` in %postun works on both. Leap 15.6 is also end of life, so the floor row should name 16.0. If you'd rather not carry openSUSE yet, dropping it from the docs and floor table is fine.

3. In %post, run `systemctl daemon-reload` before the fips-firewall reload, as the Debian postinst does.

4. The erase cleanup removes `/etc/systemd/dns-delegate/fips.dns-delegate`, but the DNS scripts use `/etc/systemd/dns-delegate.d/`. Please add the `.d`.

5. The firewalld paragraph in packaging/README.md has it backwards: fips0 is filtered by firewalld's default zone whether or not fips-firewall.service runs, and on RHEL and Fedora Server that zone blocks the transport ports (2121/udp, 8443/tcp), so a node can't accept peers until they're opened. Please say how to open them, and suggest giving fips0 its own zone or `trusted`. This is from the nftables and firewalld docs, so please check it on a firewalld host.

6. build-rpm-container.sh: refuse `--features` with `--no-build` or `--bin-dir`, as build-rpm.sh does, and take the Debian build's output with `| tail -n 1` like every other caller.

7. Release sorting is right; I checked it with rpm's own comparison. But rpm does support `~` (since 4.10), so "rpm has no ~" in build-rpm.sh, the packaging README and the commit message should say the `0.` form was a choice.

8. On your build-image question: pin plus retry is enough, no cache needed. Please add a retried `docker pull` before the first `docker run`, since the implicit pull isn't retried.

9. Smaller corrections: the spec header still says the glibc floor needs no separate check; the upgrade comments say it matches Debian, but it deliberately doesn't (the RPM behaviour is the better one, so just say so); add `%dir` entries for the doc and license directories; leave the README table's Debian column as `.deb`; note that a default RHEL install has no resolver backend for fips-dns; and bring the PR body in line with the commit.

I'll answer the Debian postrm, install-suite and SELinux questions separately.

glima:
All nine are in `6518e964`, rebased onto `b0a3af83`, and the PR body now matches it. Four of them were mine to get wrong, including one claim I asserted without testing; more on that under 7.

**1. Temporary directories.** Copied from `bf668384`. `DEST_ABS` is computed before the build block, both scratch directories are created with `mktemp -d "$DEST_ABS/.stage.XXXXXX"` and `"$DEST_ABS/.name.XXXXXX"`, stale `.name.*` and `.stage.*` siblings older than 120 minutes are swept first, and the "the name travels through … a private /tmp is the usual cause" diagnostic is there when the name does not arrive. Thank you for holding it until yours landed — I would have written something narrower.

**2. openSUSE.** Kept, since the fixes were small. `Requires(post): (shadow-utils or shadow)`. The floor row now names Leap 16.0 at glibc 2.40 and records that 15.6 went end of life in 2025-12 — I checked both in `opensuse/leap:16.0`, where `groupadd` does come from `shadow`.

The restart is now a plain `systemctl try-restart` over the three units. Your diagnosis is visible in the built package: `%systemd_postun_with_restart` had expanded to `[ -x /usr/lib/systemd/systemd-update-helper ] && … mark-restart-system-units`, which is EL9's macro baked in at build time and depends on a file trigger in that distribution's systemd to act on the mark. Where the trigger is absent the mark is written and nothing reads it, so the upgrade is silent. The spelled-out form also made the comment honest about where this deliberately differs from the Debian `postinst` and why the simpler shape is the better one here.

**3. `daemon-reload`.** Added before the firewall reload, with the reason in the comment: the unit files this transaction installed are not loaded yet, since systemd's own reload comes from a file trigger at the end, so without it the reload acts on the pre-upgrade unit.

**4. `dns-delegate.d`.** Fixed. Worth handing back: `packaging/debian/postrm` has the same wrong path — `/etc/systemd/dns-delegate/fips.dns-delegate`, without the `.d` — which is where I copied mine from. Together with the `resolved.conf.d` drop-in that purge never removes, that is two one-liners on the Debian side. Happy to send them as a separate PR whenever you want it.

**5. firewalld.** You were right and my paragraph was backwards. Checked on a firewalld host, which is what the rewrite says:

```
firewalld: active   default zone: FedoraWorkstation
fips0 zone: no zone                                  → the default zone
FedoraWorkstation: ports 1025-65535/udp 1025-65535/tcp
public:            services dhcpv6-client mdns ssh, no ports
```

So the README now separates the two halves. Inbound peers arrive on the *ordinary* interface at the transport ports, which `public` blocks and Fedora Workstation's zone happens not to, with the `--add-port=2121/udp --add-port=8443/tcp` to open them. And `fips0` itself lands in the default zone because nothing assigns it one, so mesh traffic is subject to that zone as well as to the fips baseline, with `--zone=trusted --change-interface=fips0` as the way to keep the two decisions apart. The note about the fips table being invisible to `firewall-cmd` stayed, since that part was right.

I also added the RHEL resolver-backend note from item 9 there: no `dns-delegate` on that systemd, systemd-resolved installed but not enabled, dnsmasq absent, so `fips-dns-setup` reaches its last branch and prints manual instructions.

**6. `--features` guard and `tail -n 1`.** Both done. The refusal is before the container build rather than after it, and it names `--bin-dir` as well since that implies `--no-build`.

**7. `~`.** You are right, and I should not have written that. `rpm --eval '%{lua: print(rpm.vercmp("0.6.0~dev", "0.6.0"))}'` returns `-1` on the rpm in front of me, which I could have run before making the claim. Corrected in `build-rpm.sh`, the packaging README and the commit message: the `0.` Release form is this ecosystem's pre-release convention and a choice, not a workaround for something rpm lacks.

**8. Image pull.** A retried `docker pull` now runs before the first `docker run`, guarded on `docker image inspect` so it is a no-op on a warm host. The in-container package install keeps its own retry.

**9. The smaller ones.** Spec header no longer says the floor needs no separate check — it points at `check-rpm-floor.sh` and says what moved. The upgrade comments say the divergence from Debian is deliberate and why. `%dir` entries added for `%{_docdir}/fips` and `%{_licensedir}/fips`. The README table's Debian column is `.deb` again. PR body brought in line with the commit.

## Verification

I took your approach rather than my earlier one of reading `rpm -qp --scripts`: `almalinux:9` with `/usr/sbin/init`, systemd running.

- Fresh install: `fips` and `fips-dns` both `enabled` / `inactive`, group created, `fips.yaml` seeded at 0600.
- Upgrade with the daemon running: it stays active and `MainPID` moves 121 → 210, so the `try-restart` fires; `fips-gateway` stays `disabled` / `inactive`, so a host that never opted in is not opted in by the upgrade.
- Erase: units gone, the `dns-delegate.d` drop-in removed, `/etc/fips` and its keys kept.
- `make rpm` end to end produces a package recording `GLIBC_2.34`, passing the tightened floor; packaging host-built binaries instead is refused by `check-rpm-floor.sh` at `GLIBC_2.39`.
- `cargo fmt --check`, both clippy passes, `cargo test`, and `shellcheck` on the three scripts: clean.

## Still open

The install suite is next, covering `--no-build`. I have not started it pending your answer on shape — Packit and Testing Farm for real VMs, or a container suite following `deb-install` inside `testing/ci-local.sh`.

SELinux stays untested on an enforcing host, for the reason in my last comment. I will report a result rather than an expectation, on the install-suite PR or here, whichever you prefer.

jmcorgan:
Thanks, this is close. CI is green on every leg at 6518e964, and seven of the nine items are done as asked. What's left:

1. The `%postun` restart: my suggestion last time was wrong. A plain `systemctl try-restart` waits for all three restarts, so the rpm transaction and dnf's lock wait with it; if the new daemon can't start, fips-dns-setup's 30-second wait for fips0 lands inside the upgrade. Please make it `systemctl --no-block try-restart fips.service fips-dns.service fips-gateway.service >/dev/null 2>&1 || :` and reword the comment, the PR body and the commit message to say the restart is queued, not waited on. This needs to be right in the first published RPM, because on an upgrade it's the old package's `%postun` that runs.

2. The spec header still ends "so the glibc floor needs no separate check on this side". Please say that rpm derives the glibc requirement and check-rpm-floor.sh checks it against FIPS_GLIBC_FLOOR on the finished package.

3. The `%post` comments still refer to `%%systemd_postun_with_restart`, which the spec no longer uses; point them at the try-restart in `%postun`. They also say the enables happen "as the Debian postinst does", but Debian enables on every configure and this spec enables on first install only. Please say so, and that an operator's later disable survives an upgrade as a result. "A site that has an opinion about these units gets it" also overstates it, since the explicit enable overrides a disable preset.

4. The CHANGELOG entry and packaging/README.md still describe FIPS_RPM_BUILD_IMAGE as `almalinux:9`; it's pinned by digest now, so "AlmaLinux 9, pinned by digest" in both.

5. build-floor.env's header lists two consumers; build-rpm-container.sh and check-rpm-floor.sh source it too. Please name all four.

6. On the floor: I'm fine making RHEL 9 the project-wide floor at 2.34, since both families ship the same binaries. Please make the files say so: build-floor.env and packaging/README.md both say the build image is pinned to the oldest supported distribution, which is no longer true now that EL9 sits below it, and docs/getting-started.md says the binaries are built in a container pinned to the oldest of them. Move the CHANGELOG floor bullet from Added to Changed.

7. Please drop openSUSE for now: from README.md, docs/getting-started.md, packaging/README.md, the CHANGELOG entry and the build-floor.env table. Nobody has run the package there yet, and it can come back in a later PR together with a Leap 16 leg in the install suite. Keeping `(shadow-utils or shadow)` in the spec is fine.

8. On the Debian postrm: thanks for spotting it. We'll fix that one on our side, along with the same gaps in packaging/systemd/uninstall.sh, so there's nothing for you to send.

9. On the install suite, for the follow-up PR you offered: please follow the deb-install pattern in testing/, with almalinux:9 and the current Fedora legs, covering install, upgrade, erase and the --no-build artifact, so it runs everywhere the Debian suite runs. An upgrade leg would also catch item 1 coming back.

10. On SELinux: it doesn't need to block this PR, but please post the enforcing-host result before the first tagged release that ships the RPM, here or on the install-suite PR. One Fedora VM and one RHEL-family VM is enough: install, `systemctl start fips fips-dns` (so fips-dns-setup and fips-dns-teardown under /usr/lib/fips are exercised too), fipsctl as a non-root group member, then `ausearch -m AVC -ts boot`.

Two optional ones: build-rpm-container.sh could use `retry_build` from testing/lib/image-build.sh for the pull, so a recovered failure shows up as a warning on the run instead of passing silently; and in the firewalld paragraph, "does not open it in the fips table" is more exact than "does not open it on fips0", since the latter is only true while fips-firewall is running.

